PR #2, opened 2026-02-26
feat(ui): operator filter

johallar:
## Description

* Adds operator specific bulk request (required until we update bulk API)
* Displays timeline overlays for specific operators as striped bars
* New tooltip for operator overlayed timelines

* Hook-ify bulk timeline fetching (use for 'base' and 'operator specific' timelines)
* Move some component state up to jotai, simplify props across components
* Better cache key taking into account resource id, resource type, and operator id

## Screenshots
https://github.com/user-attachments/assets/80b69f7a-e582-46be-893a-138505520f7c


ui/src/hooks/useBulkTimelineFetch.ts
```diff
@@ -0,0 +1,74 @@
...
+  queryId: string;
+  debouncedZoomRange: ZoomRange;
+  baseEntries: Record<string, TimelineRequest<TaskFilter>>;
+  operatorId?: string | null;
+  enabled?: boolean;
+}) {
+  const store = useStore();
+
+  const { data } = useQuery({
+    queryKey: ['bulkTimelines', engineId, queryId, debouncedZoomRange, operatorId, baseEntries],
```
cmatzenbach: `baseEntries` is used as a key here but is not used anywhere else in the function; instead the function reads from `visibleEntriesAtom` instead. Is there a reason `baseEntries` isn't used?

ui/src/hooks/useBulkTimelines.ts
```diff
@@ -127,51 +107,85 @@ export function useBulkTimelines({
...
-    [rootItem, selectedTypes, entities, engineId, queryId, store]
+    [rootItem, selectedTypes, entities, operatorId, engineId, queryId, store]
   );
 
-  // Re-fetch a single item (e.g. after resource type change)
   const invalidateItem = useCallback(
-    (itemId: string) => {
-      store.set(timelineDataAtom(itemId), undefined);
+    (itemId: string, resourceTypeName: string) => {
+      store.set(timelineDataAtom(timelineCacheKey(itemId, resourceTypeName)), undefined);
```
cmatzenbach: This only clears the base atom - shouldn't it also clear the operator-filtered atom?

ui/src/hooks/useBulkTimelines.ts
```diff
@@ -127,51 +107,85 @@ export function useBulkTimelines({
...
 
-      if (Object.keys(newEntries).length === 0) return;
+      if (Object.keys(newBaseEntries).length === 0) return;
 
       const zoom = store.get(debouncedZoomRangeAtom);
       const windowSec = zoom.end - zoom.start;
+
       try {
-        const response = await fetchBulkTimelines(engineId, {
+        const baseRequest = fetchBulkTimelines(engineId, {
```
cmatzenbach: This and the subsequent call below are manually calling this outside of tanstack query, bypassing the cache entirely. Not a show-stopper but would be good to tie in.

ui/src/hooks/useBulkTimelines.ts
cmatzenbach: On second thought, `invalidateItem` isn't being called anywhere - do we need this function? Is this for future features/needs?

ui/src/hooks/useBulkTimelines.ts
```diff
@@ -63,55 +47,51 @@ export function useBulkTimelines({
...
+   * we can make the bulk data initialized true (allows single timelines to fetch themselves)
+   */
   useEffect(() => {
-    if (!bulkData) return;
-    for (const [id, entry] of Object.entries(bulkData.entries)) {
-      if (entry?.status === 'ok') {
-        store.set(timelineDataAtom(id), { data: entry.data, config: bulkData.config });
-      }
+    if (baseBulkData && (operatorId != null ? operatorBulkData : true)) {
+      store.set(bulkInitializedAtom, true);
```
cmatzenbach, line 92: I don't this the atom is being set back to `false`, even when the query or engine changes. Wondering if this could have knock-on effects if this hook is re-used for a different `queryId` within the same component tree.

ui/src/hooks/useBulkTimelines.ts
johallar: This function got the axe but forgot to remove, will remove

ui/src/hooks/useBulkTimelines.ts
johallar: good question, coded this up before the header navigation, will do some investigation

ui/src/hooks/useBulkTimelineFetch.ts
johallar: Yeah good call, definitely some kludg-yness here from migrating to a hook. I am going to try and remove the `visibleEntriesAtom` from this hook completely. It will receive a list of entries (`entries` prop) and then fetch them.

ui/src/hooks/useBulkTimelines.ts
johallar: Ah yes, the jotai provider has a key of query id, so when queryID changes we're working with a new store (`bulkInitializedAtom` initialized as false)

https://github.com/rapidsai/quent/pull/2/changes/BASE..04d8b0eefadeec0a257c755f5e06bdae206c4e75#diff-47de736653a989a8a9a1a4b668bba71f224bdf4e35893f94187b4d04e6d58c2bR25

ui/src/hooks/useBulkTimelines.ts
johallar: 👍 I didn't actually know you could do this until you brought it up in the last MR i'll update to the tanstack query client.

ui/src/hooks/useBulkTimelines.ts
cmatzenbach: Awesome! Yeah wasn't sure if that would be auto-cleared or not, sounds good

cmatzenbach (approved):

PR #3, opened 2026-02-26
feat(ui): header query navigation

johallar:
Add some navigation controls to the "Profile" page, allows user to navigate through engines/query groups/queries without having to start over.

https://github.com/user-attachments/assets/5b7d0109-7c56-4e91-af96-37c8ed42e5f5


cmatzenbach (approved):
Looks great!

PR #11, opened 2026-02-27
feat(instrumentation)!: add noop exporter that short-circuits at the …

johanpel:
…send site

Introduce EventSender<T> wrapping Option<UnboundedSender<Event<T>>> so that selecting ExporterOptions::Noop skips creating a runtime, channel, forwarder, and exporter entirely. Sends become a branch on None with zero channel overhead.

BREAKING CHANGE: Context::events_sender() now returns EventSender<T> instead of UnboundedSender<Event<T>>, and Context::push_event() has been removed in favor of EventSender::send().

### DIFF
--- a/crates/instrumentation/src/lib.rs
+++ b/crates/instrumentation/src/lib.rs
@@ -7,7 +7,10 @@ use quent_exporter_msgpack::MsgpackExporter;
 use quent_exporter_ndjson::NdjsonExporter;
 use quent_exporter_postcard::PostcardExporter;
 use serde::Serialize;
-use std::sync::Arc;
+use std::sync::{
+    Arc,
+    atomic::{AtomicBool, Ordering},
+};
 use tokio::{
     runtime::{Handle, Runtime},
     sync::mpsc::{UnboundedSender, unbounded_channel},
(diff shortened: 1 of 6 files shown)


crates/instrumentation/src/lib.rs
```diff
@@ -25,15 +25,38 @@ pub enum ExporterOptions {
...
+    fn clone(&self) -> Self {
+        Self(self.0.clone())
+    }
+}
+
+impl<T> EventSender<T> {
+    pub fn send(&self, event: Event<T>) {
+        if let Some(tx) = &self.0 {
+            if let Err(e) = tx.send(event) {
+                warn!("unable to send event: {e}");
```
mbrobbel: We can drop the sender to skip the send attempts if this errors

crates/instrumentation/src/lib.rs
```diff
@@ -75,6 +112,7 @@ where
             ExporterOptions::Ndjson => Arc::new(handle.block_on(NdjsonExporter::try_new(id))?),
             ExporterOptions::Msgpack => Arc::new(handle.block_on(MsgpackExporter::try_new(id))?),
             ExporterOptions::Postcard => Arc::new(handle.block_on(PostcardExporter::try_new(id))?),
+            ExporterOptions::Noop => unreachable!(),
```
mbrobbel: Maybe change this match so we can avoid the `unreachable` i.e. move the noop exporter check here.

johanpel: https://github.com/rapidsai/quent/pull/11/commits/e5a8fd224b1530cbce493932133be86251f3593a

is the best solution if we don't want the potential overhead of spawning the runtime when noop is used.

crates/instrumentation/src/lib.rs
johanpel: Not trivial since this is shared by multiple threads, but we can reduce the log spam like so: https://github.com/rapidsai/quent/pull/11/commits/b5f059289d294eac87b6855ab6b34c770fa0a957

johanpel:
Rust CI is not running yet, going to look into this first before merging

crates/instrumentation/src/lib.rs
```diff
@@ -49,9 +83,27 @@ where
...
     ) -> Result<Self, Box<dyn std::error::Error>> {
+        let exporter = match exporter {
+            None => {
+                debug!("using noop exporter");
+                return Ok(Context {
+                    handle: None,
+                    events_sender: EventSender {
+                        tx: None,
+                        disable_error_log: Arc::new(AtomicBool::new(false)),
+                    },
```
mbrobbel, line 97: We can derive `Default` for `EventSender` and use that here. Maybe this could even be a `Default` impl for `Context`?

crates/instrumentation/src/lib.rs
```diff
@@ -115,25 +167,21 @@ where
         });
 
         Ok(Context {
-            handle,
-            events_sender,
-            exporter,
+            handle: Some(handle),
+            events_sender: EventSender {
+                tx: Some(events_sender),
+                disable_error_log: Arc::new(AtomicBool::new(false)),
```
mbrobbel, line 173: Nit you can use `Arc::default`, or maybe create/use a constructor for `EventSender`?

crates/instrumentation/src/lib.rs
johanpel: Unfortunately this would force the event type to implement `Default`, not worth it I think, this is the only place where these types should be constructed.

crates/instrumentation/src/lib.rs
johanpel: I know it's trivial but I personally like the explicitness of a bool init like this,

mbrobbel (approved):

PR #14, opened 2026-03-03
feat(ui): long entity annotation

johallar:
* Create an annotation series for task time periods
* Adds an info bar above the timeline column headers. I think we'll need this space for more in the future, but currently shows selected operator(s), added a "reset zoom" action, and a settings popover which you can use to "hide tasks"

TODO:
* Probably needs more work on the metric used to decide to display this series. Currently if it's more than "n" bins, in the current view, it will show them.
* Do we want these to roll up to resource groups? Could get too busy. 
* Sub-millisecond bins to fully take advantage

https://github.com/user-attachments/assets/cba876c7-0b04-4bc2-b9ac-8ea7c07fa8dd


ui/src/components/timeline/Timeline.tsx
```diff
@@ -44,69 +47,142 @@ export function Timeline({
...
+          yAxisIndex: 1,
+        });
+      }
+    }
+
+    return allSeries;
+  }, [series, timestamps, marks, markAreaFillOpacity, markAreaBorderOpacity, markLabelTextColor]);
+
+  const [prevSeriesCount, setPrevSeriesCount] = useState(seriesOptions.length);
+  const notMerge = seriesOptions.length < prevSeriesCount;
```
johallar, line 150: Tell's echarts not to merge with previous data when there are less series than the last render (hide tasks, de-select operator)

cmatzenbach (approved):
Looks great!

PR #16, opened 2026-03-03
refactor!: make exporter/importer paths configurable and restructure crates

johanpel:
This PR makes all event exporter and importer file paths configurable via options, removing hardcoded `data` directory everywhere. Structs are introduced to pass options to importers/exporters, which will also support easily adding arguments in the future (e.g. buffer sizes).

To prevent some cyclic dependencies there is also some crate refactoring:
- `quent-collector` -> `collector/server`, `collector/client`, and `collector/proto`
- Move `Exporter` trait into `quent-exporter-types` and add `Importer` trait
- `quent-exporter` is nonw an umbrella crate that contains some easy-to-use functions to construct event exporters/importers.
- Importer construction is now moved out of `UiAnalyzer` implementations. Instead you have to provide an iterator over events upon constructing it so these options can come in in any way you please. Eventually we'll need to move away from iterators if we're going to back this with other types of data sources (e.g. columnar data files or databases).


crates/collector/client/src/lib.rs
```diff
@@ -172,11 +172,11 @@ where
 
         debug!("opening stream ...");
 
-        // Add the engine id to the metadata of the request, so the collector knows which engine id this all belongs to.
+        // Add the application id to the metadata of the request, so the collector knows which engine id this all belongs to.
```
johanpel:
```suggestion
        // Add the application id to the metadata of the request, so the collector knows which application this belongs to.
```

crates/collector/server/src/server.rs
```diff
@@ -54,31 +60,36 @@ where
                 ))
             })?;
 
-        let engine_id = Uuid::from_str(engine_id_str).map_err(|e| {
+        let application_id = Uuid::from_str(application_id_str).map_err(|e| {
             Status::invalid_argument(format!(
                 "metadata value for key \"engine-id\" is not a UUID: {e}"
```
johanpel:
```suggestion
                "metadata value for key \"application-id\" is not a UUID: {e}"
```

dhruv9vats (approved):

PR #20, opened 2026-03-03
feat(ui): operator statistics popup

cmatzenbach:
Adds a popup to each DAG node that displays the operator name, type, id, as well as any `custom_statistics` sent through from the backend.

https://github.com/user-attachments/assets/8864e3de-7d95-4179-bc06-92ff80a56d2b


johallar:
nodes with lots of stats should be scrollable or otherwise limit their height, with the latest simulator code a lot of nodes end up looking like this:

Screenshot 2026-03-03 at 11 46 09 AM

ui/src/lib/queryBundle.utils.ts
```diff
@@ -22,3 +23,13 @@ export const ENTITY_REF_TO_ENTITIES_KEY: Partial<Record<EntityRefKey, keyof Quer
 export function entityRefToEntitiesKey(entityRef: EntityRefKey): keyof QueryEntities | undefined {
   return ENTITY_REF_TO_ENTITIES_KEY[entityRef];
 }
+
+export function parseCustomStatistics(rawNode: unknown): Array<{ key: string; value: StatValue }> {
+  const statistics = (rawNode as RawNodeStatistics)?.statistics?.custom_statistics;
```
johallar: Ran into this testing w one of the sirius dags. I think we need to safeguard the `Object.values()[0]`

```
installHook.js:1 TypeError: Cannot convert undefined or null to object
    at Object.values (<anonymous>)
    at queryBundle.utils.ts:33:19
    at Array.map (<anonymous>)
    at parseCustomStatistics (queryBundle.utils.ts:31:37)
    at QueryPlanNode (QueryPlanNode.tsx:120:22)
```

ui/src/lib/queryBundle.utils.ts
```diff
@@ -22,3 +23,13 @@ export const ENTITY_REF_TO_ENTITIES_KEY: Partial<Record<EntityRefKey, keyof Quer
...
+  if (!statistics) return [];
+
+  return Object.entries(statistics).map(([key, tagged]) => ({
+    key,
+    value: Object.values(tagged)[0] ?? null,
```
johallar: This should handle `List` types, when the statistic looks like this:
`projected_column_names:  {List: {String: ["id", "name", "ts", "amount"]}}`

it'll show the `[object Object]` stringification for these

ui/src/lib/queryBundle.utils.ts
cmatzenbach: Good catch - saw that too after rebuilding off latest main.  Fixed

ui/src/lib/queryBundle.utils.ts
cmatzenbach: Wrote some code to unpack everything - it all displays nicely now.

ui/src/services/query-plan/types.ts
```diff
@@ -42,9 +42,9 @@ export interface QueryPlanNodeData extends Record<string, unknown> {
...
 }
 
-export type StatValue = string | number | boolean | null;
-type TaggedStatValue = Record<string, StatValue>;
-export type CustomStatistics = Record<string, TaggedStatValue>;
+export type StatValue = string | number | boolean | null | string[];
+type TaggedStatValue = Record<string, unknown>;
+export type CustomStatistics = Record<string, TaggedStatValue | null>;
 
 export interface RawNodeStatistics {
```
johallar: I think we can use the `Operator` type here (`RawNodeStatistics`), which has a `OperatorStatistics` type for custom_statistics.

johallar (approved):
Looks good! would like to see the quent bindings used for the types where possible

PR #21, opened 2026-03-03
fix(ui): explicit y-axis bounds

johallar:
Define the yaxis min/max to avoid echarts determining it itself. This prevents y-axis bounds jumping 

Before:

https://github.com/user-attachments/assets/7801a0e8-704b-4272-9cf0-d148a7fb456c

After:

https://github.com/user-attachments/assets/f9a6a5ff-05d8-48c3-a84a-60be20f4b7ae


cmatzenbach (approved):
BEAUTIFUL!  Thanks for taking care of that

PR #22, opened 2026-03-03
feat(ui): zoom/pan from resource timelines

johallar:
Enables specific navigation from the resource timelines as well as from the timeline controller. Not a good flow to have to use the controller for everything. Deets:
* Hold shift + scroll to zoom in/out on the resource timelines
* Click and drag to pan

https://github.com/user-attachments/assets/09e438a8-713d-4f4b-a654-d959269e89cf


cmatzenbach (approved):
Looks and works great!

PR #23, opened 2026-03-04
feat: list engines with metadata

johanpel:
This PR provides `list_engines` route with a `?with_metadata=true` parameter which will cause the event data to be loaded until the engine init event is found. This is commonly cheap as its the first event in the file. But it can potentially be expensive in the current shape of the project, as in the worst case that event is at the end and the whole file needs to be scanned. It's also not perfect yet because we don't have pagination. It's a stopgap solution until we have an engine indexing system.

`list_engines` in the query engine server crate still had a hardcoded path to `./data`, and it also only scanned for `ndjson` files instead of also `postcard` and `msgpack`. This PR fixes that too.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1375,6 +1375,7 @@ dependencies = [
  "tokio",
  "tracing",
  "ts-rs",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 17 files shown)


mbrobbel (approved):

johallar (approved):
LGTM UI side, TY!

PR #25, opened 2026-03-04
feat(ui): sub millisecond bins

### LINKED ISSUE #9: Handle Sub-millisecond bins
Need to change timelines from 'time' type axes to 'categorical' which will track bins by a categorical start time. 

`value` would maybe be ideal, but eCharts does not allow stacking of series with a `value` type x-axis. I think the math does not quite work with a continuous xaxis

### LINKED ISSUE #41: Deselecting a DAG node should clear the timeline highlight

johallar:
~To continue using echarts, we have to switch to using a categorical x-axis instead of a 'time' type xaxis to support sub millisecond bins.~

Fixes #9
Fixes #41 

Actually, we can use fractional millisecond timestamps to get the job done while still using 'time' axis in echarts. I thought it converted it to a date internally but it doesn't, it still uses the raw values for all the positioning math and stacking so this works! Little funky but works, I like it better than switching to a categorical axis and doing everything based on index.

Things get a little funky when you get down to ~10 nanosecond bin scale... we can deal with that later.

https://github.com/user-attachments/assets/83371669-40f0-49d1-8594-99e7dd3e1374

Bonus Fixes:
* Datazoom bug - If you have an operator id selected and expand more timelines, it would display the dataset in the % start/end selected in timeline controller instead of filling the whole timeline. 
* Bugfix: Deselecting operator ids did not clear from timeline  #41


ui/src/components/QueryResourceTree.tsx
```diff
@@ -47,7 +48,7 @@ function QueryResourceTreeContent({ queryBundle, engineId }: QueryResourceTreePr
 
   const startTime = queryBundle.start_time_unix_ns;
   const durationSeconds = queryBundle.duration_s;
-  const startTimeMs = useMemo(() => Number(startTime / 1_000_000n), [startTime]);
+  const startTimeMs = useMemo(() => nanosToMs(startTime), [startTime]);
```
cmatzenbach, line 46: This is much better than having to do that division we were doing above everywhere

ui/src/lib/timeline.utils.ts
```diff
@@ -22,18 +22,22 @@ import type { TimelineConfig } from '~quent/types/TimelineConfig';
 const MAX_TIMELINE_BINS = 400;
 const LONG_ENTITIES_BIN_MULTIPLIER = 30;
 
+/** Convert a nanosecond-precision bigint epoch to milliseconds, preserving sub-ms precision. */
+export function nanosToMs(ns: bigint): number {
+  return Number(ns / 1_000_000n) + Number(ns % 1_000_000n) / 1_000_000;
+}
+
 /**
  * Computes the number of bins such that each bin is >= 1ms wide.
```
cmatzenbach: We should probably update this comment, since it's no longer relevant with the function changes

ui/src/lib/timeline.utils.ts
```diff
@@ -297,11 +303,15 @@ export const connectChart = (
   chartGroup: string = CHART_GROUP,
   activateBrushSelect = true
 ) => {
-  // Sync zoom state from any existing chart in the group before connecting
+  // Apply current zoom to this chart without replacing its dataZoom components.
+  // setOption({ dataZoom: [zoomState] }) would replace the array and break slider/inside config.
```
cmatzenbach, line 307: 👍🏻

ui/src/services/formatters.ts
```diff
@@ -125,3 +131,30 @@ export function formatDuration(ms: number, decimals: number = 2): string {
...
+ * produce distinct formatted strings.
+ * @param ms - Duration in milliseconds
+ * @param windowMs - Visible time window width in milliseconds
+ */
+export function formatDurationForWindow(ms: number, windowMs: number): string {
+  const absMs = Math.abs(ms);
+  const resolution = Math.abs(windowMs) / 1000;
+
+  let unitMs: number;
+  if (absMs < 0.001) unitMs = 1e-6;
```
cmatzenbach, line 144: stylistic nit (take it or leave it): this is the perfect situation for a `switch`

cmatzenbach (approved):
Looks great!

PR #29, opened 2026-03-05
chore: add cargo-deny checks for licenses, advisories, and bans

johanpel:


.github/workflows/rust.yml
```diff
@@ -50,6 +52,21 @@ jobs:
...
+    name: cargo-deny
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: rustup toolchain install stable --profile minimal
+      - uses: actions/cache@v4
+        id: cargo-deny-cache
+        with:
+          path: ~/.cargo/bin/cargo-deny
+          key: cargo-deny-0.19.0
```
mbrobbel: you should probably include runner arch in the key

mbrobbel (approved):

.github/workflows/rust.yml
johanpel: https://github.com/rapidsai/quent/pull/29/commits/30a1afc5c8c5bdf6ecac57828184a5c24b4958b3

PR #30, opened 2026-03-05
chore: add nix flake with dev shell and formatter

mbrobbel:
As an alternative to Pixi.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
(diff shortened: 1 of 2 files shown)


johanpel (approved):

johanpel:
@mbrobbel would you still like to merge and use this approach or should we close?

mbrobbel:
> @mbrobbel would you still like to merge and use this approach or should we close?

Yes, let's merge, but let me first remove `protobuf`.

PR #31, opened 2026-03-05
feat: add "active span" to operator

johanpel:
Adds a span of time between the first moment an operator started processing an input, and the latest moment at which an operator finished producing an output (excluding any potential back-pressure) to operators on the UI side.

There may have been gaps in this span in which this operator was not actively using any resources. Thus, this spann of time does NOT repesent e.g. "CPU time" spent, but it can be used to eyeball operator activity from a very very high level in the UI.


examples/simulator/server/ts-bindings/Operator.ts
```diff
@@ -34,4 +35,14 @@ custom_attributes: { [key in string]?: Value | null },
...
  */
-statistics: OperatorStatistics | null, };
+statistics: OperatorStatistics | null, 
+/**
+ * The span of time between the first moment an operator started processing
+ * an input, and the latest moment at which an operator finished producing
+ * an output (excluding any potential back-pressure).
+ *
+ * There may have been gaps in this span in which this operator was not
+ * actively using any resources. Thus, this spann of time does NOT repesent
```
mbrobbel:
```suggestion
 * actively using any resources. Thus, this span of time does NOT repesent
```

domains/query_engine/ui/src/lib.rs
```diff
@@ -169,6 +169,15 @@ pub struct Operator {
     /// These are attributes that are typically gathered after the work
     /// described by an [`Operator`] has completed.
     pub statistics: Option<OperatorStatistics>,
+
+    /// The span of time between the first moment an operator started processing
+    /// an input, and the latest moment at which an operator finished producing
+    /// an output (excluding any potential back-pressure).
+    ///
+    /// There may have been gaps in this span in which this operator was not
+    /// actively using any resources. Thus, this spann of time does NOT repesent
```
mbrobbel:
```suggestion
    /// actively using any resources. Thus, this span of time does NOT represent
```

mbrobbel (approved):

PR #33, opened 2026-03-05
refactor: move timeline config into individual request/response entries

### LINKED ISSUE #27: Bulk endpoint reorg
WIP requirements:

* In order to support bulk queries in the caching logic, `config` values for start/end should be passed on a per resource level. This allows the tiling code to request adjacent periods of time for the same resource instead of multiple resources over the same period of time.
* Requests should not be keyed on resourceId in `entities` object. We need to be able to request the same resource with different parameters, namely operator ID.

johanpel:
Moves the window/bins config of each of the timeline entries in a bulk request into each entry separately.

Fixes #27


johallar (approved):
🙇 tyty

PR #37, opened 2026-03-06
feat(ui): compact timelines

johallar:
Revisit after new slimmer select. 
* Needs small redesign for row labels on the left to fit compact design and look nice

Settings menu and resource type dropdown need makeovers, but POC for configurable timeline height so we can have teeny timelines.

https://github.com/user-attachments/assets/909d9402-42e2-4225-82c0-24b58f7dcedc


johallar:
Will re-do this

PR #51, opened 2026-03-10
perf(ui): consolidate bulk fetching

johallar:
Previously we were keying on resourceId for bulk requests, so couldn't make > 1 request for a resource (one with operator one without)

This creates a separate UI uuid for each request, and maps that back to a resource/operator combo. This allows us to make just one bulk request for all timelines on the page (operator specific and plain old resource timelines)

### DIFF
--- a/ui/src/hooks/useBulkTimelines.ts
+++ b/ui/src/hooks/useBulkTimelines.ts
@@ -23,7 +23,11 @@ import {
   visibleEntriesAtom,
 } from '@/atoms/timeline';
 import { selectedNodeIdsAtom } from '@/atoms/dag';
-import { useBulkTimelineFetch, buildMergedBulkEntries } from './useBulkTimelineFetch';
+import {
+  useBulkTimelineFetch,
+  buildMergedBulkEntries,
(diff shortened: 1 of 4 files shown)


ui/src/hooks/useBulkTimelines.ts
```diff
@@ -142,50 +128,32 @@ export function useBulkTimelines({
...
-            if (entry?.status === 'ok') {
-              const resourceTypeName = getResourceTypeName(newBaseEntries[id]);
-              const key = timelineCacheKey(id, resourceTypeName, operatorId);
-              store.set(timelineDataAtom(key), {
-                data: entry.data,
-                config: entry.config,
-              });
-            }
-          }
+        for (const [id, entry] of Object.entries(response.entries)) {
```
cmatzenbach: This is a repeat of the code above in `useBulkTimelineFetch` so could be made into a helper function to avoid duplicating code, but not a huge deal as it only appears twice. If we need to do similar operations again might be good to make a helper for this

cmatzenbach (approved):
One comment, not a huge deal, everything else looks good

PR #52, opened 2026-03-10
experiment(ui): load adjacent timeline data for smoother UX

johallar:
🔬 Experiment

Load a 1x buffer on each side of the timelines. This allows for a much smoother timeline scrolling/zooming experience unless zooming or panning very fast, in which case it ends up being the same as before (blank until data loaded.

https://github.com/user-attachments/assets/dab2d279-5879-4784-b1d4-f3f10dbc545c

This is a tradeoff because it's not happening in the background, it is happening in the main bulk request so request time could be increased up front on first load. Less impact on future round trips since we have data to work with already in the timeline.

Before (for reference):

https://github.com/user-attachments/assets/148b736f-55b7-40e4-b1e8-d9ecd2ae9187

### DIFF
--- a/ui/src/lib/timeline.utils.ts
+++ b/ui/src/lib/timeline.utils.ts
@@ -21,7 +21,7 @@ import type { TimelineRequest } from '~quent/types/TimelineRequest';
 import type { TaskFilter } from '~quent/types/TaskFilter';
 import type { TimelineConfig } from '~quent/types/TimelineConfig';
 
-const MAX_TIMELINE_BINS = 400;
+const MAX_TIMELINE_BINS = 250;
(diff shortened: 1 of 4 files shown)


ui/src/lib/timeline.utils.ts
```diff
@@ -116,6 +123,40 @@ export function buildBinnedTimelineSeries(
       values: [],
     };
   }
+
+  // Slice to viewport so we display only bins overlapping [viewportSec.start, viewportSec.end]
+  if (viewportSec != null) {
```
johallar, line 128: Find a better way to do this

johallar:
Reopen from new fork

PR #55, opened 2026-03-11
feat(ui): operator timelines

### LINKED ISSUE #26: Operator timelines
* Query bundle to return start/end times per operator
* Per worker show timeline of operators, stacked
* Will need some UI navigation/toggle to change views

johallar:
Fixes #26


ui/src/components/operator-timeline/OperatorGanttChart.tsx
```diff
@@ -0,0 +1,392 @@
...
+    [operators, barLabelTextColor, selectedNodeIds]
+  );
+
+  const gridOptions = useMemo(
+    () => ({
+      ...TIMELINE_SPACING,
+      top: 0,
+      bottom: 0,
+      left: TIMELINE_SPACING.left,
+      right: showYScroll ? TIMELINE_SPACING.right : TIMELINE_SPACING.right,
```
cmatzenbach: Is this a mistake, given that both sides of the ternary have the same result? Should one of them be `left`?

ui/src/components/operator-timeline/OperatorGanttChart.tsx
```diff
@@ -0,0 +1,392 @@
...
+        name: op.label,
+      })),
+    [operators]
+  );
+  const showYScroll = rowCount > MAX_VISIBLE_ROWS;
+  const yAxisZoomEnd = showYScroll ? (MAX_VISIBLE_ROWS / rowCount) * 100 : 100;
+
+  const renderItem = useCallback(
+    (
+      params: {
```
cmatzenbach: If `ECharts` provides actual types for these (`params` and `api`) it would be best to those rather than these inferred types; if not it's ok

ui/src/components/operator-timeline/utils.ts
```diff
@@ -0,0 +1,195 @@
...
+  startTimeNs: bigint,
+  workerId: string
+): OperatorActiveSpanEntry[] {
+  const operators = queryBundle.entities.operators;
+  if (!operators) return [];
+
+  const planIds = new Set(getPlanIdsForWorker(queryBundle.plan_tree, workerId));
+  if (planIds.size === 0) return [];
+
+  const entries: OperatorActiveSpanEntry[] = [];
```
cmatzenbach, line 178: This function and the one above share nearly identical inner loops.  The only differences are the plan filtering predicate and how `planId` is resolved.  We could consider creating a shared helper to eliminate code duplication.

ui/src/components/operator-timeline/utils.ts
```diff
@@ -0,0 +1,195 @@
...
+ * Order is stable (by operator id).
+ */
+export function operatorsWithActiveSpansForWorker(
+  queryBundle: QueryBundle<EntityRef>,
+  startTimeNs: bigint,
+  workerId: string
+): OperatorActiveSpanEntry[] {
+  const operators = queryBundle.entities.operators;
+  if (!operators) return [];
+
```
cmatzenbach, line 174: Should this have the same short-circuit as the function above does - `if (planId == null || planId === '') return [];`?

ui/src/components/operator-timeline/OperatorGanttChart.tsx
johallar: Got rid of the scroll bar when there are lots of bars (it still scrolls though), this used to have math will remove!

ui/src/components/operator-timeline/utils.ts
johallar: This should all be handled by the below 2 lines, the worst case there is an empty set, and then it checks for an empty set to bail out

cmatzenbach (approved):

PR #59, opened 2026-03-16
feat(ui): slimmer inline selects

johallar:
This should allow us more room to display FSM dropdown as well

Screenshot 2026-03-16 at 2 35 28 PM

Screenshot 2026-03-16 at 2 35 31 PM


cmatzenbach (approved):
Much improved, looks great!

PR #60, opened 2026-03-16
feat(ui): multiple fsm handling

johallar:
* When > 1 fsm exists, display another select dropdown to select
* When only 1 exists, show static display of that 1
* Updates the cache key to include fsm type now that it can change  
* Removes resource/group ids from title columns to make room 
* Rename `ResourceTypeSelector` to generic `InlineSelector`

https://github.com/user-attachments/assets/95c6c76d-2dcb-42db-bc4d-f7156d76f2ed

### DIFF
--- a/ui/src/components/resource-tree/ResourceGroupRow.tsx
+++ b/ui/src/components/resource-tree/ResourceGroupRow.tsx
@@ -26,7 +26,9 @@ export const ResourceGroupRow = ({
   onFsmChange,
 }: ResourceGroupRowProps): React.ReactNode => {
   const hasMultipleChildTypes = (availableResourceTypes?.length ?? 0) > 1;
-  const hasMultipleFsms = (availableFsmTypes?.length ?? 0) > 1;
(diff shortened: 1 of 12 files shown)


ui/src/components/resource-tree/ResourceGroupRow.tsx
```diff
@@ -16,24 +21,46 @@ export const ResourceGroupRow = ({
...
     <div>
       <div>
-        <span className="text-xs font-bold">{group.instance_name}</span>
+        <span className="text-sm font-bold">{group.instance_name}</span>
       </div>
-      <div className="text-xs text-muted-foreground">{group.id}</div>
       {hasMultipleChildTypes && selectedType && onTypeChange && availableResourceTypes && (
-        <ResourceTypeSelector
+        <InlineSelector
           id={id}
```
cmatzenbach: I like what you did below appending `-fsm` to the `InlineSelector` `id` field below - could we also do that here? (add something like `-resource-type`)

cmatzenbach (approved):

Issue #61, opened 2026-03-18
Blinking Timelines


johallar:
When selecting an operator to filter by from the DAG, all timelines blink

https://github.com/user-attachments/assets/d4c26d34-cb43-41aa-9431-3e66b54876ff

Caused by consolidating bulk fetches here: https://github.com/rapidsai/quent/pull/51

Suspect the root cause is the jotai timeline cache... may need to rework this to use tanstack query after all. Trying to fix it ended up essentially re-implementing tanstack query's "keepPreviousData"

PR #63, opened 2026-03-18
feat(ui): operator highlight by dimming

johallar:
Screenshot 2026-03-18 at 4 36 01 PM

Screenshot 2026-03-18 at 4 36 32 PM

Blinking while selecting operators is very annoying but have an issue to track that down: https://github.com/rapidsai/quent/issues/61


cmatzenbach (approved):
Looks good!

PR #73, opened 2026-03-25
fix(ui): timeline controller filtered

### LINKED ISSUE #72: Timeline controller filtering to scrubber
Image
Should show full timeline always

johallar:
Fixes: #72 

Timeline controller is filtering itself with datazoom:
Screenshot 2026-03-25 at 2 13 40 PM

This was a cache issue, where the resource timeline was steamrolling the timeline controller data, so the timeline controller only showed the clipped data. TLDR; this was doing way too much trying to use the jotai cache, it already fetches with react query.

Bonus: 
* Clauded up a unit test to prevent regression, fails on main, passes here. QueryResourceTree will probably be one of the hardest to test (see how many mocks were needed), as the data coordinator for all of the timeline tree components.


cmatzenbach (approved):
Works great!

PR #82, opened 2026-03-31
feat(ui): add DAG visual controls

cmatzenbach:
Adds a section above the query plan tree for DAG visual controls. Here are the list of controls added:

- Node color: colors nodes by selected quantitative field
- Edge width: sets edge width by selected quantitative field
- Edge color: colors svg edge shapes by selected quantitative field
- Display field: allows users to pick an additional field to be displayed on the node below the label

This PR also adds an advanced settings popover menu, which can be accessed via the icon in the top right. This panel contains the following fields:

- Node/Edge palette: allows user to change the palette used for node and edge coloring, respectively
- Node label: allows user to change the primary label on each DAG node

Finally, to preserve space, the panel was wrapped in the shadcn `Collapsible` component; users can collapse and expand the panel by clicking the chevron icon next to the panel title.


ui/src/components/query-plan/QueryPlanNode.tsx
```diff
@@ -124,28 +124,62 @@ function resolveOperationType(type: string): OperationType {
...
   const selectedNodeIds = useAtomValue(selectedNodeIdsAtom);
-  const isSelected = selectedNodeIds.has(data.metadata?.rawNode?.id ?? '');
-  const hasSelection = selectedNodeIds.size > 0;
-  const isDimmed = hasSelection && !isSelected;
+  const nodeColoring = useAtomValue(nodeColoringAtom);
+  const nodePalette = useAtomValue(nodeColorPaletteAtom);
+  const operatorId = data.metadata?.rawNode?.id ?? '';
+  const isSelected = selectedNodeIds.has(operatorId);
   const statistics = parseCustomStatistics(data.metadata?.rawNode);
+  const nodeDisplayField = useAtomValue(selectedNodeDisplayFieldAtom);
```
johallar: This seems ripe for a hook, with the idea that maybe it could be used in the table as well

ui/src/components/query-plan/QueryPlanNode.tsx
```diff
@@ -124,28 +124,62 @@ function resolveOperationType(type: string): OperationType {
...
       className={nodeVariants({
         operationType: resolveOperationType(data.operationType),
         selected: isSelected,
-        dimmed: isDimmed,
       })}
+      style={{
+        zIndex: 10,
+        opacity: fieldDimmed ? 0.25 : 1,
+        transition: 'opacity 150ms, background-color 150ms, border-color 150ms',
+        ...(fieldColor && { backgroundColor: fieldColor, borderColor: fieldColor }),
```
johallar: `cn` utility preferred, cleaner IMO

ui/src/components/dag/DAGControls.tsx
```diff
@@ -1,6 +1,136 @@
...
+interface DAGFieldProps {
+  label: string;
+  icon: React.ElementType;
+  options: string[];
+  value: string;
+  setValue: (value: string | null) => void;
+  placeholder: string;
+}
+
+const DAGField = ({ label, icon: Icon, options, value, setValue, placeholder }: DAGFieldProps) => (
```
johallar: Mentioned in DM but would be a nice generic ui/component

ui/src/components/query-plan/QueryPlanNode.tsx
cmatzenbach: Agreed, fixed (also did some additional `className`-based refactoring here to clean things up)

ui/src/services/query-plan/dagFieldProcessing.ts
```diff
@@ -0,0 +1,122 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { DAGNode, DAGEdge, NodeColoring, EdgeWidthConfig, EdgeColoring } from './types';
+import { parseCustomStatistics } from '@/lib/queryBundle.utils';
+import { getActivePalette } from '@/services/colors';
+import { formatDuration } from '@/services/formatters';
+
+function formatBytes(value: number): string {
```
johallar: This duplicates formatting logic from formatters.ts `formatQuantity`, we should re-use formatters logic here

ui/src/services/query-plan/dagFieldProcessing.ts
```diff
@@ -0,0 +1,122 @@
...
+  if (abs >= 1048576) return `${sign}${(abs / 1048576).toFixed(1)} MiB`;
+  if (abs >= 1024) return `${sign}${(abs / 1024).toFixed(1)} KiB`;
+  return `${sign}${Math.round(abs)} B`;
+}
+
+/**
+ * Infer a display formatter for a DAG stat field based on its name.
+ * Returns null when no special formatting applies — callers should fall back to String(value).
+ */
+export function inferFieldFormatter(fieldName: string): ((value: number) => string) | null {
```
johallar: I think the formatting logic may need one more pass, when formatting decimals things can get a little funky:

Image

johallar: It would also be nice to format "plain old" numbers, not sure if we want SI suffixes, but at least with commas etc for readability

Image

ui/src/services/colors.ts
```diff
@@ -288,3 +288,98 @@ export function darkenColor(hex: string, amount: number): string {
...
+  switch (palette) {
+    case 'blue':
+      return blendToColor(59, 130, 246, t, neutral); // blue-500
+    case 'teal':
+      return blendToColor(20, 184, 166, t, neutral); // teal-500
+    case 'purple':
+      return blendToColor(168, 85, 247, t, neutral); // purple-500
+    case 'orange':
+      return blendToColor(249, 115, 22, t, neutral); // orange-500
+    case 'viridis': {
```
johallar, line 352: Viridis legend is not quite looking righ still, it looks like it's interpolating between the first and last color but none between. 

Image

ui/src/services/query-plan/dagFieldProcessing.ts
```diff
@@ -0,0 +1,122 @@
...
+  const valueColor = new Map(uniqueValues.map((v, i) => [v, palette[i % palette.length]]));
+  return {
+    type: 'categorical',
+    colorMap: new Map(entries.map(e => [e.id, valueColor.get(String(e.value))!])),
+    labelMap: new Map(entries.map(e => [e.id, String(e.value)])),
+    categoryMap: valueColor,
+  };
+}
+
+export function formatMetricValue(v: number): string {
```
johallar: Also consolidate with formatters.ts

ui/src/components/query-plan/QueryPlanNode.tsx
```diff
@@ -80,16 +89,33 @@ export const QueryPlanNode = memo(({ data }: { data: QueryPlanNodeData }) => {
...
+          const fmt = inferFieldFormatter(colorField);
+          const formatted =
+            typeof displayValue === 'number' && fmt ? fmt(displayValue) : String(displayValue);
+          return (
+            <div
+              className="text-xs text-center mt-0.5"
+              style={{
+                color: fieldColor
+                  ? isLightColor(fieldColor)
+                    ? 'rgba(0,0,0,0.5)'
```
johallar: nit: I have consts + utils setup for these
```suggestion
                    ? withOpacity(BLACK, 0.5)
```

ui/src/components/query-plan/QueryPlanNode.tsx
```diff
@@ -80,16 +89,33 @@ export const QueryPlanNode = memo(({ data }: { data: QueryPlanNodeData }) => {
           'font-bold': data.operationType === 'stage' || isSelected,
         })}
       >
-        {data.label}
+        {resolvedLabel}
       </DataText>
+      {colorField &&
+        (() => {
+          const displayValue = statistics.find(s => s.key === colorField)?.value ?? null;
```
johallar: nit: This IIFE syntax is a little funky, can these consts just be moved to the main component?

ui/src/components/query-plan/QueryPlanNode.tsx
```diff
@@ -64,14 +72,15 @@ export const QueryPlanNode = memo(({ data }: { data: QueryPlanNodeData }) => {
       onMouseLeave={() => setIsHovered(false)}
       style={
         {
-          borderColor: color,
+          borderColor: activeColor,
           backgroundColor: bgColor,
-          '--glow-color': color,
+          '--glow-color': activeColor,
+          ...(fieldColor && isLightColor(fieldColor) ? { color: '#111827' } : {}),
```
johallar, line 88: nit: Move literal to a descriptive const

ui/src/components/dag/DAGChart.tsx
```diff
@@ -24,17 +26,175 @@ import {
...
+    }
+  } else if (edgeWidthConfig) {
+    const v = edgeWidthConfig.values.get(id);
+    if (v !== undefined) {
+      const fmt = inferFieldFormatter(edgeWidthField ?? '');
+      edgeLabelValue = fmt ? fmt(v) : String(v);
+    }
+  }
+
+  const arrowWidth = strokeWidth * 1.5 + 8;
```
johallar: nit: This component has a lot of magic numbers, descriptive consts would be helpful down the road

johallar:
We might want to make the min width of this resizable panel a bit larger as well to accommodate the new controls:

Screenshot 2026-04-06 at 12 13 03 PM

ui/src/services/query-plan/dagFieldProcessing.ts
cmatzenbach: Good catch - made a bare `formatBytes` function in `formatters.ts` that uses the existing `formatWithPrefix` function without necessitating extra overhead needed to call `formatQuantity`.

ui/src/services/query-plan/dagFieldProcessing.ts
cmatzenbach: Revised to round decimals to three significant digits and also formatted whole numbers so they now use commas - formatting logic added to `formatters.ts` for re-usability.

ui/src/services/colors.ts
cmatzenbach: Forgot about that one, modified it to add in all the color stops from the palette:
image

ui/src/services/query-plan/dagFieldProcessing.ts
cmatzenbach: No longer is use after earlier change, removed

ui/src/components/query-plan/QueryPlanNode.tsx
cmatzenbach: Fixed

ui/src/components/query-plan/QueryPlanNode.tsx
cmatzenbach: This is a good nit, that's bad practice. Moved to a variable outside the jsx

johallar (approved):
🚀 whenever you're ready

ui/src/components/dag/DAGChart.tsx
cmatzenbach: Totally agree, changed all the magic numbers into descriptive const's and added comments for anything that needed additional explanation.

PR #86, opened 2026-03-31
feat(ui): add DAG minimap, centralize operation type coloring

### LINKED ISSUE #83: DAG Minimap
Add minimap for DAG chart

cmatzenbach:
Adds a minimap to the DAG chart, which has two variants for both light and dark mode. Since the same operation type coloring is needed for the minimap like for `QueryPlanNode`, centralized this mapping so it can be used by both components.

Fixes: https://github.com/rapidsai/quent/issues/83


ui/src/components/query-plan/QueryPlanNode.tsx
```diff
@@ -65,84 +33,41 @@ const nodeVariants = cva(
...
+        '--glow-color': color,
+      } as React.CSSProperties}
     >
       {data.hasIncoming && (
         <Handle type="target" position={Position.Top} className="w-2 h-2" style={{ opacity: 0 }} />
       )}
 
       <div
-        className={`text-sm break-words text-center ${isSelected ? 'font-bold' : 'font-normal'}`}
+        className={`text-sm break-words text-center ${data.operationType === 'stage' ? 'font-bold' : isSelected ? 'font-bold' : 'font-normal'}`}
```
johallar: Preferable to use `cn` tailwind utility here with object syntax instead of string interpolation (even though I did the original string interpolation 😅)
```
cn("text-sm break-words text-center font-normal", { "font-bold": data.operationType === 'stage'  || isSelected})
```

ui/src/components/query-plan/QueryPlanNode.tsx
```diff
@@ -65,84 +33,41 @@ const nodeVariants = cva(
...
 export const QueryPlanNode = memo(({ data }: { data: QueryPlanNodeData }) => {
   const selectedNodeIds = useAtomValue(selectedNodeIdsAtom);
   const isSelected = selectedNodeIds.has(data.metadata?.rawNode?.id ?? '');
   const hasSelection = selectedNodeIds.size > 0;
   const isDimmed = hasSelection && !isSelected;
   const statistics = parseCustomStatistics(data.metadata?.rawNode);
+  const [isHovered, setIsHovered] = useState(false);
+
+  const color = OPERATION_TYPE_COLORS[data.operationType] ?? DEFAULT_OPERATION_COLOR;
+  const bgOpacity = isSelected ? '4d' : isHovered ? '38' : '26';
```
johallar: I have a utility `withOpacity` that might make this more readable, easier to change in the future (0-1 instead of hex)

ui/src/components/dag/DAGChart.tsx
```diff
@@ -218,6 +223,15 @@ const FlowLayout = ({
       }}
     >
       <Background />
+      <MiniMap
```
johallar, line 226: I'd consider adding: pannable + zoomable props, i think those add value to the minimap vs just as a visual guide

ui/src/components/dag/DAGChart.tsx
```diff
@@ -218,6 +223,15 @@ const FlowLayout = ({
...
+        nodeStrokeWidth={3}
+        style={{ background: 'hsl(var(--card))' }}
```
johallar: I think a tad smaller is nice: `style={{width: '100', height: '100', background: 'hsl(var(--card))' }}` but up to you

ui/src/components/query-plan/QueryPlanNode.tsx
cmatzenbach: Ah, very helpful. Makes it a lot more readable.

ui/src/components/dag/DAGChart.tsx
cmatzenbach: Added - good idea, these are really cool and useful (and fun to play around with)

johallar (approved):
LGTM whenever you're ready!

ui/src/components/dag/DAGChart.tsx
cmatzenbach: Yeah, after comparing the two I agree, a smaller size is better.  Bumped it up slightly to give more surface area now that we have the pan and zoom controls, `100` was just a bit small for those operations.

PR #91, opened 2026-04-02
feat(ui): url persisted state

johallar:
* Encodes/compresses jotai state as a qs parameter to handle a list of expanded item ids, resource/fsm type selections per row, etc. 
* Persists: selected operator id, selected plan id, zoom range, expanded items, selected resource/fsm types

https://github.com/user-attachments/assets/752f2e8b-d4f5-485b-a26c-18d43b7a3ad8


johallar:
Replaced by https://github.com/rapidsai/quent/pull/162

PR #92, opened 2026-04-06
fix(ui): Timeline controller labels based on axis interval

johallar:
I reduced the precision of these, and they looked bad. Adds a function to base it on axis interval to it _should_ always look good, not repeat numbers

Before:
Screenshot 2026-04-06 at 10 44 26 AM

After:
Screenshot 2026-04-06 at 10 44 03 AM


cmatzenbach (approved):
Looks good to me!

PR #99, opened 2026-04-11
fix(ui): complete missed todo, style operator timelines

johallar:
After DAG controls got merged I had a todo to consolidate node coloring + style the operators in the timeline view the same as they are shown on the DAG chart. This does that.

https://github.com/user-attachments/assets/838491b0-6c11-4ffe-b134-574c4adeb810


cmatzenbach (approved):
Works great!

PR #104, opened 2026-04-13
feat(query-engine): bulk timeline caching

### LINKED ISSUE #8: Tiling/Caching pattern for bulk endpoint
Implement tiling/caching scheme for bulk timeline endpoint

This may require a slight re-design to be able to request multiple time ranges in one request.

cmatzenbach:
- Adds cached_bulk_timeline to TimelineCache, extending the existing single-timeline chunk cache to handle bulk requests
- Bulk responses are split into per-chunk SingleTimelineResponse entries and stored under the same ChunkCacheKey scheme used by single-timeline, so chunks fetched via either endpoint are shared
- On a cache hit (full or partial), cached chunks are recombined and returned without hitting the analyzer; on a full miss, falls back to a single bulk analyzer call and caches the result
- Fixes a bug in both bulk and single-timeline caching where TimelineConfig (containing start/end) was included in the params_hash, causing a cache miss on every pan or zoom since the viewport coordinates changed with each request

**How the caching works**

The engine's full time span is divided into zoom_level equal chunks, where `zoom_level = (engine_duration × 2) / view_duration`. Each chunk is cached independently. On subsequent requests at the same zoom level, only the chunks not already in cache are fetched from the analyzer; the rest are served directly.

**Test plan**

- Cold load: verify logs show hit_count=0 and full fetch on first request     
- Pan left/right: verify subsequent requests show hit_count > 0 for previously loaded chunks
- Zoom in/out slightly within the same zoom level: verify cache hits are preserved                                                                     
- Multiple entries: verify partial cache hits correctly merge cached and freshly-fetched chunks

Fixes: #8

### DIFF
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -147,7 +147,7 @@ impl TimelineCache {
         let mut miss_count = 0u64;
 
         for chunk_idx in first_chunk..=last_chunk {
-            for (key, _entry) in &request.entries {
+            for key in request.entries.keys() {
                 let cache_key = ChunkCacheKey {
                     engine_id,
                     params_hash: entry_hashes[key],
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -156,7 +156,7 @@ impl TimelineCache {
         <A as UiAnalyzer>::TimelineParams: Hash + Eq + Clone,
     {
         let Some(geometry) = compute_chunk_geometry(analyzer, &request)? else {
-            return Ok(analyzer.bulk_resource_timeline(request)?);
+            return Ok(tokio::task::block_in_place(|| analyzer.bulk_resource_timeline(request))?);
         };
 
         let entry_hashes = compute_entry_hashes(&request.entries, &request.app_params);
@@ -263,7 +263,8 @@ impl TimelineCache {
             "bulk timeline cold cache: full fetch"
         );
 
-        let response = analyzer.bulk_resource_timeline(request)?;
+        let response =
+            tokio::task::block_in_place(|| analyzer.bulk_resource_timeline(request))?;
 
         for (key, entry_resp) in &response.entries {
             if let BulkTimelinesResponseEntry::Ok { config, data, .. } = entry_resp {
@@ -346,9 +347,11 @@ impl TimelineCache {
                     })
                     .collect();
 
-            let response = analyzer.bulk_resource_timeline(BulkTimelineRequest {
-                entries: chunk_entries,
-                app_params: app_params.clone(),
+            let response = tokio::task::block_in_place(|| {
+                analyzer.bulk_resource_timeline(BulkTimelineRequest {
+                    entries: chunk_entries,
+                    app_params: app_params.clone(),
+                })
             })?;
 
             for (key, entry_resp) in response.entries {
@@ -467,7 +470,8 @@ impl TimelineCache {
                 app_params: request.app_params.clone(),
             };
 
-            let response = analyzer.single_resource_timeline(chunk_request)?;
+            let response =
+                tokio::task::block_in_place(|| analyzer.single_resource_timeline(chunk_request))?;
             self.chunks.insert(cache_key, response.clone()).await;
             chunk_responses.push(response);
         }
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -12,9 +12,9 @@ use quent_analyzer::Span;
 use quent_query_engine_analyzer::{QueryEngineModel, ui::UiAnalyzer};
 use quent_time::{SpanNanoSec, TimeNanoSec, bin::BinnedSpan, to_nanosecs, to_secs_relative};
 use quent_ui::timeline::{
-    request::{self, BulkTimelineRequest, SingleTimelineRequest, TimelineConfig, TimelineRequest},
+    request::{BulkTimelineRequest, SingleTimelineRequest, TimelineConfig, TimelineRequest},
     response::{
-        self, BulkTimelinesResponse, BulkTimelinesResponseEntry, ResourceTimeline,
+        BulkTimelinesResponse, BulkTimelinesResponseEntry, ResourceTimeline,
         ResourceTimelineBinned, ResourceTimelineBinnedByState, SingleTimelineResponse,
     },
 };
@@ -102,6 +102,10 @@ impl TimelineCache {
         // Determine chunk geometry for this zoom level.
         let zoom_level = determine_zoom_level(view_duration, engine_duration);
         let chunk_duration = engine_duration / zoom_level;
+        debug!(
+            engine_duration,
+            view_duration, zoom_level, "bulk timeline zoom level determined"
+        );
 
         // Find which chunk indices overlap the current viewport.
         let first_chunk =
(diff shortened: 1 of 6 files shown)


domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -53,6 +58,252 @@ impl TimelineCache {
...
+        let engine_duration = engine_span.duration();
+        let epoch = engine_span.start();
+
+        // Bypass caching for degenerate cases.
+        if engine_duration == 0 || request.entries.is_empty() {
+            return Ok(analyzer.bulk_resource_timeline(request)?);
+        }
+
+        // All entries share the same viewport config — grab it from the first.
+        let timeline_config = request.entries.values().next().unwrap().config();
```
johanpel: Small nit: since we've checked `request.entries.is_empty()` above, this `unwrap` is never going to panic, but it is good practice in a codebase trying to be as panic-free as possible to leave a little comment explaining why this unwrap is safe with e.g. `// Safety: unwrap never panics as empty requests returns early above`.

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -53,6 +58,252 @@ impl TimelineCache {
...
+            return Ok(analyzer.bulk_resource_timeline(request)?);
+        }
+
+        // Determine chunk geometry for this zoom level.
+        let zoom_level = determine_zoom_level(view_duration, engine_duration);
+        let chunk_duration = engine_duration / zoom_level;
+        debug!(
+            engine_duration,
+            view_duration, zoom_level, "bulk timeline zoom level determined"
+        );
```
johanpel, line 171: Nit: since this can get quite spammy maybe we should move this to the trace! level

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -53,6 +58,252 @@ impl TimelineCache {
...
+                let canonical = entry.clone().with_config(TimelineConfig {
+                    num_bins: 0,
+                    start: 0.0,
+                    end: 0.0,
+                });
+                let serialized = serde_json::to_string(&(&canonical, &request.app_params))
+                    .map_err(|e| crate::error::ServerError::Cache(e.to_string()))?;
+                let mut hasher = DefaultHasher::new();
+                serialized.hash(&mut hasher);
+                hasher.finish()
```
johanpel: If all of the constituents of the hash key are hashable (which IIUC they are), then we could define a new struct type that captures them and `#[derive(Hash, PartialEq)]` on that type instead of serializing to a JSON string to base the hash on. This would mean adding those bounds to (IIRC):
```
<A as UiAnalyzer>::TimelineGlobalParams,
<A as UiAnalyzer>::TimelineParams,
```
and all concrete types of A, which should be fine.

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -53,6 +58,252 @@ impl TimelineCache {
...
+            }
+        }
+
+        debug!(
+            hit_count,
+            miss_count,
+            zoom_level,
+            n_entries = request.entries.len(),
+            "bulk timeline cache check"
+        );
```
johanpel: Nit: same as above, maybe make this trace! level

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -53,6 +58,252 @@ impl TimelineCache {
...
+
+        // Fast path: cold cache — no chunks were cached, so fetch everything in one bulk call.
```
johanpel: This functions is starting to feel rather lengthy from here. I would like to suggest factoring the stages out into separate functions.

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -53,6 +58,252 @@ impl TimelineCache {
...
+            let params_hash = {
+                // Strip the viewport config (start/end/num_bins) before hashing: start/end vary
+                // with every pan or zoom, and num_bins is already a separate field in ChunkCacheKey.
+                // Only the query-identity fields (resource, filters, app params) should determine
+                // whether two requests map to the same cached chunks.
+                let canonical = entry.clone().with_config(TimelineConfig {
+                    num_bins: 0,
+                    start: 0.0,
+                    end: 0.0,
+                });
```
johanpel: I'm not entirely sure if I'm following why we would want to add these constant properties to the serialized string to hash. Could you please explain?

domains/query_engine/server/src/ui.rs
```diff
@@ -263,13 +263,16 @@ async fn bulk_timelines<A>(
...
     let analyzer = state.analyzers.get(engine_id).await?;
-    let response =
-        tokio::task::spawn_blocking(move || analyzer.bulk_resource_timeline(request)).await??;
-    Ok(Json(response))
+    Ok(Json(
+        state
+            .timelines
+            .cached_bulk_timeline(&*analyzer, engine_id, request)
+            .await?,
+    ))
```
johanpel, line 278: We don't want to run long computationally intensive tasks on the regular async executor threads, as this will potentially block other requests from getting served quickly, also see: https://docs.rs/tokio/latest/tokio/index.html#cpu-bound-tasks-and-blocking-code

johanpel (changes requested):
Looks good overall already, thank you! Some nits and a few things that need to be improved. Most notably, we must retain running timeline building tasks on a `spawn_blocking` thread.

domains/query_engine/server/src/timeline_cache.rs
cmatzenbach: Not going to lie, this is a bit beyond my current rust knowledge, but fed it to claude and it resolved it; do you mind double-checking to ensure that was the right solution? Here's the commit hash: `be8748c778f77427dd414d2dde73f262a3f9936a`

domains/query_engine/server/src/timeline_cache.rs
cmatzenbach: Completely agree, that was way too big.  Refactored so related logic now sits in helper functions.

domains/query_engine/server/src/timeline_cache.rs
cmatzenbach: Good catch - that pattern was confusing. The idea was to hash enough of the entry to uniquely identify "which resource/query this is", but exclude the viewport (start/end/num_bins) because those change on every pan or zoom. I've replaced it with an explicit `EntryParamsKey` enum that names only the identity fields, so the viewport exclusion is structural rather than a zeroing trick. The `canonical` variable and JSON hashing are gone in the latest commits.

domains/query_engine/server/src/ui.rs
cmatzenbach: Good call, I didnt think about that. Talked about this with claude, and went ahead and wrapped all the `analyzer.bulk_resource_timeline` and `analyzer.single_resource_timeline` calls with `tokio::task::block_in_place` — including the degenerate-case fallback in `cached_bulk_timeline` that bypasses chunk geometry entirely.

I used `block_in_place` rather than spawn_blocking because the analyzer is passed as `&A` , so moving it into a spawn_blocking closure would require ownership or an `Arc`. `block_in_place` avoids that by keeping everything on the current thread — it just signals to `tokio` to move other tasks off it while the CPU work runs. Happy to switch to spawn_blocking if you'd prefer the stronger isolation, but that would require threading an `Arc<A>` through to the cache methods.

One question: `split_response_into_chunks` and `combine_chunks` also do `O(n)` work over the response data arrays (slicing and concatenating bin vectors). For large datasets these could also be non-trivial. Should we be wrapping those too, or is the analyzer call the main concern?

domains/query_engine/server/src/timeline_cache.rs
cmatzenbach: All the instances of these were diagnostic logs added while verifying the caching behavior - the plan was to remove them before merge once you'd confirmed things looked right. Happy to drop them all now, or keep the aggregate hit/miss count at trace! if that's useful for future debugging. What would you prefer?

johanpel: No strong preference here, but I'd say it's fine to remove it then

domains/query_engine/server/src/ui.rs
johanpel: The `Analyzer` is already wrapped in an `Arc` when returned from `AnalyzerCache::get`, but we de-reference the `Arc` and create a normal `&` to it immediately here. If you dereference that `Arc` after moving it into `spawn_blocking` (see original code) I think it should be fine, or otherwise simply get `cached_bulk_timeline` to accept an `Arc<Analyzer>`.

Since `split_response_into_chunks` is a part of `cached_bulk_timeline`, if this all runs on a blocking thread, then we don't need to worry.

I don't see `block_in_place` being used but in general if you don't REALLY need it, I wouldn't use it. Just use `spawn_blocking`, especially since we're not worried here about context switching overhead. `block_in_place` causes more work for the runtime because it has to move other tasks away from this thread, see: https://docs.rs/tokio/latest/tokio/task/index.html#block_in_place

PR #106, opened 2026-04-14
feat(ui): replace operator stats popup with fixed panel

### LINKED ISSUE #103: Re-envision DAG tooltip
The current tooltips are too cumbersome and distracting when trying to navigate the DAG chart. Think of another way to display this information - maybe on a panel that updates as you move over different DAG nodes?

cmatzenbach:
Through testing we found the existing operator statistics pop-up to be too intrusive when trying to explore/navigate the DAG.  We looked through several potential designs and decided to re-write it as a panel, fixed to the bottom-left side of the DAG. Some notes:
- Hovering over a node will update the data that is displayed in the panel to show statistics for that specific node
- Hovering over blank space empties the panel
- Clicking on a node to select it will cause that node's statistics to stay in the panel, allowing the user to interact with it
- To dismiss a selected node, click anywhere in the DAG chart

Fixes: #103

### DIFF
--- a/ui/src/atoms/dag.ts
+++ b/ui/src/atoms/dag.ts
@@ -2,8 +2,21 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { atom } from 'jotai';
-import type { NodeColoring, EdgeWidthConfig, EdgeColoring } from '@/services/query-plan/types';
+import type { NodeColoring, EdgeWidthConfig, EdgeColoring, StatValue } from '@/services/query-plan/types';
 import type { ContinuousPaletteName } from '@/services/colors';
+
+export interface InspectedNodeData {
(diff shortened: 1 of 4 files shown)


johallar (approved):
I like! One gotcha that we may run into is if you want to look at a nodes details without actually selecting + filtering timelines by that operator, but it seems natural to do so to me.

cmatzenbach:
> I like! One gotcha that we may run into is if you want to look at a nodes details without actually selecting + filtering timelines by that operator, but it seems natural to do so to me.

Yeah I did think about that, but like you said, I feel this is a logical path - selecting an operator lets you see the details as well as highlighting its resources on the timelines. We can adjust later if people find themselves doing this often without wanting to filter the timelines.

PR #110, opened 2026-04-15
feat(ui): operator table

### LINKED ISSUE #69: Operator Table
Refactor GTC version of operator table for the main app

Thoughts:
* Split out a very generic pivot table component, consider using tanstack table as a base. Everything operated from props + callback props, no knowledge of operators, plans, specific data structure, etc.
* Build an operator table component on top of the generic pivot table, knows about operators/plans, quent API structures etc.

johallar:
Refactoring w claude, should give us a good/better starting place to get this feature into mainline quent:

Introduces a generic pivot table system (GroupedDataTable, PivotedStatTable, useStatGroupTableControls) with an operator-specific implementation of it (OperatorTable + PivotTableToolbar). 

Table supports grouping by levels of query plan types/instances, aggregation modes when aggrgated, bidirectional hover sync with the DAG, sync with DAG settings.

Adds nested routes for /timeline and /operators tabs, replacing the previous Radix Tabs component. Adds supporting atoms (hoveredOperatorId, hoveredStat, etc.) and exports operatorTypeColor and formatWithPrefix utilities.

Fixes #69 

https://github.com/user-attachments/assets/24f90cd9-d413-444f-8307-5b3dcc26660a

### Example Implementation:
I took a simple tabular dataset from kaggle and had claude build me an implementation of the table using this dataset, check it out here:

https://github.com/johallar/quent/pull/1

I think the API is pretty nice actually, most of that example is being able to pick display + pivot columns. 

AI Roundup:

## Summary

This branch introduces a new operator statistics table experience built on reusable grouped/pivot table primitives, and integrates it into a dedicated **Operators** tab under query pages. It also refactors shared toolbar/filter/highlight behavior so Timeline, table, and DAG interactions stay synchronized while keeping responsibilities separated.

### GroupedDataTable

- Added `GroupedDataTable` (`ui/src/components/pivot-table/GroupedDataTable.tsx`) as the generic grouped-table renderer.
- Provides core table mechanics:
  - sorting
  - sticky headers + sticky group columns
  - group rendering modes (`rowSpan` and `compact`)
  - optional virtualization (`@tanstack/react-virtual`)
- Supports renderer injection (`DataHeader`, `GroupCell`, `DataCell`) and shared props for custom behavior/styling.

### PivotedStatTable

- Added `PivotedStatTable` (`ui/src/components/pivot-table/PivotedStatTable.tsx`) as the schema-driven pivot/stat adapter on top of `GroupedDataTable`.
- Handles:
  - row expansion from schema
  - pivoting and aggregation
  - stat formatting and heatmap coloring
  - row/column hover and DAG-linked highlighting
- Uses supporting types/utilities:
  - `ui/src/components/pivot-table/types.ts`
  - `ui/src/components/pivot-table/utils.ts`

### OperatorTable

- Added `OperatorTable` (`ui/src/components/operator-table/OperatorTable.tsx`) as the query-specific implementation.
- Builds operator rows from query entities and custom stats.
- Defines operator grouping schema (`partition`, `parent_item_type`, `parent_item`, `item_type`, `item`).
- Wires selection/highlight interactions with DAG atoms and selected plan context.
- Composes:
  - `QueryToolbar`
  - `PivotTableToolbar`
  - `PivotedStatTable`

## Separation of Responsibility

- **Generic rendering layer:** `GroupedDataTable`
- **Schema/data + pivot semantics layer:** `PivotedStatTable` + pivot `types`/`utils`
- **Feature/domain layer:** `OperatorTable`
- **Control state hook:** `useStatGroupTableControls`
- **Toolbar UI components:** `PivotTableToolbar` and shared `QueryToolbar`

This separation keeps table mechanics reusable while isolating operator-specific logic.

### Hook

Use `useStatGroupTableControls` (`ui/src/components/pivot-table/useStatGroupTableControls.ts`) to manage:

- index enable/disable + reorder
- stat selection + ordering
- aggregation mode (`sum`, `mean`, `min`, `max`, `stdev`)
- derived visible stats and active grouping keys

### Components

- Use `PivotTableToolbar` to render controls from hook outputs.
- Use `PivotedStatTable` with:
  - `rows`
  - `schema` (`PivotedStatTableSchema<T>`)
  - `activeIndices`
  - `visibleStats`
  - optional interaction/color hooks (`onHoverStat`, `getGroupTypeColor`, `getGroupCellHandlers`)

`OperatorTable` is the reference implementation of this pattern.

## Routing / App Shell Updates

- Query page is split into explicit tabs:
  - `/profile/engine/$engineId/query/$queryId/timeline`
  - `/profile/engine/$engineId/query/$queryId/operators`
- Query index route redirects to `timeline`.
- Query layout renders Timeline/Operators tab navigation.

## Interaction Model Changes

- Added consolidated cross-view highlight state:
  - `highlightedNodeIdsAtom`
  - `hoveredStatAtom`
- Table hover drives DAG heatmap/highlight behavior.
- DAG hover drives table highlight/scroll behavior.
- Source-aware highlight state (`dag` vs `table`) prevents interaction loops.

## Dependencies

- Added:
  - `@tanstack/react-table`
  - `@tanstack/react-virtual`


cmatzenbach:
Small UI thing - when you hover over the group by buttons, the cursor is a drag cursor, should make it a `pointer` instead to indicate clickability

cmatzenbach:
The node coloring while hovering over table columns seems inconsistent - see the attached video. I'm assuming when I hover over the `duration_s` column all the values node who have values from .1902 to .1899 should get dark blue highlighting, but you can see it's inconsistent.

Also about halfway through the video you can see when I move the cursor out of the table and still one node on the tree is highlighted - they all should go back to normal coloring once you move out of the pivot table.

https://github.com/user-attachments/assets/64106167-80fd-4bb2-8701-9a1ef20b1179

johallar:
> The node coloring while hovering over table columns seems inconsistent - see the attached video. I'm assuming when I hover over the `duration_s` column all the values node who have values from .1902 to .1899 should get dark blue highlighting, but you can see it's inconsistent.
> 
> Also about halfway through the video you can see when I move the cursor out of the table and still one node on the tree is highlighted - they all should go back to normal coloring once you move out of the pivot table.
> 
>  operator-table-node-coloring.mp4

Agreed on the mouse out... i already added this catch all to try and handle mouse out events. I gotta go back and try again:
https://github.com/rapidsai/quent/pull/110/changes#diff-bed8a1d22ad7f69fd23d70cebd8213afc7969a49ae2da8c2f7ddeb9eafe849afR540-R559

I'm not seeing the coloring inconsistencies though, the hovers look like this for me, is it because something is highlighted so the colors are looking off maybe in the video?

Screenshot 2026-04-24 at 1 56 23 PM
Screenshot 2026-04-24 at 1 56 16 PM

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -0,0 +1,498 @@
...
+      if (gk.key === 'parent_item_type') {
+        return {
+          onMouseEnter: () =>
+            setHighlightState(prev => ({
+              ...prev,
+              ids: itemsByParentType.get(gk.id) ?? null,
+              source: 'table',
+              primaryOperatorId: null,
+            })),
+          onMouseLeave: () =>
```
cmatzenbach: Should add a `prev.source === 'table'` guard here like above, as this could clobber a DAG-sourced highlight

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -0,0 +1,498 @@
...
+      if (gk.key === 'parent_item') {
+        return {
+          onMouseEnter: () =>
+            setHighlightState(prev => ({
+              ...prev,
+              ids: itemsByParentName.get(gk.id) ?? null,
+              source: 'table',
+              primaryOperatorId: null,
+            })),
+          onMouseLeave: () =>
```
cmatzenbach: Same as above, add the `prev.source === 'table'` guard

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -0,0 +1,498 @@
...
+      if (gk.key === 'item_type') {
+        return {
+          onMouseEnter: () =>
+            setHighlightState(prev => ({
+              ...prev,
+              ids: itemsByItemType.get(gk.id) ?? null,
+              source: 'table',
+              primaryOperatorId: null,
+            })),
+          onMouseLeave: () =>
```
cmatzenbach: Same as above, add the `prev.source === 'table'` guard

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -0,0 +1,498 @@
...
+  // operator rows. If the selection is non-empty but matches nothing in the
+  // current sibling-plan scope (e.g. a stage node was selected), fall back to
+  // the unfiltered rows so the table doesn't appear inexplicably empty.
+  const rows = useMemo(() => {
+    if (selectedNodeIds.size === 0) return allRows;
+    const filtered = allRows.filter(r => selectedNodeIds.has(r.itemId));
+    return filtered.length > 0 ? filtered : allRows;
+  }, [allRows, selectedNodeIds]);
+
+  const itemsByParentType = useMemo(() => {
```
cmatzenbach: This and the following two `useMemo` statements are identical except for one field being swapped - probably worth a small helper:
```
  function buildItemIdIndex<T extends keyof OperatorTableRow>(              
    rows: OperatorTableRow[],                                               
    field: T,                                                               
    skipDash = true,                                                        
  ): Map<string, Set<string>> { ... }                                       
```

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -0,0 +1,498 @@
...
+          onMouseLeave: () => {
+            setHighlightState(prev =>
+              prev.source === 'table' && prev.ids?.size === 1 && prev.ids.has(firstItemId)
+                ? { ...prev, ids: null, source: null, primaryOperatorId: null }
+                : prev
+            );
+          },
+        };
+      }
+      if (gk.key === 'parent_item_type') {
```
cmatzenbach: These three conditionals only differ by which lookup map they use. Once we have the `buildItemIdIndex` helper from above, can collapse these to a small lookup table:
```
  const groupMaps: Partial<Record<IndexKey, Map<string, Set<string>>>> = {
    parent_item_type: itemsByParentType,                                    
    parent_item: itemsByParentName,
    item_type: itemsByItemType,                                             
  };
```
Then we can use a single `if (groupMaps[gk.key])` statement to handle all three, reducing code duplication.

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -0,0 +1,498 @@
...
+  }, [rows]);
+
+  const scopeLabelValue = useMemo(() => {
+    for (const row of rows) {
+      if (row.scopeLabel !== '-') return row.scopeLabel;
+    }
+    return 'Current';
+  }, [rows]);
+
+  const indexLabels: Record<string, React.ReactNode> = useMemo(
```
cmatzenbach: I believe we should swap `string` for `IndexKey` in the type here, since we've already been working with that union.

ui/src/components/pivot-table/GroupedDataTable.tsx
```diff
@@ -0,0 +1,335 @@
...
+      )}
+      <div className="flex-1 min-h-0 overflow-auto" ref={scrollRef}>
+        <table className="text-xs border-separate border-spacing-0 relative isolate">
+          <thead className="bg-card">
+            <tr className="border-b border-border">
+              {groupColumnIds.map(columnId => (
+                <th
+                  key={columnId}
+                  ref={el => {
+                    const idx = groupColumnIds.indexOf(columnId);
```
cmatzenbach: No need to build `idx` here - it can be passed directly from the `map` call above (`groupColumnIds.map(columnId, idx) => ...`)

ui/src/components/pivot-table/GroupedDataTable.tsx
```diff
@@ -0,0 +1,335 @@
...
+  const groupHeaderRefs = useRef<Array<HTMLTableCellElement | null>>([]);
+
+  const table = useReactTable({
+    data,
+    columns,
+    state: { sorting },
+    onSortingChange: setSorting,
+    getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel(),
+    getRowId: row => getRowId(row as TRow),
```
cmatzenbach: This would allocate a new function on every render - I think it would be better to just pass `getRowId` directly (if a wrapper is needed, could use `useCallback`, but I don't think it's necessary)

ui/src/components/pivot-table/GroupedDataTable.tsx
```diff
@@ -0,0 +1,335 @@
...
+                          )}
+                        </React.Fragment>
+                      );
+                    }
+
+                    const groupKey = row.groupKeys[col] as
+                      | GroupedDataTableGroupKeyEntry
+                      | undefined;
+                    if (!groupKey) return null;
+                    const isRepeated =
```
cmatzenbach: Reading through, I was worried about this potentially being expensive on larger tables with lots of columns, so asked claude, and here's the feedback it gave. Not critical, but might be worth a potential optimization, will leave that up to you.
```
For a virtualized 1000-row × 5-col table that's tolerable, but it's also exactly the kind           
of work computeRowSpans is doing for the rowSpan path. Worth precomputing a
repeatedFlags: boolean[][] (or similar) once via useMemo keyed on
sortedRows, mirroring how rowSpan mode already memoizes rowSpans (line
101). Then the inner loop just reads repeatedFlags[rowIndex][col].
```

ui/src/components/pivot-table/PivotedStatTable.tsx
```diff
@@ -0,0 +1,707 @@
...
+  );
+  useEffect(() => {
+    setTableStatOrder(prev => {
+      if (prev.length === 0) return resolvedVisibleStats;
+      const visibleSet = new Set(resolvedVisibleStats);
+      const kept = prev.filter(stat => visibleSet.has(stat));
+      const additions = resolvedVisibleStats.filter(stat => !kept.includes(stat));
+      return [...kept, ...additions];
+    });
+  }, [resolvedVisibleStats, resolvedVisibleStatsKey]);
```
cmatzenbach: I don't think `resolvedVisibleStatsKey` is necessary in the deps here

ui/src/components/pivot-table/PivotedStatTable.tsx
```diff
@@ -0,0 +1,707 @@
...
+      : (leftAccentShadow ?? existingBoxShadow);
+  const baseStyle = typeColor
+    ? {
+        ...style,
+        // Opaque tint so scrolled rows never bleed through sticky group cells.
+        backgroundColor,
+        boxShadow: combinedBoxShadow,
+      }
+    : style;
+  const mergedStyle = {
```
cmatzenbach: This doesn't seem necessary, it doesn't do anything - remove and just use `baseStyle` where needed

ui/src/components/pivot-table/PivotedStatTable.tsx
```diff
@@ -0,0 +1,707 @@
...
+      header: stat,
+      enableSorting: true,
+      sortUndefined: 'last',
+      accessorFn: (row: PivotedRow) => getSortValue(row, stat, isAggregating, aggMode) ?? undefined,
+    }));
+    return [...groupCols, ...statCols];
+  }, [activeIndices, effectiveVisibleStats, resolvedIndexLabels, isAggregating, aggMode]);
+
+  const hasSelection = (selectedItemIds?.size ?? 0) > 0;
+  const isSelected = (row: PivotedRow) => [...row.itemIds].some(id => selectedItemIds?.has(id));
```
cmatzenbach: I've seen this pattern a few times now, where we're allocating an array just to call `.some` on it. Probably not a big deal, although these components will likely re-render a lot. Could make a helper to avoid the extra array creations, up to you.
```
function itemHasId (items, target) {
  for (const id of items) {
    if (target.has(id)) return true;
  }
  return false;
}
```

ui/src/components/pivot-table/PivotedStatTable.tsx
```diff
@@ -0,0 +1,707 @@
...
+    };
+  }, [handleTableMouseLeave]);
+
+  useEffect(() => {
+    if (!hoveredItemId) return;
+    const row = visiblePivotedRows.find(r => r.itemIds.has(hoveredItemId));
+    if (!row) return;
+    const el = rowRefs.current.get(row.rowKey);
+    el?.scrollIntoView({ block: 'center', behavior: 'smooth' });
+  }, [hoveredItemId, visiblePivotedRows]);
```
cmatzenbach, line 583: Do we really want to auto-scroll when `visiblePivotedRows` changes?  Trying to think through all the actions that would cause that to change - I get why we would do this when `hoveredItemId` changes, but I think several actions could cause a change in `visiblePivotedRows`, no? If it was just to satisfy the `useEffect` deps, maybe store in a ref. Not 100% sure though, let me know what you think.

ui/src/components/pivot-table/useColumnDragDrop.ts
```diff
@@ -0,0 +1,117 @@
...
+    return () => {
+      window.removeEventListener('keydown', onKeyDown);
+    };
+  }, [draggedId, resetDragState]);
+
+  useEffect(() => clearDragPreview, [clearDragPreview]);
+
+  return {
+    draggedId,
+    getDropTargetPosition: (itemId: string): DropPosition | undefined =>
```
cmatzenbach: This should probably be memoized like the other functions being exported here

ui/src/components/pivot-table/PivotedStatTable.tsx
johallar: I think either `resolvedVisibleStats` or `resolvedVisibleStatsKey` is since this is used in the effect, lemme know if i'm missing something though

ui/src/components/pivot-table/PivotedStatTable.tsx
johallar: This is to satisfy the effects dependencies, otherwise visible pivoted rows could get stale. This should not actually scroll when `visiblePivotedRows` changes however (`hoveredItemId` will be null)

ui/src/components/pivot-table/PivotedStatTable.tsx
johallar: mm yeah, slop, `some` will never mutate the array there's no reason to spread it over a new array.

johallar: Oh it's because `itemIds` is a `Set`

ui/src/components/pivot-table/PivotedStatTable.tsx
cmatzenbach: Yeah I agree `resolvedVisibleStats` is needed, I was just saying to remove `resolvedVisibleStatsKey` from the deps array as it's not doing anything here

ui/src/components/pivot-table/PivotedStatTable.tsx
cmatzenbach: Ok yeah, you're right, it will short-out unless the hovered item has changed.  Nevermind!

ui/src/components/pivot-table/PivotedStatTable.tsx
johallar: Ahh I see, yes

ui/src/components/pivot-table/GroupedDataTable.tsx
johallar: 👍 Easy perf boost by just keeping track of "is repeating" outside of the loop, instead of iterating everything each time.

cmatzenbach (approved):

PR #111, opened 2026-04-15
feat(query-engine): bulk timeline caching

### LINKED ISSUE #8: Tiling/Caching pattern for bulk endpoint

cmatzenbach:
- Adds `cached_bulk_timeline` to `TimelineCache`, extending the existing single-timeline chunk cache to handle bulk requests
- Bulk responses are split into per-chunk `SingleTimelineResponse` entries and stored under the same `ChunkCacheKey` scheme used by single-timeline, so chunks fetched via either endpoint are shared
- On a cache hit (full or partial), cached chunks are recombined and returned without hitting the analyzer; on a full miss, falls back to a single bulk analyzer call and caches the result
- Fixes a bug in both bulk and single-timeline caching where `TimelineConfig` (containing start/end) was included in the `params_hash`, causing a cache miss on every pan or zoom since the viewport coordinates changed with each request

**How the caching works**

The engine's full time span is divided into zoom_level equal chunks, where `zoom_level = (engine_duration × 2) / view_duration`. Each chunk is cached independently. On subsequent requests at the same zoom level, only the chunks not already in cache are fetched from the analyzer; the rest are served directly.

**Test plan**

- Cold load: verify logs show hit_count=0 and full fetch on first request
- Pan left/right: verify subsequent requests show hit_count > 0 for previously loaded chunks
- Zoom in/out slightly within the same zoom level: verify cache hits are preserved
- Multiple entries: verify partial cache hits correctly merge cached and freshly-fetched chunks

Fixes: #8

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1520,7 +1520,6 @@ dependencies = [
  "quent-ui",
  "rust-embed",
  "serde",
- "serde_json",
  "thiserror",
  "tokio",
  "tonic",
--- a/domains/query_engine/server/Cargo.toml
+++ b/domains/query_engine/server/Cargo.toml
@@ -20,7 +20,6 @@ quent-query-engine-ui = { path = "../../../domains/query_engine/ui" }
 quent-time = { path = "../../../crates/time" }
 quent-ui = { path = "../../../crates/ui" }
 serde.workspace = true
-serde_json.workspace = true
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread", "sync"]}
 tonic.workspace = true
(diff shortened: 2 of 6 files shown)


cmatzenbach:
@johanpel I had to open up a new PR because our existing forks died once the repo became public. The last commit is the one replacing `block_in_place` with `spawn_blocking`. Could you check that over and let me know if that was done correctly following your recommendations?

Also I will take out the current debug logs once you have time to test it out. I am wondering though, since we may have additional iterations on this to do things like improve cache performance, should we maybe leave them in and move them somewhere less intrusive (and take them out later once we are satisfied with the improvement)? Is there a way we currently handle debug logs such as these?

johanpel (approved):
LGTM, I think we just need to fix a formatting issue.

Regarding debug logs. I think we should log things that can be interesting for users to help them debug the deployment of the analysis service. As long as that the case its fine to keep debug logs. Things that are super detailed can go to the trace level.

Issue #101, opened 2026-04-13
Trace instrumentation


johanpel:
After codegen #81, add a trace entity type to the stdlib, which should make it easy to add metric streams to an application's instrumentation API.

There is one now already but it would be possible to express this in terms of a model if we allow entities to emit certain events multiple times. The latter is also useful for #100

Issue #100, opened 2026-04-13
Metric stream instrumentation


johanpel:
Add a metric stream entity type to the stdlib, which should make it easy to add metric streams to an application's instrumentation API.

Issue #74, opened 2026-03-26
Bottom-up examples


johanpel:
We have an example of instrumenting the simulator engine, which was implemented first in order to feed UI dev, but is not a "friendly introduction".

We must introduce the modeling approach in baby steps. E.g. start with a memory resource and a simple FSM for an allocation, then do a channel, then a thread and a thread pool, and then show how to combine these things in the domain-specific context of query engines.

johanpel:
While #681 has added some basics, it doesn't give examples of instrumenting some existing code yet, which would be a next step.

PR #112, opened 2026-04-16
feat!: code generation from single model definition for instrumentation and parts of analysis

johanpel:
# Overview

This PR overhauls Quent, specifically the instrumentation and analysis parts, into using a code generation approach in order to finalize the PoC.

This was something I've aspired to have from the start, but so far we've been writing all the boilerplate manually to get an idea of what should be generated in the first place. Now that that is clear, it is time to switch over. So, rather than requiring an application to be instrumented to implementing several layers of boilerplate manually, and keeping it all in sync, including:

- Event type definitions
- Instrumentation API
- Bridge code e.g. for C++ clients
- Built-in (domain-specific) (resource) types
- Event aggregators to produce complete entities

All of the above can now be generated from a single Rust source in which you define your model components annotated with some macros.

# Instrumentation API breaking changes

- This adds FSM handles to the instrumentation API. The handle is stateful, because it adds sequence numbers to FSM events.
- This adds entity handles to the instrumentation API. This handle is also stateful as entity events are for now all still one-shot. Non-FSM multi-event entities will be introduced in #100 or #101.

# Impact

I find the main strengths of this approach to be as follows:

- Compile-time checks of proper use of the modeling framework
- Convenient syntax for quickly changing an application model
- No worries about consistency as everything is derived from a single source

If you care about LoC as well, for the query engine domain, 353 lines of model definitions replace 1,018 lines of hand-written event types, observers, and trait implementations — a 65% reduction. Adding a new model component requires ~15-30 lines of definition instead of ~80-250 lines of boilerplate. The C++ bridge is generated entirely: a 24-line build script produces 272 lines of CXX bridge code.

# User-facing APIs

## Modeling (compile time)

| Construct               | Purpose                                                          |
| ----------------------- | ---------------------------------------------------------------- |
| `resource!`             | Resource definition with optional capacity and init attributes   |
| `entity!`               | Entity definition (single-event, multi-event, or resource group) |
| `state!`                | FSM state with optional attributes and resource usages           |
| `fsm!`                  | FSM definition with states, transitions, entry and exit points   |
| `model!`                | Model composition — collects components and generates event enum |
| `instrumentation!`      | Generates context struct with observer factory methods           |
| `#[derive(Attributes)]` | Metadata generation for event payload and attribute structs      |

## Instrumentation API (run time)

Besides the generated API which completely depends on the model:

| Construct          | Purpose                                                     |
| ------------------ | ----------------------------------------------------------- |
| `Ref<T>`           | Typed entity reference (zero-cost wrapper over UUID)        |
| `usage()`          | Ergonomic construction of resource usages for transitions   |
| `CustomAttributes` | Runtime key-value metadata for dynamic attributes           |
  
# [Example](https://github.com/johanpel/quent/blob/codegen/examples/readme/src/lib.rs)
Example above (click ⬆️) explains those user facing APIs quite well, but in a follow-up we should provide more examples and guidance, this is tracked in #74.

# Minor fixes

- Exporters now create their target directory if it doesn't exist. This would otherwise produce rather... unfamiliar errors to a C++ dev's eye when using the instrumentation API from C++, but is also a good improvement in general.

# Known issues / follow-ups

These aren't critical issues, so I would prefer to address them in follow-ups. I will create issues once this lands.

- `model!` only uses the last path segment as the event enum variant name, so two components with the same type name from different modules (e.g., `engine::Task` and `scheduler::Task`) cannot coexist. An aliasing syntax or automatic disambiguation is needed.
- some codegen functions panic for internal errors instead of returning `Result`. Build script failures will sometimes produce backtraces rather than clean error messages. A future refactor should thread `Result` through the codegen pipeline. 
- FSM handles can be used to restrict transitions at compile time or at the very least warn about illegal transitions at run-time
- `FsmEventsBuilder` could automatically extract `parent_group_id` and `resource_type_name` from the entry transition via `TransitionInfo` traits, the same way it already captures `instance_name` (checked once on first push, O(1)). This would eliminate manual extraction in domain-specific analyzers that currently match on the initializing state just to forward these fields.
- The instrumentation API could generate UUIDs internally instead of requiring callers to supply them. Observer methods and FSM entry calls currently take `id: Uuid` as the first argument, but this could be generated automatically (UUIDv7 for time-ordering) and returned to the caller, simplifying every call site from observer.foo(Uuid::now_v7(), ...)` to `let id =  observer.foo(...)`. I would put this behind some feature flag so you can choose between e.g. uuid v4 or v7. Losing this control may be annoying to some though, so needs to be carefully considered.
- Query engine domain `Engine` needs to be an FSM as timestamp order is relevant
- Analyzer code for simulator no longer needs to convert to RtResources.


dhruv9vats (approved):
Thanks @johanpel, this is an excellent starting point in making this much more ergonomic!

PR #114, opened 2026-04-16
fix(ui): stop timeline zoom once limit is reached

### LINKED ISSUE #46: Limit zoom
Currently you can zoom in infinitely, and bins will get funky, echarts will start stacking them. 

Cap the zoom once we get down to ~10ns bin size

Image

cmatzenbach:
Creates a minimum zoom window and stops timelines from scrolling at that point. Also needed to stop the shift+scroll from panning once you could not zoom any further - this seemed to be the default echarts functionality. Additionally, prevents the browser from catching this event once echarts cannot intercept and using it to zoom in on the webpage. See comments for additional information.

Fixes: #46

### DIFF
--- a/ui/src/lib/timeline.utils.ts
+++ b/ui/src/lib/timeline.utils.ts
@@ -48,6 +48,10 @@ export function nanosToMs(ns: bigint): number {
   return Number(ns / 1_000_000n) + Number(ns % 1_000_000n) / 1_000_000;
 }
 
+/**
+ * Currently static but may be used in the future to prevent sub
+ * nanosecond bin sizes
+ */
 export function getAdaptiveNumBins(): number {
   return MAX_TIMELINE_BINS;
 }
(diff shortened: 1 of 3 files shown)


ui/src/lib/timeline.utils.ts
```diff
@@ -30,9 +30,19 @@ import type { TimelineRequest } from '~quent/types/TimelineRequest';
 import type { TaskFilter } from '~quent/types/TaskFilter';
 import type { TimelineConfig } from '~quent/types/TimelineConfig';
 
-const MAX_TIMELINE_BINS = 400;
+export const MAX_TIMELINE_BINS = 400;
 const LONG_ENTITIES_BIN_MULTIPLIER = 30;
 
+/** Minimum bin duration in nanoseconds — prevents ECharts from stacking bins when zoomed too far. */
+export const MIN_BIN_DURATION_NS = 10;
```
johallar: Crank this to 250 and it's working great!

johallar (approved):
One comment, turn it up to 250 and it's good to go thanks for doing this one!

PR #115, opened 2026-04-17
fix!: Export stdlib entities under their module paths

dhruv9vats:
Exports stdlib entities under their module paths, same as how it is done for the query engine domain entities:  https://github.com/rapidsai/quent/blob/c5508d77ff2a726315ff9d9b9211c4f317868253/domains/query_engine/model/src/lib.rs#L8-L14

This allows consistent re-exports: https://github.com/sirius-db/sirius/pull/648/changes#diff-8f6110d51df61b895aec59a39ab0eb4e79e2e98e7e99ede2a06e31b75d911ad9

### DIFF
--- a/crates/stdlib/src/lib.rs
+++ b/crates/stdlib/src/lib.rs
@@ -8,10 +8,6 @@
 //! - [`Processor`]: unit resource for computation
 //! - [`Channel`]: unidirectional data transfer resource with `bytes` capacity
 
-mod channel;
-mod memory;
-mod processor;
-
-pub use channel::*;
-pub use memory::*;
-pub use processor::*;
+pub mod channel;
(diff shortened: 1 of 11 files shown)


crates/stdlib/src/lib.rs
```diff
@@ -8,10 +8,6 @@
...
-mod channel;
-mod memory;
-mod processor;
-
-pub use channel::*;
-pub use memory::*;
-pub use processor::*;
+pub mod channel;
+pub mod memory;
+pub mod processor;
```
dhruv9vats, line 13: for the reviewer: this is the main change.

johanpel (approved):
:+1: thanks!

PR #119, opened 2026-04-27
feat(ui): replace custom theme hook with echarts theme

### LINKED ISSUE #118: Replace custom echarts theme hook w built in theme api
Didn't know about echarts theme api, use instead of custom theming hook

johallar:
Fixes: #118


cmatzenbach (approved):
Looks great!  Much cleaner now, great improvement

PR #121, opened 2026-04-29
refactor(ui): use context/provider pattern for pivot table intra communication

johallar:
Instead of using a custom hook to stabilize the "custom" header/cell components, use a context/provider pattern to share data between components.

More readable, and context/provider verboseness/uglyness internal to the table components, not part of API.

### DIFF
--- a/ui/src/components/pivot-table/PivotedStatTable.tsx
+++ b/ui/src/components/pivot-table/PivotedStatTable.tsx
@@ -8,11 +8,11 @@ import { GroupedDataTable } from './GroupedDataTable';
 import { cn } from '@/lib/utils';
 import type { AggMode, PivotedRow, HoveredStatInfo, PivotedStatTableSchema } from './types';
 import type {
-  GroupedDataTableGroupKeyEntry,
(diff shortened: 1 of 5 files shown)


ui/src/components/pivot-table/PivotedStatTable.tsx
```diff
@@ -645,53 +569,61 @@ export function PivotedStatTable<TRow>({
...
+          renderGroupHeader={renderGroupHeader}
+          DataHeader={DataHeader}
+          GroupCell={GroupCell}
+          DataCell={DataCell}
+          virtualization={virtualization}
+          groupRenderMode={effectiveGroupRenderMode}
+          stickyGroupColumns={stickyGroupColumns}
+          sorting={sorting}
+          onSortingChange={onSortingChange}
+          getRowRef={rowKey => el => {
```
cmatzenbach: Since it is using refs, we should memoize this function with `useCallback` so that it is not re-created on every render.

cmatzenbach (approved):
Just one comment - everything else looks great!

PR #122, opened 2026-04-29
fix(ui): repair blinking timelines

### LINKED ISSUE #61: Blinking Timelines

johallar:
Resume smoothness, Fixes #61 

https://github.com/user-attachments/assets/ac109317-68dd-473a-a102-f40ece7dda70


cmatzenbach (approved):
Works great (now that we know the bulk commit is messing things up)

PR #130, opened 2026-05-04
fix(ui): overlapping operator timelines

### LINKED ISSUE #116: Operator timeline overlap
Image

Many concurrent operators in the operator timeline can overlap, adjust spacing + scroll behaviors so this doesn't happen.

johallar:
Refactors the size calculations for operator timeline bars so they won't overlap. Simplify the vertical scroll fallback when (too many operators for max height) by just using browser scroll on the container instead of trying to implement in echarts datazoom.

For easy testing just reduce the max height to 100 which won't hold all overlapping operators from the current simulator.

Fixes #116


cmatzenbach (approved):
Approving, but may want to considering copying the nice-looking scrollbar from the query plan selection list rather than using the default HTML one

PR #134, opened 2026-05-04
refactor(ui): ensure only 1 tooltip at a time

### LINKED ISSUE #124: Multiple tooltips
Timelines frequently stuck in a state with multiple open tooltips, ensure one at a time

Image

johallar:
Breaks tooltips out of echarts, this will be crucial later if/when we want to customize these tooltips more (pinned tooltips, etc). 

This also fixes weirdness introduced by using the `connect` echarts API and frees us from that so makes it easier to swap charting libraries in the future if we want. 

Also switches us to the SVG renderer officially, anti aliasing out of hand at high zoom levels. Number of bins shown at any given time has a hard limit, and svgs use less memory than rendering multiple canvases.

Screenshot 2026-05-06 at 11 57 20 AM

After:
Screenshot 2026-05-06 at 11 59 09 AM

Fixes #124


cmatzenbach:
The tooltip is overflowing horizontally from the bounds of the timeline - most noticeable to the left of the timelines. Maybe this is due to the offset being used?

https://github.com/user-attachments/assets/f09a25d7-f822-4e24-90c1-d4a76dd689c4

cmatzenbach (commented):
Code looks good, but may want to address the hover boundaries so we're not overflowing the container.

cmatzenbach (approved):
Works perfectly with the new fix!

PR #139, opened 2026-05-06
fix(analyzer): bulk timeline cache correctness and cold-cache performance parity

cmatzenbach:
This PR fixes two related issues in the bulk timeline caching path:      

1. A correctness bug in the original bulk caching commit (#111) where cached data didn't match its cache-key labels, causing timelines to shift on pan, highlights to drift, and leaf timelines to disagree with their aggregated parents.                                              
2. A performance regression that the correctness fix introduced — cold bulk requests went from one analyzer call to N sequential calls (one per canonical chunk).

**Background: the original bug**                                             
   
The bulk cache uses `ChunkCacheKey { engine_id, params_hash, zoom_level, chunk_idx, num_bins }`. Each chunk represents a canonical span - e.g., on an engine [0, 100] at zoom level 4, chunks are [0,25], [25,50], [50,75], [75,100].

The correctness invariant is: data cached under chunk key `chunk_idx=N` must contain the bins for the canonical span of chunk `N`, regardless of what viewport triggered the fetch. The original cold-fetch path (`cold_fetch_and_cache`) violated this. It:
1. Called `bulk_resource_timeline()` with the user's exact viewport (e.g., [30, 80]).
2. Sliced the viewport response into chunk-sized pieces via `split_response_into_chunks`.                                              
3. Cached each piece under its canonical `chunk_idx` key — but the data inside was sliced at the viewport's bin boundaries, not the chunk's.

**Fix layer 1: canonical chunk fetches**

The first commit (`c25b2ce`) replaces the slice-the-viewport  approach with direct canonical-chunk fetching. Cold and partial paths now both fetch (`canonical_chunk_span`, `num_bins`) configs and cache the responses unaltered.
  
Tests added in `timeline_cache.rs::tests` lock in the correctness invariant:
- `cold_bulk_fetches_canonical_chunks_not_original_viewport` directly asserts canonical spans (25,50), (50,75), (75,100) for a viewport-misaligned request — would fail against the old code.
- `panned_bulk_view_reuses_overlapping_cached_chunks` asserts the cache-reuse path no longer drifts.
- Plus tests for partial-miss merging, error preservation, and operator-filter cache key correctness.                        
                                                                           
`combine_chunks` was also fixed to derive the response span from actual contributing bin boundaries instead of `req_span`, eliminating a smaller related label/data drift when chunk boundaries don't align with the viewport.

**Why c25b2ce alone wasn't enough**
         
Cold cache went from one bulk call to N sequential calls (one per canonical chunk). The dominant cost in `bulk_resource_timeline` is the `for task in self.model.tasks.values()` loop - it scales with engine size, not viewport, and is paid per call. Original #111 explicitly aimed for cold-cache parity with pre-caching; the correctness fix regressed it.

**Fix: chunked analyzer call**

New trait method `UiAnalyzer::bulk_chunked_resource_timeline` accepts `N` time windows per entry and emits `N` outputs per entry. The default impl loops over `bulk_resource_timeline` once per config. `SimulatorUiAnalyzer` overrides with a single-pass version: `N×M` builders (one per (entry, config)), one iteration over `model.tasks`, each usage dispatched to all matching builders via the existing reverse index. Restores cold-cache parity. New `BulkChunkedTimelineRequest` / `BulkChunkedTimelinesResponse` are server-internal - the bulk endpoint shape is unchanged.                                

**Test plan**                                                                

- All 5 `timeline_cache::tests` pass (canonical-span cold fetch, pan reuse, partial-miss merging, operator-filter cache key, error preservation).
- Manually verified in the dev server: no timeline shift on pan, no highlight drift, leaf timelines match aggregated parents.                
                             
**Future considerations / known limitations**                                

1. `long_entities_threshold_s` across chunk boundaries (pre-existing). Threshold is evaluated per builder, i.e. per chunk. An entity that's long globally but short within each chunk window may be classified inconsistently. Not introduced here but worth a separate ticket if the intent is "global duration."
2. Default trait impl is the regressed path. Only `SimulatorUiAnalyzer` overrides it. Any future `UiAnalyzer` impl inherits the N-call behavior unless it also overrides. Doc comment flags this.
3. `bulk_resource_timeline` is becoming vestigial. The cache only calls the chunked method now; the old method survives as a fallback for degenerate inputs (empty entries, zero-duration engine) and as the underpinning of the default trait impl. Follow-up candidate for consolidation.           
4. No automated equivalence test for the simulator override. The simulator analyzer crate has no test fixtures (data/ ndjson files are gitignored), and synthetic event construction is non-trivial. A small committed fixture would unlock a property test ("chunked output ≡ N independent `bulk_resource_timeline calls`") protecting both methods against future drift. Validation here was manual.

### DIFF
--- a/domains/query_engine/analyzer/src/ui.rs
+++ b/domains/query_engine/analyzer/src/ui.rs
@@ -1,12 +1,17 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::HashMap;
+
 use quent_analyzer::AnalyzerResult;
 use quent_events::Event;
 use quent_query_engine_ui as ui;
 use quent_ui::timeline::{
-    request::{BulkTimelineRequest, SingleTimelineRequest},
-    response::{BulkTimelinesResponse, SingleTimelineResponse},
+    request::{BulkChunkedTimelineRequest, BulkTimelineRequest, SingleTimelineRequest},
+    response::{
+        BulkChunkedTimelinesResponse, BulkTimelinesResponse, BulkTimelinesResponseEntry,
+        SingleTimelineResponse,
+    },
 };
 use uuid::Uuid;
 
@@ -65,4 +70,48 @@ pub trait UiAnalyzer {
         &self,
         request: BulkTimelineRequest<Self::TimelineGlobalParams, Self::TimelineParams>,
     ) -> AnalyzerResult<BulkTimelinesResponse>;
+
+    /// Return chunked bulk timelines: multiple time windows per entry.
+    ///
+    /// The default implementation falls back to one `bulk_resource_timeline`
+    /// call per config — correct, but pays the per-call setup cost N times.
+    /// Implementors should override this to amortize expensive per-call work
+    /// (e.g. iterating every task in the model) across all configs in a
+    /// single pass.
+    fn bulk_chunked_resource_timeline(
+        &self,
+        request: BulkChunkedTimelineRequest<Self::TimelineGlobalParams, Self::TimelineParams>,
+    ) -> AnalyzerResult<BulkChunkedTimelinesResponse>
+    where
+        Self::TimelineGlobalParams: Clone,
+        Self::TimelineParams: Clone,
+    {
+        let mut entries: HashMap<String, Vec<BulkTimelinesResponseEntry>> = request
+            .entries
+            .keys()
+            .map(|k| (k.clone(), Vec::with_capacity(request.configs.len())))
+            .collect();
+
+        for config in &request.configs {
+            let inner_entries = request
+                .entries
+                .iter()
+                .map(|(k, e)| (k.clone(), e.clone().with_config(config.clone())))
+                .collect();
+            let mut response = self.bulk_resource_timeline(BulkTimelineRequest {
+                entries: inner_entries,
+                app_params: request.app_params.clone(),
+            })?;
+            for (k, slot) in entries.iter_mut() {
+                let entry = response.entries.remove(k.as_str()).unwrap_or_else(|| {
+                    BulkTimelinesResponseEntry::Error {
+                        message: format!("missing entry '{k}' in chunked fallback"),
+                    }
+                });
+                slot.push(entry);
+            }
+        }
+
+        Ok(BulkChunkedTimelinesResponse { entries })
+    }
 }
--- a/domains/query_engine/analyzer/src/ui.rs
+++ b/domains/query_engine/analyzer/src/ui.rs
@@ -96,7 +96,7 @@ pub trait UiAnalyzer {
             let inner_entries = request
                 .entries
                 .iter()
-                .map(|(k, e)| (k.clone(), e.clone().with_config(config.clone())))
+                .map(|(k, e)| (k.clone(), e.clone().with_config(*config)))
                 .collect();
             let mut response = self.bulk_resource_timeline(BulkTimelineRequest {
                 entries: inner_entries,
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -285,8 +285,7 @@ impl TimelineCache {
             .flat_map(|(chunk_idx, keys)| keys.iter().map(move |k| (k.clone(), *chunk_idx)))
             .collect();
 
-        let mut miss_entry_keys: std::collections::HashSet<String> =
-            std::collections::HashSet::new();
+        let mut miss_entry_keys: HashSet<String> = HashSet::new();
         for keys in chunk_misses.values() {
             for k in keys {
                 miss_entry_keys.insert(k.clone());
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -25,7 +25,7 @@ use quent_ui::timeline::{
 use tracing::{debug, trace};
 use uuid::Uuid;
 
-use crate::error::ServerResult;
+use crate::error::{ServerError, ServerResult};
 
 /// Target number of chunks visible in the current view range.
 const TARGET_CHUNKS_PER_VIEW: u64 = 2;
@@ -338,6 +338,14 @@ impl TimelineCache {
         .await??;
 
         for (key, per_chunk) in response.entries {
+            if per_chunk.len() != miss_chunk_indices.len() {
+                return Err(ServerError::Cache(format!(
+                    "chunked analyzer returned {} slots for entry '{}', expected {}",
+                    per_chunk.len(),
+                    key,
+                    miss_chunk_indices.len()
+                )));
+            }
             for (slot_idx, entry_resp) in per_chunk.into_iter().enumerate() {
                 let chunk_idx = miss_chunk_indices[slot_idx];
                 // Skip slots that weren't actually a miss for this entry —
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -356,10 +356,8 @@ impl TimelineCache {
                         entry_chunks.entry(key.clone()).or_default().push(single);
                     }
                     BulkTimelinesResponseEntry::Error { message } => {
-                        error_entries.insert(
-                            key.clone(),
-                            BulkTimelinesResponseEntry::Error { message },
-                        );
+                        error_entries
+                            .insert(key.clone(), BulkTimelinesResponseEntry::Error { message });
                     }
                 }
             }
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -277,13 +277,19 @@ impl TimelineCache {
             return Ok(());
         }
 
-        // Set of (entry_key, chunk_idx) pairs that actually missed. Used to
-        // discard responses for pairs the analyzer recomputed redundantly when
-        // entries have non-uniform miss patterns (rare).
-        let miss_pairs: std::collections::HashSet<(String, u64)> = chunk_misses
-            .iter()
-            .flat_map(|(chunk_idx, keys)| keys.iter().map(move |k| (k.clone(), *chunk_idx)))
-            .collect();
+        // For each entry that missed at least one chunk, the set of chunk indices
+        // it missed. Used to discard responses for pairs the analyzer recomputed
+        // redundantly when entries have non-uniform miss patterns (rare).
+        let mut entry_miss_chunks: HashMap<String, std::collections::HashSet<u64>> =
+            HashMap::new();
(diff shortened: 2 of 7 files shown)


domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -246,137 +245,121 @@ impl TimelineCache {
...
+        // Set of (entry_key, chunk_idx) pairs that actually missed. Used to
+        // discard responses for pairs the analyzer recomputed redundantly when
+        // entries have non-uniform miss patterns (rare).
+        let miss_pairs: std::collections::HashSet<(String, u64)> = chunk_misses
+            .iter()
+            .flat_map(|(chunk_idx, keys)| keys.iter().map(move |k| (k.clone(), *chunk_idx)))
+            .collect();
+
+        let mut miss_entry_keys: std::collections::HashSet<String> =
+            std::collections::HashSet::new();
```
dhruv9vats, line 294:
```suggestion
        let mut miss_entry_keys: HashSet<String> = HashSet::new();
```

nit: including `std::collections::HashSet` is typical.

domains/query_engine/analyzer/src/ui.rs
```diff
@@ -65,4 +70,48 @@ pub trait UiAnalyzer {
...
+            .entries
+            .keys()
+            .map(|k| (k.clone(), Vec::with_capacity(request.configs.len())))
+            .collect();
+
+        for config in &request.configs {
+            let inner_entries = request
+                .entries
+                .iter()
+                .map(|(k, e)| (k.clone(), e.clone().with_config(config.clone())))
```
dhruv9vats: nit: it may be useful to derive `Copy` for trivial types in the future.

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -246,137 +245,121 @@ impl TimelineCache {
...
+                    start: to_secs_relative(chunk_start, ctx.geometry.epoch),
+                    end: to_secs_relative(chunk_end, ctx.geometry.epoch),
+                }
+            })
+            .collect();
+
+        let chunked_entries: HashMap<String, TimelineRequest<<A as UiAnalyzer>::TimelineParams>> =
+            miss_entry_keys
+                .iter()
+                .map(|k| (k.clone(), request.entries[k].clone()))
```
dhruv9vats:
```suggestion
                .into_iter()
                .map(|k| {
                    let request = request.entries[&k].clone();
                    (k, request)
                })
```
nit: typical to consume a container once it is not longer needed.

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -246,137 +245,121 @@ impl TimelineCache {
...
+                .collect();
+
+        let a = Arc::clone(&analyzer);
+        let app_params_clone = request.app_params.clone();
+        let response = tokio::task::spawn_blocking(move || {
+            a.bulk_chunked_resource_timeline(BulkChunkedTimelineRequest {
+                entries: chunked_entries,
+                configs,
+                app_params: app_params_clone,
             })
```
dhruv9vats, line 341: Not something to change rn, but previously we have had seen compute heavy analysis parts to start dominating API calls. Especially once we start dealing with finer grained things like movement data. At which point something like this might come in handy: https://ryhl.io/blog/async-what-is-blocking/ . But that's usually if the compute heavy parts are ... really heavy, as which point we would start using some parallelism.

Just flagging this here now, so our future selves can find it more easily -- rather than having to drill down in flamecharts. 🙂

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -823,8 +729,12 @@ fn combine_chunks(
             }
         }
 
+        let combined_span = SpanNanoSec::try_new(
+            combined_start.unwrap_or(req_span.start()),
+            combined_end.unwrap_or(req_span.end()),
+        )?;
```
dhruv9vats, line 754:
```suggestion
        let combined_span = SpanNanoSec::try_new(
            bin_start,
            bin_end,
        )?;
```
just directly construct combined span? since combined_start and combined_end will always be `None` to begin with and `combined_start = Some(combined_start.map_or(bin_start, |start| start.min(bin_start)))` will always return `bin_start`? Or maybe I am missing something here?

Ditto above?

domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -842,6 +752,20 @@ fn combine_chunks(
...
+    chunk: &SingleTimelineResponse,
+    start_idx: usize,
+    end_idx: usize,
+    epoch: TimeNanoSec,
+) -> (TimeNanoSec, TimeNanoSec) {
+    let chunk_start = epoch + to_nanosecs(chunk.config.span.start());
+    let bin_duration_ns = to_nanosecs(chunk.config.bin_duration);
+    (
+        chunk_start + start_idx as u64 * bin_duration_ns,
+        chunk_start + end_idx as u64 * bin_duration_ns,
```
dhruv9vats:
```suggestion
        chunk_start + (start_idx as u64 * bin_duration_ns),
        chunk_start + (end_idx as u64 * bin_duration_ns),
```

μNit

dhruv9vats (approved):
Thanks @cmatzenbach!

examples/simulator/analyzer/src/lib.rs
```diff
@@ -512,6 +533,177 @@ impl UiAnalyzer for SimulatorUiAnalyzer {
...
+        // span internally, so out-of-window usages are no-ops.
+        for task in self.model.tasks.values() {
+            let task_operator_id = task.operator_id();
+            for usage in task.usages() {
+                let resource_id = usage.resource_id();
+                if let Some(builder_indices) = plain_index.get(&resource_id) {
+                    for &builder_idx in builder_indices {
+                        let builder = &plain_builders[builder_idx];
+                        if builder
+                            .4
```
dhruv9vats: usually preferable to make such types `struct`s.

examples/simulator/analyzer/src/lib.rs
```diff
@@ -512,6 +533,177 @@ impl UiAnalyzer for SimulatorUiAnalyzer {
...
+            }
+        }
+
+        // Reassemble per-entry Vec aligned with `request.configs` order. Slots
+        // start as `None` and must all be filled by the end — every (entry,
+        // config_idx) had a builder, and every builder produces an `Ok`.
+        let mut slots: HashMap<String, Vec<Option<BulkTimelinesResponseEntry>>> = request
+            .entries
+            .keys()
+            .map(|k| (k.clone(), (0..n_configs).map(|_| None).collect()))
```
dhruv9vats, line 666:
```suggestion
            .map(|k| (k.clone(), vec![None; n_configs]))
```

examples/simulator/analyzer/src/lib.rs
```diff
@@ -512,6 +533,177 @@ impl UiAnalyzer for SimulatorUiAnalyzer {
...
+
+        for (entry_id, config_idx, builder, _, _) in plain_builders {
+            let built = builder.build();
+            let config = built.config.try_to_secs_relative(epoch)?;
+            let resp = BulkTimelinesResponseEntry::Ok {
+                message: String::new(),
+                config,
+                data: self.timeline_to_ui(built, epoch)?,
+            };
+            slots.get_mut(&entry_id).unwrap()[config_idx] = Some(resp);
```
dhruv9vats:
```suggestion
            slots
            	.get_mut(&entry_id)
            	.expect("known key, instead found unknown key {}", entry_id)[config_idx] = Some(resp);
```
nit. Ditto below.

examples/simulator/analyzer/src/lib.rs
```diff
@@ -512,6 +533,177 @@ impl UiAnalyzer for SimulatorUiAnalyzer {
...
+        }
+        for (entry_id, config_idx, builder, _, _) in per_state_builders {
```
dhruv9vats: Preferable to have a struct at this size.

domains/query_engine/server/src/timeline_cache.rs
cmatzenbach: No you're right, that was pointless.  Revised as suggested.

domains/query_engine/analyzer/src/ui.rs
cmatzenbach: Added `Copy` to `TimelineConfig`'s derive list

examples/simulator/analyzer/src/lib.rs
cmatzenbach: Ok I see, yeah that makes sense. Replaced the two type aliases with proper structs - `PlainBuilderSlot` and `PerStateBuilderSlot`; update all code utilizing these as well.

examples/simulator/analyzer/src/lib.rs
cmatzenbach: 👍🏻 Taken care of - addressed in the struct conversion on the comment above.

examples/simulator/analyzer/src/lib.rs
cmatzenbach: Very cool! Newbie me didn't realize you could do that

cmatzenbach: Ok on second thought, I reverted that as it didn't work. `BulkTimelinesResponseEntry::Ok` carries the response data (nested `HashMap<String, Vec<f64>>` etc.), so we deliberately don't derive `Clone` on it - it would mean making the whole response graph cheaply cloneable. Figured it's not worth it for one line, if you disagree let me know though.

examples/simulator/analyzer/src/lib.rs
```diff
@@ -512,6 +533,177 @@ impl UiAnalyzer for SimulatorUiAnalyzer {
...
+                    per_state_builders.push((
+                        entry_id.clone(),
+                        config_idx,
+                        ResourceTimelineByKeyBuilder::try_new(
+                            resource_type,
+                            entry_config,
+                            long_entities_threshold,
+                        )?,
+                        resource_id_filter.clone(),
+                        task_filter.clone(),
```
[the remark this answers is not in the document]
cmatzenbach: Applied - wrapped `resource_id_filter` in `Arc<HashSet<Uuid>>` so the N config slots for an entry share a single allocation. Left `task_filter` as plain `TaskFilter` since it's tiny (`Option<Uuid>` plus a few flags) and still read in the dispatch loop. The non-chunked bulk_resource_timeline impl above wasn't touched in this PR but has the same pattern; could be a follow-up to do the same to `BulkEntryPrep` itself.

johanpel (commented):
I've tried this out manually. It looks good to me overall.

This is probably unrelated to this specific PR, but one thing seems to be able to get out of sync arre the operator spans:

To reproduce try this:
```
cargo run -p quent-simulator -- --num-query-groups 1 --num-queries 1 --num-workers 1 --num-tasks 1 --num-threads 1 --exporter ndjson
```

Then zoom in to some part where an operator span lives as well, e.g.:

Image

if I proceed to zoom in on the operator spans themselves (it doesnt work on the binned timeline parts), I get this:

Image

PR #128, opened 2026-05-01
refactor!: exploration: improve modeling and code generation infrastructure, DO NOT MERGE

johanpel:
This PR tracks the proposal for #191

PR #140, opened 2026-05-06
feat!: Add codegen for pyo3 model bindings

wence-:
Building on the changes in #112, we can now generate Python bindings for a quent model in a similar manner to the C++ bindings.

The example and query_engine domain tests show example usage. I use maturin to build the pyo3 module. The generation phase generates both a pyo3 module and a type stub so that the installed wheel provides useful information in IDEs. The type stub generation can probably be removed once pyo3-introspection matures sufficiently.


crates/codegen/src/common.rs
```diff
@@ -0,0 +1,140 @@
...
+pub(crate) fn to_pascal_case(s: &str) -> String {
+    s.to_case(Case::Pascal)
+}
+
+/// Convert a model-provided Rust identifier into the Python identifier spelling
+/// exported by generated bindings.
+///
+/// The model definition language supplies ordinary Rust identifiers, so the
+/// only Python-specific conflict we handle here is a Python reserved word.
+pub(crate) fn py_export_name(name: &str) -> String {
```
johanpel: Nit: since this is Python-specific, let's move it into the `pyo3_bridge` module

examples/python-integration/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+This example exposes the same Rust model used by `examples/readme` and
+`examples/cpp-integration` as a Python extension module generated with PyO3.
+
+Build and install the extension into the active Python environment:
+
+```bash
+cd examples/python-integration
+maturin develop
+python main.py
```
johanpel, line 11: We have a root `pixi.toml` to provide the development environment into which I think we should add the required deps:
```
python = ">=3.11"
maturin = ">=1.10,<2"
```
This allowed me to run this, but with `--uv` instead:
```
maturin develop --uv
```

examples/python-integration/README.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+The generated module is named `quent_readme`. Its schema comes from the normal
+Rust `model!` / `entity!` / `fsm!` definitions in `examples/readme`; the bridge
+crate only asks `quent-codegen` to emit PyO3 wrappers from `AppModel::build`.
+The same build script also writes `quent_readme.pyi` so the wheel includes type
+stubs for IDEs and type checkers.
```
johanpel: Could we please add this file to a `.gitignore`?

examples/python-integration/main.py
```diff
@@ -0,0 +1,88 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from pathlib import Path
+
+import quent_readme as quent
+
+
+def main() -> None:
+    output_dir = Path("./events")
+    cluster_id = quent.now_v7()
```
johanpel, line 11: Nit: I think it would be nice to be able to use Python's builtin [UUID](https://docs.python.org/3/library/uuid.html) type. Can be a follow-up.

domains/query_engine/tests/python/test_query_engine.py
johanpel, line 1: Let's add this test to CI

domains/query_engine/tests/python/README.md
```diff
@@ -0,0 +1,11 @@
+# Query Engine Python Integration Test
+
+This test exposes the query engine domain model as a Python extension module
+generated with PyO3. It mirrors the C++ integration test under
+`domains/query_engine/tests/cpp`.
+
+```bash
+cd domains/query_engine/tests/python
+maturin develop
+python test_query_engine.py
```
johanpel, line 10: Same comment here as for the Python example README

johanpel (commented):
This is an exciting addition to the code generation feature, thank you!

I've tried the example and it works well. There are some minor considerations in the comments below.

One question I have is about `entry_state_on_handle`. I'm not sure why this is needed, as it implies we can create FSM handles that don't (yet, or even never) emit events. Is this needed to construct a bunch of handles first, but defer emitting events to some later moment?

examples/python-integration/README.md
johanpel: I think this should also solve the Rust check issue seen in CI

crates/codegen/src/common.rs
wence-: Done.

examples/python-integration/main.py
wence-: Do you need v7 uuids, since those are only generated/supported by Python 3.14 and above.

examples/python-integration/main.py
```diff
@@ -0,0 +1,88 @@
...
+    context = quent.Context(cluster_id, "ndjson", str(output_dir))
```
wence-, line 12: TODO: followup, support the other serialisation formats.

examples/python-integration/README.md
wence-: So I guess there are a few options here:

- commit the generated file (bad, gets out of date), but someone browsing `main.py` before building gets the type stubs
- add to gitignore
- figure out a way so that this temporary file is only built into the build script, not written to the current directory.

Of the three, I think the last is best so I'll try and get that to work.

wence-:
> One question I have is about `entry_state_on_handle`. I'm not sure why this is needed, as it implies we can create FSM handles that don't (yet, or even never) emit events. Is this needed to construct a bunch of handles first, but defer emitting events to some later moment?

The `TaskHandle` in the `example` FSM advertises a `queued` method that one can access in rust. I was trying to make that available in the python and C++ bindings.

The macro-expansion of `TaskHandle` contains in its impl block, amongst other things`:

```rust
    pub fn queued(
        &mut self,
        instance_name: &str,
        index: u64,
        worker: Ref<Worker>,
        queue: Option<quent_model::Usage<Queue>>,
    ) {
        let state = Queued {
            instance_name: instance_name.to_string(),
            index,
            worker,
            queue,
        };
        self.emit_transition(TaskTransition::from(state));
    }
    pub fn computing(
        &mut self,
        thread: Option<quent_model::Usage<Thread>>,
        memory: Option<quent_model::Usage<MemoryPool>>,
    ) {
        let state = Computing { thread, memory };
        self.emit_transition(TaskTransition::from(state));
    }
```

So a handle can transition from queued to computing, but also from computing to queued if I understand things correctly?

wence-:
Removed the entry_state_on_handle flag for now

examples/python-integration/README.md
wence-: Done.

domains/query_engine/tests/python/test_query_engine.py
wence-: I migrated it to pytest and hopefully hooked up CI properly.

examples/python-integration/README.md
wence-: Did the last

examples/python-integration/main.py
johanpel: Ah I see. Let's just keep this for now then. There are no restrictions on the UUID version so v4 is also fine.

FWIW for #128 I'm actually thinking of moving UUID generation into the implementation of the instrumentation API, with a globally configurable style at either compile time or run time (probably either v4 or v7). You would still be able to read it from handles though.

johanpel:
> The `TaskHandle` in the `example` FSM advertises a `queued` method that one can access in rust. 

Hmm, this method shouldn't exist on the handle.

> So a handle can transition from queued to computing, but also from computing to queued if I understand things correctly?

No, this should not be possible because that transition is not declared as possible. This looks like a bug in the proc macro to me, made an issue here: https://github.com/rapidsai/quent/issues/144

In #128 I'm taking this a bit further where I'm working on generating a Rust instrumentation API following the typestate pattern. This will prevent such illegal transitions at compile time. I'm sure we can do something similar for the generated C++ code at some point.

My Python is very basic so not sure if we could get something like type hints to help guard against misuse when linting.

wence-:
> > 
> 
> No, this should not be possible because that transition is not declared as possible. This looks like a bug in the proc macro to me, made an issue here: #144

It's possible that I am misreading the generated code.

> In #128 I'm taking this a bit further where I'm working on generating a Rust instrumentation API following the typestate pattern. This will prevent such illegal transitions at compile time. I'm sure we can do something similar for the generated C++ code at some point.
> 
> My Python is very basic so not sure if we could get something like type hints to help guard against misuse when linting.

Typestate patterns can definitely be applied, but you can't consume the input object so there's nothing stopping you calling the same method (or a different one) again on the input. We can probably do something where the type stubs provide the same typestate transitions as the rust, and then when calling a transition you leave the consumed object in some kind of "moved-from" state that is recognisable and can raise an error at runtime if you try and use it again. OTOH maybe that is too much effort and you just do the typing.

wence-:
I think I glued everything together properly and managed to squash the clippy warnings. It _is_ possible to not have to do `allow-needless-borrow` in the generated pyo3 code but the amount of additional bookkeeping I had to do didn't make it worth it to my mind.

wence-:
Ready for another look @johanpel

johanpel (approved):
This is a great addition to the project. I'm very happy to see we can now add instrumentation in Python too. Thank you!

Added a few labels to get the final ✅ on CI checks, this is good to go.

FYI I've also added this issue to think about moving forward: https://github.com/rapidsai/quent/issues/155

wence-:
Thanks, I don't have any merge powers here so you'll have to do the honours :)

PR #141, opened 2026-05-06
refactor(ui): trigger dev tools with env var

### LINKED ISSUE #137: Default off for tanstack router/query dev tools
At the beginning folks will probably be running a setup more like our dev setup, so we should hide these dev tools badges by default unless explicitly turned on.

johallar:
Currently these are on for all dev environments, but they are pretty intrusive for how frequently they are needed. Push them back behind an env variable setup for debugging.

Fixes #137

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -167,10 +167,26 @@ pnpm dlx shadcn@latest add dropdown-menu
 
 ## Development Tools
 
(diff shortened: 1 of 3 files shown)


cmatzenbach (approved):
Hallejuah 🙌🏻

PR #143, opened 2026-05-06
feat(ui): dynamic operator coloring

### LINKED ISSUE #136: Color dag types dynamically
Currently dag node types are still hardcoded, we need to instead use the `unique_operator_names` array I believe to assign colors from the query bundle response.

cmatzenbach:
Replaces the old system of coloring by operator type lookups with dynamic, hash-based coloring with probe-forwarding to reduce collisions.

Fixes: #136


ui/packages/@quent/utils/src/colors.ts
```diff
@@ -280,28 +280,38 @@ export function isLightColor(hex: string): boolean {
...
+  return OPERATOR_PALETTE[index]!;
+}
+
+export function buildOperatorColorMap(operatorTypes: string[]): Map<string, string> {
+  const sorted = [...new Set(operatorTypes.map(t => t.toLowerCase()))].sort();
+  const used = new Set<number>();
+  const map = new Map<string, string>();
+  for (const type of sorted) {
+    let index = hashString(type) % OPERATOR_PALETTE.length;
+    while (used.has(index)) index = (index + 1) % OPERATOR_PALETTE.length;
```
johallar: This will loop forever if we use every color in the palette I think?

It might be nice to genericize the hash + walk algorithm, which is used in `getColorForKey`, and use it in both places.

johallar (changes requested):
Possible infinite looping, some code to be potentially consolidated, otherwise looking good!

ui/packages/@quent/utils/src/colors.ts
cmatzenbach: Ah good catch, don't know how I missed that one.  Great idea on genericizing as well, went ahead and made that into a reusable helper, now called in both places.

johallar (approved):

PR #145, opened 2026-05-07
fix(ui): theme change resets zoom range

### LINKED ISSUE #131: Sections of timelines disappear when switching themes
When you zoom in on the timelines and then switch themes, part of the timelines will disappear.  See attached screencast.

https://github.com/user-attachments/assets/2cfd289d-1b01-4ba0-bd81-9534ff6e7811

Note: Tested on a build with bulk timelines PR reverted, so the issue is unrelated to that.

johallar:
When changing theme all timeline charts re-render so getting zoom range from chart instances via DOM is not viable strategy.

Consolidates a couple pieces of zoom handling logic:
1. Use the `zoomRangeAtom` as single source of truth
2. Creates a `useChartConnect` to capture common `connect` logic between all of our chart types
Fixes #131

### DIFF
--- a/ui/packages/@quent/components/src/index.ts
+++ b/ui/packages/@quent/components/src/index.ts
@@ -98,7 +98,6 @@ export {
   buildTimelineMarks,
   collectVisibleEntries,
   getAdaptiveNumBins,
-  getChartGroupZoomState,
   getFsmTypeName,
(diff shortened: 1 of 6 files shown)


cmatzenbach (approved):
Works great!

PR #148, opened 2026-05-07
feat(ui): virtualize tree table

johallar:
Noticed slow downs when timelines get into the 50+ range, so virtualization is in order there's no reason to render timelines that are not visible.

* Adds virtualization + props to toggle/config virtualization to tree table component

https://github.com/user-attachments/assets/e0a3ddac-b243-49ea-8a2d-10ee017a8a2e


cmatzenbach (approved):
Looks good!

PR #149, opened 2026-05-07
fix(ui): ensure consistent timeline group highlighting

### LINKED ISSUE #132: Plan hovering not consistently working
Hovering to highlight the related timelines per physical plan is unconsistent and often times does not work. Also seems to rarely work when one of the plan's operator timeline trees are expanded.

https://github.com/user-attachments/assets/52400d4a-a182-415b-920e-4cd5995e4aee

cmatzenbach:
Previously, when hovering over a query plan to highlight a timeline group, the hovering itself was very inconsistent - it would randomly stop working. The issue was due to where the hover handlers were being attached; this creates a new prop in the base UI component and attaches the events to the actual row containers to ensure consistent highlighting. This also makes a small improvement, as previously `onMouseLeave` didn't have a `item.workerId` guard, so leaving a workerless plan could clobber a highlight set by another row.

Fixes: #132


johallar (approved):
💯

PR #151, opened 2026-05-07
refactor(ui): consolidate + move MAX_BINS const

### LINKED ISSUE #150: Reduce number of bins requested w/bulk caching strategy implemented
Image

FE can get more bins than exactly the # requested with tiling/caching scheme, think we should turn this down a little bit as we are approaching/exceeding the useful number to see on the screen probably

johallar:
We had this defined in a couple places 😓 this consolidates the const into one place and reduces the number of bins requested, which downsamples timelines with new caching changes: https://github.com/rapidsai/quent/pull/139

This starts looking more normal when zoomed in because of the timeline shape anyways, but when zoomed out it's very noticeable.

Before (400 bins requested)
Screenshot 2026-05-07 at 4 08 15 PM
Screenshot 2026-05-07 at 4 08 21 PM

After (200 bins requested)
Screenshot 2026-05-07 at 4 08 00 PM
Screenshot 2026-05-07 at 4 08 06 PM

Fixes #150 

Note: An alternative would be to not use the requested # bins directly as the tile size (assuming this is the case). If the target number of tiles per view is `n`, we could make the tile size `# bins / n`, which would give us approximately `# bins` when tiles are re-assembled


cmatzenbach (approved):
Looks much better, good to go

PR #159, opened 2026-05-11
feat: Add instrumentation API microbenchmark

### LINKED ISSUE #158: Instrumentation API microbenchmarks
For users instrumenting their code base, it would be useful to have a set of microbenchmarks for instrumentation API calls and exporters so they can get an idea of the instrumentation overhead.

johanpel:
Fixes #158

This adds microbenchmarks (plus profiling option) of the instrumentation API for Rust clients.

This has no payload other than timestamp acquisition, so it only measures the pure timestamp + send + export path.


dhruv9vats (approved):
Thanks @johanpel !

PR #160, opened 2026-05-11
feat: Add zero entropy event emitting executable for e2e / integration testing

### LINKED ISSUE #157: Zero entropy event emitting executable for e2e / integration testing
Something that would be very useful is a small binary emitting events similar to the simulator, but for which the entire information contents of the events is determined at compile-time, 0 entropy at run-time. This way we can build a pretty deep integration test all the way up to even visual parts of UI around it. If we make it generate data in some pattern that is easy to commit to our mind's memory as well, then it should also help when manually debugging stuff.

johanpel:
Closes #157

Adds an executable that emits events with zero entropy so they can be used in integration and e2e tests, as well as provide a simple enough event pattern that devs can memorize when manually debugging things across the stack.


johanpel:
@johallar @cmatzenbach one small TS change, but more importantly, please let me know if this seems more useful for debugging timelines than the more randomized simulator output.

johallar:
> @johallar @cmatzenbach one small TS change, but more importantly, please let me know if this seems more useful for debugging timelines than the more randomized simulator output.

TS change LGTM. We've been talking about e2e testing as well so definitely useful there. I wouldn't think we'll be making _too_ many assertions about the data itself, but this executable will be v helpful/necessary to create a small suite of data correctness or screenshot comparison smoke tests. 🙌

johallar (approved):

dhruv9vats (approved):

PR #162, opened 2026-05-12
feat(ui): deep linking

johallar:
mess right now, do not merge, may need to start over after the workspace refactor


johallar:
Too old, try again

PR #166, opened 2026-05-14
fix!: Remap paths through facade

dhruv9vats:
The public instrumentation crate API currently breaks with custom models because FSM bridge methods can pass cxx-generated `ffi::*` structs directly into model callbacks, or emit resource usage types like `quent_stdlib::memory::Memory` directly in generated Rust as direct source-crate paths bypass the public  instrumentation crate API. This does not work because then direct source-crate paths and actual model structs are treated as distinct Rust types.

This updates the generated Rust code so the model/resouce paths are resolved through the configured instrumentation crate facade instead of mixing facade paths with the originally defined crates.

This means that the instrumentation crate must now re-export the entities it import from other crates, so it acts as a single source of truth.


crates/codegen/Cargo.toml
```diff
@@ -16,4 +16,5 @@ syn = { version = "2", features = ["full", "parsing"] }
 quent-query-engine-model = { path = "../../domains/query_engine/model" }
 quent-readme-example = { path = "../../examples/readme" }
 quent-simulator-instrumentation = { path = "../../examples/simulator/instrumentation" }
+serde = { version = "1", features = ["derive"] }
```
johanpel:
```suggestion
serde = { workspace = true, features = ["derive"] }
```

johanpel (approved):
Thanks! One nit:

pixi.toml
```diff
@@ -6,6 +6,7 @@ requires-pixi = ">=0.68"
 
 [dependencies]
 rust = ">=1.93"
+rust-src = ">=1.93"
```
dhruv9vats, line 9: needed for static analyzers to use the the pixi env.

mbrobbel (approved):

PR #171, opened 2026-05-15
feat(ui): y axis redesign, compact timelines, 'relative' axis for resource timelines

### LINKED ISSUE #84: Timeline Y-Axis improvements
There's lots of padding to the left of the timelines RN to accommodate the y-axis labels, try moving onto the timelines themselves a la johan GTC changes. 

Also maybe do an iteration of rendering these outside of the echarts canvas for more control, since the number of ticks is a "suggestion" in echarts API and it can add ticks where we don't necessarily want them.

### LINKED ISSUE #156: Add a secondary time axis for the zoom window
Right now we have a global time axis across the entire query duration. Here the zoom window is also shown.

However, the data rendered on the timelines is not using that axis, so it is hard to "eyeball" durations without using the cursor. 

Example, can't easily eyeball the duration of the purple part here:

Image

It would be great to add a secondary time axis following the zoom window.

I can think of three flavors:
- One that has absolute timestamps.
- One that has relative timestamps to zoom window start, displayed separately.
- One that mixes the above

The advantage of the second one is that in large queries at zoom windows in high time ranges (e.g. 1337.42s), the tick labels may get harder to read/interpret quickly, plus they may get pretty big visually. When I use relative timestamps to the zoom window start, it would be easier to eyeball something like "oh this thing lasted from (window start) +3s to (window start) +7s = 4 seconds without using a cursor.

The third one is inspired by NSight Systems, see below. Note it doesn't have a global axis, which I think we should not drop.

Image

johallar:
## Refactors/Redesigns

* More compact timelines
* Refactor Y axis labels to remove negative space to the left
* Timeline controller axis labels don't overflow the page
* Column resize in real time, not debounced
* Remove the "Resource" and "Usage" headers, no real value there, merged these with the timeline controller "sub header" row and moved those up to be the headers.

## Feature
New feature in place of datazoom labels, show secondary axis above the resource timelines that shows "absolute" min/max times, and then relative intervals between eg. +20ms.
* Timeline ruler "relative" vs "absolute" time is controllable via prop, so can plug into a setting in the future

## Issues
Fixes #156 
Fixes #84 

## Screenshots
https://github.com/user-attachments/assets/5fd0b257-d540-4f55-bef6-d2eaf461251b


ui/packages/@quent/components/src/lib/timeline.utils.ts
```diff
@@ -678,3 +694,22 @@ export function collectVisibleEntries(
   }
   return result;
 }
+
+/** Max stacked value across non-dimmed, non-overlay bins within [zoomStartMs, zoomEndMs]. */
+export function computeVisibleMaxValue(
```
johallar, line 698: Just relocated

ui/packages/@quent/components/src/lib/useChartResize.ts
```diff
@@ -0,0 +1,52 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useCallback, useEffect, useRef } from 'react';
```
johallar, line 4: Isolated from useChartConnect

ui/packages/@quent/components/src/lib/useChartResize.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (dom && typeof ResizeObserver !== 'undefined') {
+      // Skip first fire to preserve entrance animations (mirrors useChartConnect).
+      let initial = true;
+      const observer = new ResizeObserver(() => {
+        if (initial) {
+          initial = false;
+          return;
+        }
+        try {
+          // `'auto'` forces zrender to re-measure; bare resize() reuses cached init pixels.
```
cmatzenbach: This AI comment doesn't seem that valuable, it's hard to read and seems to deal with the internal workings of `instance.resize` that aren't that relevant. I would remove.

ui/packages/@quent/components/src/resource-tree/ResourceGroupRow.tsx
```diff
@@ -36,32 +36,40 @@ export const ResourceGroupRow = ({
...
-          id={`${id}-fsm`}
-          label="FSM"
-          value={selectedFsmType ?? FSM_ALL}
-          options={fsmOptions}
-          onChange={(_, value) => onFsmChange(id, value === FSM_ALL ? null : value)}
-          className="mt-1"
-        />
+    <div className="pb-1">
+      <DataText className="text-xs font-bold leading-none">{group.instance_name}</DataText>
+      {hasMetadata && (
```
cmatzenbach, line 47: Good refactor, much clearer

ui/packages/@quent/components/src/timeline/Timeline.tsx
```diff
@@ -424,17 +419,34 @@ export function Timeline({
...
-      option={eChartOptions}
-      style={style}
-      onChartReady={handleChartReady}
-      notMerge={false}
-      lazyUpdate={false}
-      replaceMerge={['series']}
-    />
+    <div className="relative w-full h-full">
+      {maxValue != null && (
+        <span
```
cmatzenbach, line 424: Love this

ui/packages/@quent/components/src/timeline/TimelineController.tsx
```diff
@@ -323,10 +310,7 @@ export function TimelineController({
     onReady: onChartReady,
   });
 
-  // Restore the dataZoom slider position from the persisted atom whenever the
-  // zoom range changes or a (re)created chart instance becomes ready. The
-  // `readyTick` dependency ensures recreated instances inherit the saved zoom
-  // even when the atom value itself is unchanged across the remount.
+  // Restore the persisted zoom on range change or instance (re)creation.
```
cmatzenbach, line 313: ❤️

ui/packages/@quent/components/src/timeline/useVisibleMaxValue.ts
```diff
@@ -0,0 +1,21 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useMemo } from 'react';
+import { useZoomRange } from '@quent/hooks';
+import { computeVisibleMaxValue } from '../lib/timeline.utils';
+import type { TimelineSeries } from './types';
+
+/** Max stacked value across visible bins in the current zoom window. */
```
cmatzenbach: This is a super nit, but calling it a "stacked value" bothers me; I think it makes the comment more confusing to someone reading it by making it sound more complicated.

ui/packages/@quent/components/src/ui/tree-table.tsx
```diff
@@ -573,11 +573,13 @@ const VirtualizedTreeView = React.forwardRef<HTMLDivElement, TreeViewProps>(
       [normalizedData, effectiveExpandedIds]
     );
 
+    // Key measurements by item id so cached sizes survive expand/collapse reorderings.
```
cmatzenbach, line 576: 👍🏻

cmatzenbach (approved):
Code looks good, a few suggestions on some comment changes, everything else is good to go.

ui/packages/@quent/components/src/timeline/useVisibleMaxValue.ts
johallar: I think this is actually an important detail, it's not the max value of any state in the FSM, but the max combined value of all (stacked) states.

If you think "stacked" is not the right word, happy to swap it for whatever, but I think that's an important distinction for callers of this hook.

johallar: I'll just remove, the comment in the utility function i think does the job
`/** Max stacked value across non-dimmed, non-overlay bins within [zoomStartMs, zoomEndMs]. */`

PR #172, opened 2026-05-15
refactor(ui): stats panel re-design

### LINKED ISSUE #133: Redesign statistics panel to be collapsible
Currently, the statistics panel can be annoying when you want to click between operators in the tree without having it pop up. This will need some re-thinking/re-design, as making it collapsible might be awkward since it disappears when no operators are selected.  Perhaps make it permanent, with a "no operator selected" text when nothing is picked?

cmatzenbach:
Re-designs the stats panel so users can filter operators from the DAG without always having it take up real estate on the chart. Previously, whenever you selected an operator, the stats panel would always appear, and there was no way to close it.

Now, the stats panel has two states - a minimzed state and an expanded state. The panel always sits in the minimized state, but can be expanded by clicking a button. The button is only active when an operator is selected. When an operator is selected, the panel can be expanded to view operator details. The panel will stay expanded, allowing users to select other operators without having to re-open the panel. If a user clicks elsewhere on the DAG to de-select the current operator (or de-selects via clicking the same operator again), the stats panel will automatically collapse.

This is the second iteration of PR #168, as instead of having the panel sit on top of the DAG chart, it now sits in a panel below the DAG - when it is expanded, instead of overlaying, it pushes the DAG upwards to prevent it from blocking part of the chart. 

Also, previously selected operator statistics weren't being set in state by the gant chart, so added that so that the stats panel can be utilized when viewing operators selected on the gant chart.

Fixes: #133

### DIFF
--- a/ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
+++ b/ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
@@ -21,6 +21,7 @@ import {
   useSelectedNodeIds,
   useSetSelectedNodeIds,
   useSetSelectedOperatorLabel,
+  useSetSelectedNodeData,
   useSetSelectedPlanId,
   useNodeColoringValue,
   useNodeColorPalette,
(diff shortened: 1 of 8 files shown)


ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
```diff
@@ -296,16 +298,30 @@ export function OperatorGanttChart({
         if (selectedNodeIds.has(op.operatorId)) {
           setSelectedNodeIds(new Set());
           setSelectedOperatorLabel(null);
+          setSelectedNodeData(null);
         } else {
           setSelectedNodeIds(new Set([op.operatorId]));
           setSelectedOperatorLabel(op.label);
+          setSelectedNodeData({
```
johallar, line 305: 🙇

johallar (approved):
💯

PR #173, opened 2026-05-15
refactor(ui): cursor pointer where needed

johallar:
Sweep of the UI code, adds interactive cursor where needed so folks know they can click things 😅

### DIFF
--- a/ui/packages/@quent/components/src/dag/DAGControls.tsx
+++ b/ui/packages/@quent/components/src/dag/DAGControls.tsx
@@ -41,11 +41,11 @@ export const DAGControls = ({ operatorStatFields, portStatFields, isDark }: DAGC
   return (
     <Collapsible open={open} onOpenChange={setOpen} className="border-b bg-card">
       <div className="flex items-center justify-between px-4 py-2">
-        <CollapsibleTrigger className="flex items-center gap-2 group">
+        <CollapsibleTrigger className="flex items-center gap-2 group cursor-pointer">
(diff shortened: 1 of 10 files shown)


cmatzenbach:
One thing - did we end up making the top row collapsible or make it non-collapsible but removed `cursor: pointer`? Not sure if this isn't updated with latest `main` (or if that PR made it in), but the top row has `cursor: pointer` but isn't collapsible. Also, do we want to only have the pointer by the caret icons on the timeline or do we want to keep it over the whole area (the whole area is clickable, so maybe this will lead to discovery? if we want that)

https://github.com/user-attachments/assets/0aa51349-eda2-41d9-92f2-1ff625bac4c3

cmatzenbach (approved):
Approving, one small comment, make sure to address if necessary, but everything else looks good!

johallar:
> One thing - did we end up making the top row collapsible or make it non-collapsible but removed `cursor: pointer`? Not sure if this isn't updated with latest `main` (or if that PR made it in), but the top row has `cursor: pointer` but isn't collapsible. Also, do we want to only have the pointer by the caret icons on the timeline or do we want to keep it over the whole area (the whole area is clickable, so maybe this will lead to discovery? if we want that)
> 
>  cursor-pointer.mp4

Root item is collapsible, so this will fit together nicely. wrt the whole cell having `cursor: pointer`, I think this makes sense because the whole cell is a trigger to expand/collapse.

PR #181, opened 2026-05-18
refactor(ui): Use json-custom-numbers to parse big ints

johallar:
Sentinel value big int parser was a temporary fix to keep development moving, not optimal.

This uses a small fast secure dep instead to parse big ints from JSON which keeps it as a client only fix, as opposed to sending the big int as a string and parsing in the UI.


cmatzenbach (approved):
Looks good!

PR #183, opened 2026-05-19
experiment(ui): diff UI POC

johallar:
Building a mock API + UI to iterate on diff API

V0 API types:
https://github.com/johallar/quent/blob/8f986c3536068a47fde938fa841282edefdbc14f/ui/packages/%40quent/client/src/queryProfileDiffTypes.ts

Screenshot 2026-05-19 at 1 43 52 PM

Feel pretty good about the modularization so far, this uses the pivot table, radix, and timeline components to implement the diff view and the biggest component so far is the diff page's layout:

```
510  +510 -0  ui/src/pages/DiffSelectionPage.tsx
414  +414 -0  ui/src/components/query-diff/QueryDiffTimeline.tsx
227  +227 -0  ui/src/components/query-diff/QueryDiffTable.tsx
211  +211 -0  ui/packages/@quent/components/src/stat-card/StatisticCard.tsx
169  +169 -0  ui/src/components/query-diff/queryProfileDiffFromBundles.ts
152  +152 -0  ui/src/components/query-diff/QueryDiffStats.tsx
135  +135 -0  ui/src/components/query-diff/QueryDiffTimeline.utils.ts
 87   +87 -0  ui/src/components/query-diff/QueryDiffTable.utils.ts
 78   +78 -0  ui/src/components/query-diff/QueryDiffStats.utils.ts
 75   +75 -0  ui/packages/@quent/client/src/queryProfileDiffTypes.ts
 69   +69 -0  ui/packages/@quent/client/src/queryProfileDiff.ts
 45   +45 -0  ui/src/components/query-diff/QueryDiffColors.ts
 40   +33 -7  ui/packages/@quent/components/src/pivot-table/PivotedStatTable.tsx
 30   +30 -0  ui/packages/@quent/client/src/api.ts
 28   +28 -0  ui/src/index.css
 21   +21 -0  ui/packages/@quent/client/src/index.ts
 20   +20 -0  ui/src/routes/diff.engine.$engineId.query.$queryAId.compare.$queryBId.tsx
 19   +19 -0  ui/packages/@quent/components/src/pivot-table/types.ts
 14   +14 -0  ui/src/routes/__root.tsx
 13   +13 -0  ui/src/routes/diff.index.tsx
 12   +12 -0  ui/src/routes/diff.tsx
 11   +11 -0  ui/packages/@quent/components/src/index.ts
 ```

### CODE REVIEW

File: ui/packages/@quent/client/src/queryProfileDiffTypes.ts [hunk 1]
```diff
@@ -0,0 +1,76 @@
...
+  TaskFilter,
+  TimelineConfig,
+} from '@quent/utils';
+
+export interface QueryProfileDiffRequest {
+  query_a_id: string;
+  query_b_id: string;
+}
+
+export type QueryProfileDiffScenario = 'plans_equal' | 'plans_different' | 'plans_incomparable';
```
cmatzenbach: We'll need to define what makes two plans "incomparable" vs "different" (e.g. different SQL text? different engines?)

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface QueryProfileDiffQuerySummary {
```
cmatzenbach: add:
```
  duration_s: number; // rather than deriving it from the query bundle
```

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export interface QueryProfileDiffOperatorRef {
+  id: string;
+  label: string;
+  operator_type_name: string | null;
+  plan_id: string | null;
+}
+
+export interface QueryProfileDiffStatDelta {
```
cmatzenbach: We could consider adding a FE-only derived type that wraps this with a `direction` field ('better' | 'worse' | 'neutral'). The direction is already being re-derived from the delta in multiple places (`QueryDiffTable.utils`, `QueryDiffStats.utils`)

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface QueryProfileDiffOperatorDelta {
+  operator_a: QueryProfileDiffOperatorRef | null;
+  operator_b: QueryProfileDiffOperatorRef | null;
+  stats: Record<string, QueryProfileDiffStatDelta>;
+}
+
+export interface QueryProfileDiffPlanComparison {
+  /* Big question here, how do we represent query plan graph diffs */
+  match_kind: 'structural' | 'different' | 'incomparable';
```
cmatzenbach: `match_kind` duplicates information already in `QueryProfileDiffScenario` - if scenario is 'plans_equal', match_kind is always 'structural', etc

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+  percent_delta: number | null;
+}
+
+export interface QueryProfileDiffOperatorDelta {
+  operator_a: QueryProfileDiffOperatorRef | null;
+  operator_b: QueryProfileDiffOperatorRef | null;
+  stats: Record<string, QueryProfileDiffStatDelta>;
+}
+
+export interface QueryProfileDiffPlanComparison {
```
cmatzenbach: The backend could do graph-based matching instead - actually traversing the plan graph and pairing operators by their role and connections, not just their order. This would solve plans not being structurally identical. Then we could add match_strategy: 'positional' | 'graph' to record how operators were paired. As we will have the backend doing matching, the UI needs to know which was used

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+  /* Big question here, how do we represent query plan graph diffs */
+  match_kind: 'structural' | 'different' | 'incomparable';
+  matched_operator_count: number;
+  unmatched_operator_a_count: number;
+  unmatched_operator_b_count: number;
+}
+
+export interface QueryProfileDiffResponse {
```
cmatzenbach: Add:
```
shared_resource_types: string[]; // covers the timeline resource type dropdown, server already knows which resource types are shared
```

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+  query_a: QueryProfileDiffQuerySummary;
+  query_b: QueryProfileDiffQuerySummary;
+  plan_comparison: QueryProfileDiffPlanComparison;
+  operator_diffs: QueryProfileDiffOperatorDelta[];
+  warnings?: string[];
+}
+
+export type QueryProfileDiffTimelineEntries<T> = [T, T, ...T[]];
+
+export interface QueryProfileDiffTimelineRequest {
```
cmatzenbach: Right now we are having the FE build two `SingleTImelineRequest` objects and pass them up; however, the server already has all this info. Instead we should treat this like we do `QueryProfileDiffRequest`, here's my suggestion for the revised type once we have the API compute everything:

```
export interface QueryProfileDiffTimelineRequest {
  query_a_id: string;
  query_b_id: string;
  resource_type: string;
  config: TimelineConfig;
}
```

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+  timelines: QueryProfileDiffTimelineEntries<SingleTimelineRequest<QueryFilter, TaskFilter>>;
+  delta_config: TimelineConfig;
+}
+
+export interface QueryProfileDiffTimelineResponse {
```
cmatzenbach: Here would be the response proposed for the request above:
```
export interface QueryProfileDiffTimelineResponse {
  timeline_a: SingleTimelineResponse;
  timeline_b: SingleTimelineResponse;
  delta: SingleTimelineResponse;
  warnings?: string[];
}
```

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
johallar: Yeah this will definitely need some figuring, It's also likely we'll need this flag on various things (eg. query plan, resource tree), so should probably make this a more generic "can these things be compared" flag or something

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
johallar: Agree, plan to remove this for now

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,76 @@
...
+  QueryFilter,
+  SingleTimelineRequest,
+  SingleTimelineResponse,
+  StatValue,
+  TaskFilter,
+  TimelineConfig,
+} from '@quent/utils';
+
+export interface QueryProfileDiffRequest {
+  query_a_id: string;
```
johallar: This needs to be either an array (with the assumption that the first item is the "baseline"), or have explicit `baseline` and `competitors` (or similarly named) properties

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
johallar: We'll need to handle `n` queries in the comparison requests, so should use the same strategy (whatever is landed on) as `QueryProfileDiffTimelineRequest`, otherwise agree

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
johallar: Definitely, this type will be the result of the graph isomorphism problem. The whole type will change once we have more of that strategy figured out I assume.

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
johallar: This could add some semantic value... i'm thinking if the analyzer can infer something that the FE can't (maybe a higher value is better for specific stats?)  then this could be very useful.

ui/packages/@quent/client/src/queryProfileDiffTypes.ts
```diff
@@ -0,0 +1,106 @@
...
+} from '@quent/utils';
+
+export interface DiffQueryRef {
+  // @chris tell me if engine is needed, is group needed?
+  engine_id: string;
+  query_id: string;
+}
+
+export interface DiffRequest {
+  // Alternatively make this one array, assume first item is baseline
```
johallar, line 20: One thought: We talked about comparing query groups (eg. compare tpch runs from 2 commits), we should think that through and update this to accommodate. Not sure about comparing engines, but probably should handle that as well.

johallar:
POC did its job, replaced by `johallar:workload-diff-ui-stage-1` and https://github.com/rapidsai/quent/pull/193

PR #185, opened 2026-05-19
feat(ui): add tabs to plan viewer and clean up plan UI

cmatzenbach:
- Move settings to a tab so user can switch between the query plan selection and the DAG settings
- Reduce height of container holding the query plan selection/settings to fit the settings panel
- Small UI cleanup items


ui/src/components/QueryPlan.tsx
```diff
@@ -25,13 +24,24 @@ import {
...
 import { DataText } from '@quent/components';
 import { useTheme, THEME_DARK } from '@/contexts/ThemeContext';
+import { cn } from '@quent/utils';
 
 // Lazy load DAGChart to split elkjs (~1.6MB) into a separate chunk
 const DAGChart = lazy(() => import('@quent/components').then(mod => ({ default: mod.DAGChart })));
 
+type PlanTab = 'plan' | 'controls';
+
+const tabClass = cn(
```
johallar: Would prefer https://ui.shadcn.com/docs/components/radix/tabs here, so primitives pull from theming/etc

ui/packages/@quent/components/src/ui/tree-view.tsx
```diff
@@ -420,7 +420,7 @@ function TreeLeaf<T extends TreeDataItem = TreeDataItem>({
   return (
     <div
       className={cn(
-        'ml-5 flex text-left items-center py-2 cursor-pointer before:right-1',
+        'ml-5 flex text-left items-center py-0.5 cursor-pointer before:right-1',
```
johallar: Yeah maybe `py-1`, it is a tad squished

johallar (changes requested):

ui/src/components/QueryPlan.tsx
cmatzenbach: Agreed - good idea. Replaced

ui/package.json
```diff
@@ -44,12 +44,20 @@
     "@radix-ui/react-scroll-area": "^1.2.10",
     "@radix-ui/react-select": "^2.2.6",
     "@radix-ui/react-slot": "^1.2.4",
+    "@radix-ui/react-tabs": "^1.1.13",
     "@tanstack/react-query": "^5.90.21",
     "@tanstack/react-query-devtools": "^5.91.3",
+<<<<<<< matzy/query-plan-design-cleanup
```
johallar: Remove merge conflict tags, little curious how this is not failing CI, are these real?

ui/src/components/QueryPlan.tsx
```diff
@@ -129,47 +134,53 @@ export function QueryPlan({ queryId, engineId }: { queryId: string; engineId: st
...
-            onSelectChange={handlePlanSelect}
-            onItemHover={item => setHoveredWorkerId(item?.workerId ?? null)}
-            renderItem={renderItem}
-          />
+        <ResizablePanel defaultSize="15%" className="flex flex-col">
+          <Tabs defaultValue={TABS.PLAN} className="flex flex-col flex-1 overflow-hidden">
+            <TabsList className="h-9 w-full shrink-0 justify-start rounded-none border-b bg-transparent p-1">
+              <TabsTrigger
+                value={TABS.PLAN}
+                className="rounded-md px-3 py-1 text-sm font-normal data-[state=active]:bg-muted data-[state=active]:font-semibold data-[state=active]:shadow"
```
johallar: Most of these classes should be baked into the `TabsTrigger` yeah? Hoping we can have little to no classname customization in this component, same for `TabsList`

johallar: Maybe same for `Tabs` as well? Not exactly sure what all can be consolidated to the shadcn component.

ui/package.json
cmatzenbach: Not sure what happened there...fixed

ui/src/components/QueryPlan.tsx
cmatzenbach: Good catch, I always forget to add those on the original components. Moved all tailwind classes to the shadcn root component.

johallar (approved):

Issue #191, opened 2026-05-29
Schema-based instrumentation architecture


johanpel:
# Introduction

[#112](https://github.com/rapidsai/quent/pull/112) has marked the end of the PoC phase of Quent by demonstrating we can define composable and reusable application event models from single & relatively straightforward definitions from which an enormous amount of boilerplatey code can be generated to tie everything from instrumentation library all the way to analysis and UI together.

The current PoC-grade implementation has quite a lot of tech debt that needs to be erased in order to stabilize the project and provide a solid foundation for both users and a next wave of features. To this end, https://github.com/rapidsai/quent/pull/128 has dissected the entire project into somewhat atomic parts, and provides a proposal for a more solid foundation.

This is a tracking issue for migrating the project to that architecture. The end result will be a first beta release (v0.1, this may require a few follow-ups after this issue is completed, so consider this an umbrella issue).

# Schema-based instrumentation architecture

## Schema

The new architecture centers itself around a "schema" for event data, which at the core only describes those things required to emit, transport, and store event data. This only includes:
- Entity types, including their events and event attributes
- Record types (compile-time defined attributes)

Each element in the schema can carry metadata for arbitrary purposes, as well as "constraints"

## Constraints

The idea is to map all other currently existing concepts (tree-forming references, FSMs, resources, query engine stuff) onto this schema as opaque "constraints". Constraints will be a special type of metadata that impose certain restrictions or rules on the entities, events, or data. These restrictions/rules are (should be) validated in any layer of Quent that does something with these constraints. 

Whether some layer of Quent can/should do something with these constraints depends on the constraint. Therefore, the benefit of introducing these constraints as explicitly separated from the core concepts, is that certain layers can be made oblivious to these constraints such that they can continue to work even if these constrains change. In the long term, the idea is that third parties can also author domain- or application-specific constraints to impose certain semantics on the events.

### Example

Quent's stack consists of several layers, roughly speaking:
1. Model definition: e.g. a YAML-based DSL to capture a schema
2. Instrumentation: code generation step to produce a strongly typed library + potential bindings for C++, Python, ...
3. Application: <application developer instruments code>
4. Capture & store: event export & storage
5. Analysis: semantic analysis of events & serving a front-end
6. Presentation: e.g. a UI, CLI tool, MCP endpoint, etc.

[...]

PR #192, opened 2026-05-29
feat: add schema and constraint crates

johanpel:
Adds a schema for capturing an application event model and a canonical path for validating conventions against the schema.

Based off proposal explored in #128. Part of #191.


crates/convention/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+
+/// A trait for types that implement a "convention".
+///
+/// A convention is a set of properties that can be added to an application
+/// event model. It is conveyed through opaque data attached to the constituents
+/// of a [`Schema`].
+///
+/// By applying the convention to a model, it may obtain properties that need to
+/// be validated against the entire schema, which is the main purpose of this
+/// trait.
```
mbrobbel: Maybe constraint?

crates/convention/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+pub trait Convention {
+    /// A unique name for this convention.
+    ///
+    /// While no restrictions are imposed on convention names (other than that
+    /// they are valid UTF-8 strings) it is recommended to follow the
+    /// human-readable dot-separated pattern `project.convention.version`. For
+    /// example: `quent.fsm.v1`. This reduces the probability of name clashes
+    /// between dependencies, and provides a means of easily detecting breaking
+    /// changes to the convention's own schema.
+    fn name(&self) -> &'static str;
```
mbrobbel: Why not `const NAME: &'static str`?

crates/convention/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    /// Validate this convention against `schema`.
+    fn validate(&self, schema: &Schema) -> Result<(), Vec<Error>>;
```
mbrobbel: Maybe multiple errors should be handled inside `Error` (instead of returning `Vec<Error>`?)

crates/convention/src/lib.rs
```diff
@@ -0,0 +1,174 @@
...
+            Entry::Occupied(_) => Err(Error::DuplicateConvention(convention.name())),
+            Entry::Vacant(entry) => {
+                entry.insert(Box::new(convention));
+                Ok(self)
+            }
+        }
+    }
+
+    /// Run validation of all registered conventions against `schema`.
+    pub fn run(&self, schema: &Schema) -> Result<(), Vec<Error>> {
```
mbrobbel: 🏃 ? `validate`?

crates/schema/src/metadata.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Convention {
+    /// The name of the convention.
+    pub name: String,
+    /// Whether the convention is one that requires validation.
+    pub validated: bool,
+    /// Convention-specific metadata.
+    pub data: Option<String>,
```
mbrobbel, line 17: Why use `String` here?

crates/schema/src/data_type.rs
```diff
@@ -0,0 +1,38 @@
...
+    /// The top-level [`crate::Schema::records`] field must contain a
+    /// [`crate::record::Record`] with this name, otherwise the
+    /// [`crate::Schema`] is ill-formed.
+    Record(Identifier),
+    /// A record whose fields are determined by the instrumentation client at
+    /// run-time.
+    DynamicRecord,
+    /// A reference to an entity optionally carrying data.
+    EntityRef {
+        data: Option<Box<DataType>>,
```
mbrobbel: Just curious; why a named `data` field vs tuple field (like the other variants)?

crates/schema/src/identifier.rs
```diff
@@ -0,0 +1,178 @@
...
+pub enum IdentifierError {
+    /// The string is empty.
+    Empty,
+    /// The first character is an ASCII letter `[A-Za-z]`.
+    InvalidStart(char),
+    /// A character `ch` at byte offset `index` is not in `[A-Za-z0-9_]`.
+    InvalidChar { ch: char, index: usize },
+}
+
+impl std::fmt::Display for IdentifierError {
```
mbrobbel, line 27: Just curious; why not also use `thiserror` here?

crates/schema/src/identifier.rs
```diff
@@ -0,0 +1,178 @@
...
+                return Err(IdentifierError::InvalidChar { ch, index });
+            }
+        }
+        Ok(Self(s))
+    }
+
+    /// Borrows the identifier as a string slice.
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
```
mbrobbel: This overlaps with `impl AsRef<str>`.
```suggestion
```

crates/schema/src/identifier.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Consumes the identifier, returning the inner [`String`].
+    pub fn into_string(self) -> String {
+        self.0
+    }
```
mbrobbel: Overlaps with `impl From<Identifier> for String`.
```suggestion
```

crates/schema/src/identifier.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl std::borrow::Borrow<str> for Identifier {
+    fn borrow(&self) -> &str {
+        &self.0
+    }
+}
```
mbrobbel: `AsRef<str>`
```suggestion
```

crates/schema/src/identifier.rs
```diff
@@ -0,0 +1,178 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&self.0)
+    }
+}
+
+impl AsRef<str> for Identifier {
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
```
mbrobbel: fyi https://doc.rust-lang.org/stable/std/convert/trait.AsRef.html#generic-implementations

mbrobbel (commented):
Some questions and suggestions. This looks a lot like Arrow schema's and their metadata (+ extension types).

crates/schema/src/metadata.rs
johanpel: On the one hand I felt like encouraging some human-readable format for this metadata because I know humans will be looking at these schemas for various reasons. On the other hand, I don't want to restrict it too much. What do you think? We can either go bytes or constrain it more.

crates/schema/src/identifier.rs
johanpel: I may have overoptimized on making it depend on _nothing_ else in order to make it as stable as possible. Of course te serde exception exists in case folks want to serialize it before we come up with some stable serialized format ourselves.

crates/schema/src/identifier.rs
johanpel: https://github.com/rapidsai/quent/pull/192/commits/0e133bc929928691b9c4ca7aa2df9b405c36d423

crates/schema/src/data_type.rs
johanpel: No reason anymore, https://github.com/rapidsai/quent/pull/192/commits/59e04954d930e525dabb8626b55e0dfe8c6afb0d

crates/convention/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/192/commits/68c5d3fef70913d38f68e328097910982add91d0

crates/convention/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/192/commits/68c5d3fef70913d38f68e328097910982add91d0

crates/convention/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/192/commits/124ca7b93a16bcec2fa01b8ee709d8d9369ad905

crates/convention/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/192/commits/1dfa802aa74236ccc1dcad2e8f19c49ec3070ff9

mbrobbel (approved):

dhruv9vats (approved):

9prady9 (commented):
audit (for KT)

PR #194, opened 2026-06-01
feat: add FSM constraint crate

johanpel:
This PR adds the `quent-fsm` crate. This is a built-in constraint (also see #192) ("quent.fsm.v1") that restricts the order of an entity's events according to some FSM topology that an entity can be annotated with in a schema. 

This can in turn be leveraged by code generation steps to produce a more specific instrumentation API for this entity. For example e.g. in Rust following the typestate pattern, the Rust compiler will prevent you from emitting events that make an illegal transition.

This is widely applicable wherever clients want to trace any type of lifecycle of entities in their code. 
It can even be used to form more traditional traces, e.g. of trees of "spans" (a span could have one "started/open" state like in OTel tracing, or go through cycles of "enter, exit, ..." , like in the case of the Rust `tracing` crate where this helps trace the execution of async tasks.

The reference instrumentation library generator and modeling APIs will support this constraint.

Integration with codegen was explored in #128. This is part of the larger effort described in #191.

### DIFF
--- a/crates/fsm/src/lib.rs
+++ b/crates/fsm/src/lib.rs
@@ -40,6 +40,14 @@ impl Transition {
     }
 }
 
+/// A non-empty set of states from which an `Fsm` can transition out of
+/// existence.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+struct ExitStates {
+    state: Identifier,
+    others: Vec<Identifier>,
+}
+
 /// The state-transition topology of a finite state machine.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Fsm {
@@ -48,15 +56,21 @@ pub struct Fsm {
     initial_state: Identifier,
     /// The possible transitions of this FSM.
     transitions: Vec<Transition>,
-    /// The names of states from which this FSM can exit to go out of existence.
-    exit_from_states: Vec<Identifier>,
+    /// The states (at least one) from which this FSM can exit to go out of
+    /// existence.
+    exit_from_states: ExitStates,
 }
 
 impl Fsm {
     /// Return a builder for an [`Fsm`] constraint over `entity`, beginning in
-    /// `initial_state`.
-    pub fn builder(entity: &Entity, initial_state: Identifier) -> FsmBuilder<'_> {
-        FsmBuilder::new(entity, initial_state)
+    /// `initial_state` and able to exit from `exit_state`. Further exit states
+    /// can be added on the builder.
+    pub fn builder(
+        entity: &Entity,
+        initial_state: Identifier,
+        exit_state: Identifier,
+    ) -> FsmBuilder<'_> {
+        FsmBuilder::new(entity, initial_state, exit_state)
     }
 
     /// The name of the initial state this FSM transitions into when it comes
@@ -65,14 +79,17 @@ impl Fsm {
         &self.initial_state
     }
 
-    /// The transitions of this FSM.
+    /// The transitions of this FSM between states.
+    ///
+    /// This excludes exit transitions. Obtain these through
+    /// [`Self::exit_from_states`].
     pub fn transitions(&self) -> &[Transition] {
         &self.transitions
     }
 
-    /// The names of states from which this FSM can exit to go out of existence.
-    pub fn exit_from_states(&self) -> &[Identifier] {
-        &self.exit_from_states
+    /// The states from which this FSM can exit to go out of existence.
+    pub fn exit_from_states(&self) -> impl Iterator<Item = &Identifier> {
+        std::iter::once(&self.exit_from_states.state).chain(self.exit_from_states.others.iter())
     }
 }
 
@@ -106,13 +123,13 @@ impl Fsm {
 /// For every entity carrying this constraint:
 ///
 /// 1. No event in the entity may be named `exit` (case-insensitively).
-/// 2. There is at least one exit transition.
-/// 3. Every state named by the FSM corresponds to an event name in the entity.
-/// 4. Every event in the entity appears as a state in the FSM.
-/// 5. Every state is reachable from the initial state.
-/// 6. An exit transition is reachable from every state.
-/// 7. A state on a cycle has [`Cardinality::Multi`], otherwise
+/// 2. Every state named by the FSM corresponds to an event name in the entity.
+/// 3. Every event in the entity appears as a state in the FSM.
+/// 4. Every state is reachable from the initial state.
+/// 5. An exit transition is reachable from every state.
+/// 6. A state on a cycle has [`Cardinality::Multi`], otherwise
 ///    [`Cardinality::Once`].
+/// 7. At least one exit transition exists (enforced by [`ExitStates`]).
 pub struct FsmConstraint;
 
 impl Constraint for FsmConstraint {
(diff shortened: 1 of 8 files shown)


crates/fsm/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+    Exit,
+    /// A regular state, named after the event that transitions into it.
+    Named(Identifier),
+}
+
+/// A directed transition between two [`State`]s in an [`Fsm`].
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct Transition {
+    pub source: State,
+    pub target: State,
```
mbrobbel: `pub` intended?

crates/fsm/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+}
+
+/// The state-transition topology of a finite state machine.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Fsm {
+    pub transitions: Vec<Transition>,
```
mbrobbel: Same question

crates/fsm/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+///    transition has [`State::Entry`] as its target.
+/// 3. At least one transition has [`State::Exit`] as its target, and no
+///    transition has [`State::Exit`] as its source.
+/// 4. There is at least one [`State::Named`] state.
+/// 5. Every [`State::Named`] state is reachable from [`State::Entry`].
+/// 6. Every [`State::Named`] state can reach [`State::Exit`].
+/// 7. Every [`State::Named`] state corresponds to an event in the entity.
+/// 8. Every event in the entity appears as some [`State::Named`] state.
+/// 9. A state on a cycle has [`Cardinality::Multi`], otherwise
+///    [`Cardinality::Once`].
```
9prady9, line 131: 👍🏾

crates/fsm/src/lib.rs
mbrobbel: Instead of `Vec<Transition>` and `State` with `Entry` and `Exit` variants, maybe we can do `Fsm { entry: State, transitions: Vec<Transition>, exit: State }` ,then `State` is just `Named(Identifier`)?

crates/fsm/src/lib.rs
```diff
@@ -0,0 +1,389 @@
...
+            for node in scc {
+                if let GraphNode::Named(name) = node {
+                    on_cycle.insert(name);
+                }
+            }
+        }
+    }
+    // It also sits on a cycle if it has a self-loop, but `tarjan_scc` does not
+    // report those as cyclic, so do them separately:
+    for t in &fsm.transitions {
```
9prady9, line 297: perhaps a test for this to ensure it (this function) captures cycles as expected to avoid introducing bugs later.

crates/fsm/src/lib.rs
9prady9: I think existing choice is more apt, but neither is wrong in any way. Just a choice I suppose - what helps writing the downstream code more efficiently/easy-to-read.

johanpel: Transitions from multiple states to exit are possible.
This could work for entry as that is currently restricted to go into one single named state.

crates/fsm/src/lib.rs
johanpel: Yes, downstream/upstream tools need to be able to read/modify these fields

crates/fsm/src/lib.rs
johanpel: Same as above

crates/fsm/src/lib.rs
johanpel: This is tested here: https://github.com/rapidsai/quent/pull/194/changes/BASE..9ef277e95c23bb66645959b41eec95bef010dbaa#diff-1d57f1871a4754db14869e8e550135e071ba047ec8fbdad3704dd9d8dc781b1dR104-R124

9prady9 (approved):
lgtm

crates/fsm/src/lib.rs
mbrobbel: Right, or `exit: Vec<State>`?

mbrobbel: I'm suggesting this to avoid data for something that can be captured at the type level.

crates/fsm/src/lib.rs
mbrobbel: Maybe add read/modify methods to avoid: `source: State::Entry, target: State::Entry`? Or remove those variants as suggested below.

crates/fsm/src/lib.rs
9prady9: what does that enable us to do ?

johanpel: I'm processing this right now. This is a good idea as it prevents us from forming some nonsensical structures

crates/fsm/src/lib.rs
johanpel: Agreed to make this as hard to misuse as possible.

We could have a builder for `Fsm` struct that takes a reference to an existing `quent_schema::entity` and verifies we're not building anything nonsensical (e.g. refer to events the entity's schema part doesn't define).

johanpel: https://github.com/rapidsai/quent/pull/194/commits/b7226798f4ebc04d7353deae85d669d39b7619f6

and

https://github.com/rapidsai/quent/pull/194/commits/7d8330611487210e64982f071f93842c38499c2e

crates/fsm/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/194/commits/b7226798f4ebc04d7353deae85d669d39b7619f6
and
https://github.com/rapidsai/quent/pull/194/commits/7d8330611487210e64982f071f93842c38499c2e

9prady9:
Updates make sense, nice!

crates/fsm/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+/// The state-transition topology of a finite state machine.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Fsm {
+    /// The name of the initial state this FSM transitions into when it comes
+    /// into existence.
+    initial_state: Identifier,
+    /// The possible transitions of this FSM.
+    transitions: Vec<Transition>,
+    /// The names of states from which this FSM can exit to go out of existence.
+    exit_from_states: Vec<Identifier>,
```
mbrobbel: > 2. There is at least one exit transition.

We could skip this check by encoding "at least one" (non-empty vec) as `ExitState { exit: Identifier, other: Vec<Identifier> }`, (and then provide `IntoIterator<Item = Identifier>`).

crates/fsm/src/lib.rs
```diff
@@ -0,0 +1,362 @@
...
+
+    // Requirement 4: every entity event appears as a state.
+    for event in &entity.events {
+        if !states.contains(&event.name) {
+            errors.push(FsmError::UncoveredEvent {
+                entity: entity.name.clone(),
+                event: event.name.clone(),
+            });
+        }
+    }
```
mbrobbel: I guess error messages are better like this, but we could also just check `event_names == states`?

If you want to keep the error message details you could use https://doc.rust-lang.org/stable/std/collections/struct.HashSet.html#method.difference

crates/fsm/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/194/commits/a32c36ea6c18f10c0625e6d69ecd77a9662c2d9c

crates/fsm/src/lib.rs
johanpel: Prefer to keep the details as these two situations are not exactly the same: https://github.com/rapidsai/quent/pull/194/commits/a32c36ea6c18f10c0625e6d69ecd77a9662c2d9c

mbrobbel (approved):

Issue #195, opened 2026-06-02
Introduce reference role constraints


johanpel:
With the introduction of `quent-schema` in #192 as a part of #191, entity reference can be imbued with data determined at run time. The schema can also carry annotations (constraints, metadata) in fields that use the entity reference type.

The combination of data and field annotations can be leveraged to e.g.:
1. constraining the reference to target a specific type of entity
2. carrying usage information when using the framework's resource-oriented concepts
3. within the full entity graph, express specific paths, e.g. to provide a canonical path to "drill down" in a trace

All of the above exist in the current PoC state already, but it is implemented in a very tightly integrated fashion that is hard to extend.

PR #199, opened 2026-06-03
feat: add ref-target constraint crate

johanpel:
Adds `quent-ref-target` crate with the `quent.ref-target.v1` constraint, which restricts an `EntityRef` to a named target entity and validates that the target is a declared entity in the schema.

One part of https://github.com/rapidsai/quent/issues/195

### DIFF
--- a/crates/ref-target/src/lib.rs
+++ b/crates/ref-target/src/lib.rs
@@ -1,0 +1,132 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Constraint for entity reference restricting the type of entity it can point to.
+
+use quent_constraints::Constraint;
+use quent_schema::{Schema, data_type::DataType, entity::Entity, identifier::Identifier};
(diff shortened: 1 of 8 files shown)


crates/ref-target/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+    Multiple(Vec<RefTargetError>),
+}
+
+fn join_errors(errors: &[RefTargetError]) -> String {
+    errors
+        .iter()
+        .map(|e| format!("  - {e}"))
+        .collect::<Vec<_>>()
+        .join("\n")
+}
```
mbrobbel, line 132: Maybe create a helper to avoid duplication?
https://github.com/rapidsai/quent/blob/60a43376f0fda4cd6745ac4e1bec7b222f10a7cb/crates/fsm/src/lib.rs#L354-L360)

mbrobbel (approved):

crates/ref-target/src/lib.rs
johanpel: Will address this separately, this is also duplicated in #200 and the same goes for building invalid test data

johanpel: https://github.com/rapidsai/quent/issues/202

PR #200, opened 2026-06-03
feat: add ref-tree constraint crate

### LINKED ISSUE #195: Introduce reference role constraints

johanpel:
Adds the `quent-ref-tree` crate which implements an opt-in schema constraint that requires the definition of a tree of entities in a graph of entities referring to each other.

Part of #191 and closes #195 to feature parity with pre-schema-based arch

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2206,6 +2206,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "quent-ref-tree"
+version = "0.1.0"
+dependencies = [
+ "quent-constraints",
+ "quent-ref-target",
+ "quent-schema",
+ "serde_json",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "quent-schema"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2209,9 +2209,11 @@ dependencies = [
 name = "quent-ref-tree"
 version = "0.1.0"
 dependencies = [
+ "petgraph",
  "quent-constraints",
  "quent-ref-target",
  "quent-schema",
+ "rustc-hash",
  "serde_json",
  "thiserror 2.0.18",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,6 +52,7 @@ members = [
     "crates/constraints",
     "crates/schema",
     "crates/ref-target",
+    "crates/ref-tree",
     "crates/fsm",
 ]
 
--- a/crates/ref-target/src/lib.rs
+++ b/crates/ref-target/src/lib.rs
@@ -5,7 +5,7 @@
 
 use quent_constraints::{Constraint, utils::bullet_list};
 use quent_schema::{
-    DataType, Identifier,
+    Annotations, DataType, Identifier,
     visitor::{Cursor, Element, Visitor},
 };
 use serde::{Deserialize, Serialize};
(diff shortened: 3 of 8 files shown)


9prady9 (commented):
nice work, I am a bit confused though on whether the following should pass or fail. 

Based on my current understanding the following should fail ?  - maybe I am understanding that incorrectly.

```rust
#[test]
fn direct_and_record_borne_refs_in_one_event_is_rejected() {
    // Req 2: an event carries AT MOST ONE tree-forming reference, counting the
    // ones reached through record-typed fields. Here `Worker.created` carries:
    //   - a direct parent ref      (field "direct" -> Cluster)
    //   - a record-borne parent ref (field "meta" -> Meta.owner -> Cluster)
    // That is two tree-forming references in one event, so it must be rejected.
    let meta = record("Meta", vec![field("owner", tree_ref_to("Cluster"))]);
    let worker = entity(
        "Worker",
        vec![event(
            "created",
            vec![
                field("direct", tree_ref_to("Cluster")),
                field("meta", DataType::Record(ident("Meta"))),
            ],
        )],
    );
    let schema = schema_with_records(vec![root("Cluster"), worker], vec![meta]);
    // EXPECTED: exactly one MultiplePerEvent violation.
    // ACTUAL (current crate): validate() returns NO errors, so `single_error`
    // panics on `assert_eq!(errors.len(), 1)` -> test fails, exposing the bug.
    assert!(matches!(
        single_error(&schema),
        RefTreeError::MultiplePerEvent { .. }
    ));
}
```

johanpel:
> nice work, I am a bit confused though on whether the following should pass or fail.
> 
> Based on my current understanding the following should fail ? - maybe I am understanding that incorrectly.
> 
> ```rust
> #[test]
> fn direct_and_record_borne_refs_in_one_event_is_rejected() {
>     // Req 2: an event carries AT MOST ONE tree-forming reference, counting the
>     // ones reached through record-typed fields. Here `Worker.created` carries:
>     //   - a direct parent ref      (field "direct" -> Cluster)
>     //   - a record-borne parent ref (field "meta" -> Meta.owner -> Cluster)
>     // That is two tree-forming references in one event, so it must be rejected.
>     let meta = record("Meta", vec![field("owner", tree_ref_to("Cluster"))]);
>     let worker = entity(
>         "Worker",
>         vec![event(
>             "created",
>             vec![
>                 field("direct", tree_ref_to("Cluster")),
>                 field("meta", DataType::Record(ident("Meta"))),
>             ],
>         )],
>     );
>     let schema = schema_with_records(vec![root("Cluster"), worker], vec![meta]);
>     // EXPECTED: exactly one MultiplePerEvent violation.
>     // ACTUAL (current crate): validate() returns NO errors, so `single_error`
>     // panics on `assert_eq!(errors.len(), 1)` -> test fails, exposing the bug.
>     assert!(matches!(
>         single_error(&schema),
>         RefTreeError::MultiplePerEvent { .. }
>     ));
> }
> ```

Nice catch, that was indeed a missing test case, fixed here: https://github.com/rapidsai/quent/pull/200/commits/7ce70ebb3fbe054029afebad7ec91045e051bbef

9prady9 (approved):

Issue #202, opened 2026-06-04
Schema and constraint utils


johanpel:
There are some repeated patterns in #194, #199, and #200 that we can DRY up

- Add schema builder
- Add invalid schema builder for tests
- Add some other utils e.g. to join multiple errors into a single error

Issue #197, opened 2026-06-02
Base schema validator


johanpel:
A `quent_schema::Schema` introduced in #192 can be malformed in some ways (besides through invalid constraint annotations):

- Fields that reference other things by their `Identifier` could hold an `Identifier` that isn't present in the model.
- A datatype can make itself recursive when it is a `DataType::Record` referring to itself.

Possibly some more things we need to look at carefully.

Implement a base schema validation step in the validator for these things.

PR #206, opened 2026-06-04
feat: add schema builder & visitor API

### LINKED ISSUE #198: Minimize schema traversal in validation
In `quent-constraint`, each constraint will traverse the schema independently of other constraints.

Consider providing an interface for constraints to piggy back onto a single pass for all constraints (that want that).

A solution for this could also include a mechanism to flag all misplaced annotations in one pass.

No perf issues so far, so this is just a nice to have.

johanpel:
This PR adds the means to build and visit schemas.
Closes #198 + #197 + #202 

- Constraint validation now leverages the new visitor pattern. This way, schema elements are only visited once instead of N times for N constraints to validate.
- One schema validation todo was also to check its internal consistency (internal references through Record datatypes), which has been added as a default step in validation.
- Validation now delivers a report for consumers to do whatever they want with it.
- The in-memory schema elements now hold their constituents through maps retaining insertion order (using `indexmap` with a fast hasher for short strings) so we can do fast lookups.
- Schema fields are now private and are now completely read-only through accessor functions.
- Migrates schema errors to use `thiserror`.
- Deduplicate some minor common logic used in built-in constraints.
- Add a test-util feature to schema to be able to build test schemas ergonomically. This is opt-in and should just be used for tests since these functions can panic.


crates/constraints/src/lib.rs
```diff
@@ -32,145 +35,99 @@ pub trait Constraint {
...
-    /// Register a [`Constraint`] to be validated.
-    pub fn try_with<C: Constraint + 'static>(mut self, constraint: C) -> Result<Self, Error> {
-        match self.constraints.entry(C::NAME) {
-            Entry::Occupied(_) => Err(Error::DuplicateConstraint(C::NAME)),
-            Entry::Vacant(entry) => {
-                entry.insert(Box::new(move |schema: &Schema| constraint.validate(schema)));
-                Ok(self)
-            }
-        }
+pub fn validate<C: Constraints + Default>(schema: &Schema) -> Report<C::Output> {
```
mbrobbel: Maybe move `Default` requirement to `Constraints`?

crates/constraints/src/utils.rs
```diff
@@ -0,0 +1,15 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared helpers for constraint implementations.
+
+use std::fmt::Display;
+
+/// Join errors into a bulleted, newline-separated list.
+pub fn join_errors<E: Display>(errors: &[E]) -> String {
```
mbrobbel: Nit: this doesn't seem to be specific for errors?

crates/constraints/tests/validator.rs
```diff
@@ -1,66 +1,60 @@
...
     }
 }
+impl Constraint for NoopB {
+    const NAME: &'static str = "b";
+}
 
-// A constraint reports failures through any error type; here a minimal one.
+// A minimal error type for a failing constraint.
 #[derive(Debug)]
 struct Boom(&'static str);
```
mbrobbel, line 59: 🌳

crates/ref-target/src/lib.rs
```diff
@@ -16,98 +19,60 @@ pub struct RefTarget {
...
-                        &schema.entities,
-                        &mut errors,
-                    );
+    fn visit(&mut self, cursor: &Cursor) {
+        let Element::DataType(DataType::EntityRef { annotations, .. }) = cursor.current() else {
+            return;
+        };
+        let Some(constraint) = annotations.constraint(RefTargetConstraint::NAME) else {
+            return;
+        };
```
mbrobbel: fyi https://rust-lang.github.io/rfcs/2497-if-let-chains.html

crates/schema/src/schema/metadata.rs
```diff
@@ -8,11 +8,31 @@
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Metadata {
     /// The name of the metadata entry.
-    pub name: String,
+    name: String,
```
mbrobbel, line 11: So this can be empty too?

crates/schema/src/schema/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+pub mod data_type;
+pub mod entity;
+pub mod event;
+pub mod field;
+pub mod identifier;
+pub mod metadata;
+pub mod record;
+
+/// Container type for named elements.
+pub(crate) type Map<K, V> = indexmap::IndexMap<K, V, rustc_hash::FxBuildHasher>;
```
mbrobbel, line 21: Why `rustc_hash::FxBuildHasher`?

crates/schema/src/builder.rs
```diff
@@ -0,0 +1,273 @@
...
+
+/// Error returned while assembling a schema element.
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
+pub enum BuilderError {
+    /// A name was added more than once within the same collection.
+    #[error("duplicate name \"{0}\"")]
+    DuplicateName(String),
+}
+
+fn insert_unique<K, V>(map: &mut Map<K, V>, key: K, value: V) -> Result<(), BuilderError>
```
mbrobbel, line 28: fyi https://docs.rs/indexmap/latest/indexmap/map/enum.Entry.html

crates/schema/src/test_utils.rs
```diff
@@ -0,0 +1,66 @@
...
+
+//! Helpers for constructing schema elements in tests.
+//!
+//! These helpers are opt-in through the `test-utils` feature.
+//!
+//! # Warning
+//!
+//! The functions in this module can panic and should only be used in tests.
+
+use crate::builder::{EntityBuilder, EventBuilder, RecordBuilder, SchemaBuilder};
```
mbrobbel, line 12: `builder` is behind `cfg(feature..`. Maybe we should use separate crates instead of features?

mbrobbel (approved):

crates/schema/src/schema/mod.rs
johanpel: Don't need cryptographic hash, we already have it in the workspace, it is battle tested, and the common case is looking up things that are likely to hit. Comes out well in some benchmarks shown here: https://github.com/orlp/foldhash

crates/constraints/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/206/commits/3de79cf96a101cffc20e8c6902aad06678a5644c

crates/constraints/src/utils.rs
johanpel: https://github.com/rapidsai/quent/pull/206/commits/3de79cf96a101cffc20e8c6902aad06678a5644c

crates/schema/src/test_utils.rs
johanpel: This would force the fields of the schema elements to be `pub`, so going through the builders could be circumvented. I want to only allow this in tests through the test_utils feature. That said there's not really a good reason to put this behind a cfg so I will remove the feature flag.

crates/schema/src/builder.rs
johanpel: https://github.com/rapidsai/quent/pull/206/commits/3de79cf96a101cffc20e8c6902aad06678a5644c

crates/ref-target/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/206/commits/bff5fae10f5380d54ce18b680ff95693ee2a9de5

crates/schema/src/schema/metadata.rs
johanpel: no longer: https://github.com/rapidsai/quent/pull/206/commits/5f0337d1ffd32320aaa780470a40b71b5588870e

crates/schema/src/test_utils.rs
johanpel: https://github.com/rapidsai/quent/pull/206/commits/fabd2bca867fc15825fba4a341848542bde50f99

PR #208, opened 2026-06-08
build(update): Bump up pixi lock file to v7

9prady9:
Also adds `'requires-pixi` constraint to `pixi.toml` to ensure pixi which can support `v7` lock format can only be used.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -1,27 +1,25 @@
-version: 6
+version: 7
+platforms:
+- name: linux-64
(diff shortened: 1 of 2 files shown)


mbrobbel (commented):
We should set `requires-pixi` in `pixi.toml`: https://pixi.prefix.dev/latest/reference/pixi_manifest/#requires-pixi-optional

mbrobbel (approved):

PR #212, opened 2026-06-09
feat: add base schema constraint preventing recursive record data types

johanpel:
Adds a schema validation base constraint (one that is always validated) that disallows recursive records.
Recursive records add complexity and make storing events in certain formats (e.g. as Arrow data) hard.
There is also no compelling use case for now.

Based off #206 so that needs to go in first and then I will rebase this.


9prady9 (approved):
lgtm

PR #218, opened 2026-06-11
feat: generate base event and record types from schema

johanpel:
# Description

This adds `quent-instrumentation-build` (crate name following a pattern that users may know from `prost-build` or `tonic-build`).

This crate will ultimately generate the full barebones instrumentation library, based solely on schema input.

For now, it only generates record types and entity event types from the schema.
Follow-up PRs will add a full `<Schema>Context` which deals out `<Entity>Observer` which deals out `<EntityHandle>`.

When this base instrumentation lib generation is completed, Rust and cross-language bindings sugar for built-in constraints (e.g. FSM typestate pattern API) can be resprinkled on top.

## Related Issues

Part of #191


9prady9 (approved):
lg

PR #219, opened 2026-06-11
feat: generate TS bindings for schema

johanpel:
# Description

This PR adds a small @quent/schema TS lib generated from the Rust schema definition using TS-RS.

Not used anywhere yet, but will be useful later when the UI starts rendering some basic schema info.

## Related Issues

Nice to have for #191


cmatzenbach (approved):
Tested both generation methods, worked great

PR #225, opened 2026-06-14
fix: timeline cache tiny zoom window arithmetic

fallintoplace:
## Summary

Clamp timeline cache zoom levels so tiny viewport windows cannot produce zero-duration chunks.

## Details

The cache used `engine_duration / zoom_level` after deriving `zoom_level` from the viewport duration. For very small windows, the zoom level could exceed the engine duration, which made `chunk_duration` zero and later divisions panic.

This change keeps the existing canonical chunk calculations and clamps `determine_zoom_level()` into `1..=total_duration.max(1)` using saturating multiplication. That guarantees `chunk_duration` is at least 1 whenever the engine duration is nonzero.

It also adds regression coverage for tiny-window single and bulk timeline cache requests.

## Validation

- `cargo test -p quent-query-engine-server timeline_cache`
- `cargo test -p quent-query-engine-server`
- `cargo clippy -p quent-query-engine-server --all-targets -- -D warnings`

### DIFF
--- a/domains/query_engine/server/src/timeline_cache.rs
+++ b/domains/query_engine/server/src/timeline_cache.rs
@@ -498,7 +498,8 @@ fn determine_zoom_level(view_duration: TimeNanoSec, total_duration: TimeNanoSec)
     if view_duration == 0 {
         return 1;
     }
-    ((total_duration * TARGET_CHUNKS_PER_VIEW) / view_duration).max(1)
+    (total_duration.saturating_mul(TARGET_CHUNKS_PER_VIEW) / view_duration)
+        .clamp(1, total_duration.max(1))
 }
 
 /// Compute chunk geometry from engine metadata and the current viewport.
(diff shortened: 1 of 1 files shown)


domains/query_engine/server/src/timeline_cache.rs
```diff
@@ -513,7 +515,64 @@ fn determine_zoom_level(view_duration: TimeNanoSec, total_duration: TimeNanoSec)
     if view_duration == 0 {
         return 1;
     }
-    ((total_duration * TARGET_CHUNKS_PER_VIEW) / view_duration).max(1)
+    ((total_duration.saturating_mul(TARGET_CHUNKS_PER_VIEW)) / view_duration)
+        .max(1)
+        .min(total_duration.max(1))
```
johanpel: I think the entire problem is fixed by simply clamping here, could be written as:

```suggestion
(total_duration.saturating_mul(TARGET_CHUNKS_PER_VIEW) / view_duration)
    .clamp(1, total_duration.max(1))
```

This guarantees chunk_duration >= 1, so `compute_chunk_duration`, `chunk_bounds`, and the overflow plumbing aren't needed.

johanpel (commented):
Thanks for spotting this issue.

Could you please simplify the fix?

johanpel:
Thanks @fallintoplace !

johanpel (approved):

PR #226, opened 2026-06-14
fix: instrumentation context runtime blocking

fallintoplace:
## Summary

Make `Context::try_new` and `Context` teardown safe when called from inside an existing Tokio runtime.

## Details

`Context::try_new` detected `Handle::try_current()` and then called `handle.block_on(create_exporter(...))`. Tokio does not allow blocking the current runtime from a runtime worker thread, so async callers could panic during context construction. `Drop` had the same shape while waiting for the forwarder and flushing the exporter.

This keeps the existing synchronous API and preserves the original runtime behavior: contexts reuse an existing Tokio runtime when one is present, and only create an owned runtime when no runtime exists. When construction or teardown happens from inside Tokio, the blocking `Handle::block_on` work runs on a helper thread instead of the async worker thread. Noop contexts still short-circuit without creating or using a runtime.

Regression coverage now exercises:

- creating an exporter-backed context inside an existing Tokio runtime, without allocating an owned runtime
- dropping an exporter-backed context inside an existing Tokio runtime

## Validation

- `cargo test -p quent-instrumentation`
- `cargo clippy -p quent-instrumentation --all-targets -- -D warnings`
- `cargo test`

Note: the broad `cargo test` run emitted existing C++ bridge build warnings from macOS `ar -D`, but completed successfully.

### DIFF
--- a/crates/instrumentation/src/lib.rs
+++ b/crates/instrumentation/src/lib.rs
@@ -7,9 +7,12 @@ use quent_events::Event;
 use quent_exporter::{ExporterOptions, create_exporter};
 use quent_exporter_types::Exporter;
 use serde::Serialize;
-use std::sync::{
-    Arc,
-    atomic::{AtomicBool, Ordering},
+use std::{
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+    thread,
 };
 use tokio::{
     runtime::{Handle, Runtime},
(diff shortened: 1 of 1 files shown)


crates/instrumentation/src/lib.rs
```diff
@@ -201,35 +195,109 @@ where
...
+{
+    // `Runtime::block_on` panics from a Tokio worker thread, so keep the sync
+    // constructor API by doing the blocking exporter setup on a helper thread.
+    if Handle::try_current().is_ok() {
+        let join_result = thread::spawn(move || {
+            let exporter = runtime
+                .block_on(create_exporter(kind, id))
+                .map_err(|e| e.to_string())?;
+            Ok::<_, String>((runtime, exporter))
+        })
```
dhruv9vats, line 234: Thanks for the PR @fallintoplace!
Can we not just spawn a new thread as here on an existing runtime if one exists instead of creating another runtime? Also, can we consider spawn_blocking here?

johanpel:
Closing this, should be solved in refactor #241

Issue #229, opened 2026-06-16
Give each observer its own exporter


johanpel:
For #191 we want to split out exporters across observers rather than having a single exporter. This gives us the ability to more easily address concerns around event volume (e.g. volumous event streams go to different types of sinks or can be no-opped completely), in-flight analysis (e.g. for live application progress use cases), and exporting (e.g. to Parquet files through Narrow).

PR #231, opened 2026-06-16
feat(exporter): embed build-time provenance header in artifacts

mbrobbel:
File exporters (`ndjson`, `msgpack`, `postcard`) write an `ArtifactInfo` header as the first record of each artifact: the quent framework's git/build info plus the model's identity (name, cargo package, Rust type-path, kind) and the git of the crate that defines it. Downstream tools (e.g. `quent-open`) read it to locate/check out the producing crate and pick the type to build a viewer, without a hand-maintained config.

- New `quent-build-info` crate: `BuildInfo`/`ModelInfo`/`ArtifactInfo` types + a `build.rs` capturing the quent repo's git (degrades to "unknown" without .git).
- Model identity is derived in `create_exporter` from `std::any::type_name::<T>()` (type path + name); the cargo package and the model crate's own git come from a per-model `ModelSource` impl that `model!` generates. Out-of-repo models (e.g. sirius) get full support by using `model!` and adding a one-line `emit_source()` `build.rs`; in-repo models fall back to quent's git. Provenance stays off the call sites — only a `T: ModelSource` bound is threaded — so the codegen bridges and collector client are untouched.
- `ndjson` uses a `__quent_header__` sentinel line; `msgpack`/`postcard` prepend a `QNT1` + format-byte magic before a length-prefixed header record. Importers skip the header and tolerate legacy headerless files; `read_header(_from_bytes)` exposes the header without consuming events.


johanpel (commented):
Before I go into detail, have you considered writing the `ModelInfo` into its own file alongside events? E.g. an `<exporter path>.ndjson` would be accompanied by an `<exporter path>.qmi`.

This would allow us to have just one implementation for all exporters. In cases where we use an exporter format that is common such as `ndjson`, it might help to simplify third-party event imports (e.g. some custom analysis script) as there will only be one top-level object type.

mbrobbel:
> Before I go into detail, have you considered writing the `ModelInfo` into its own file alongside events? E.g. an `<exporter path>.ndjson` would be accompanied by an `<exporter path>.qmi`.
> 
> This would allow us to have just one implementation for all exporters. In cases where we use an exporter format that is common such as `ndjson`, it might help to simplify third-party event imports (e.g. some custom analysis script) as there will only be one top-level object type.

Yes, I have considered this. But I think having everything in one file is much easier, because you don't have to remember or think about sharing the sidecar file, and tools don't need a mechanism to discover the corresponding sidecar file.

johanpel:
> Yes, I have considered this. But I think having everything in one file is much easier, because you don't have to remember or think about sharing the sidecar file, and tools don't need a mechanism to discover the corresponding sidecar file.

There are plans to change the architecture such that each observer gets its own exporter, see:
https://github.com/rapidsai/quent/issues/229
which will typically mean a trace will consist of multiple files already.

What are your thoughts on this?

mbrobbel:
> > Yes, I have considered this. But I think having everything in one file is much easier, because you don't have to remember or think about sharing the sidecar file, and tools don't need a mechanism to discover the corresponding sidecar file.
> 
> There are plans to change the architecture such that each observer gets its own exporter, see: #229 which will typically mean a trace will consist of multiple files already.
> 
> What are your thoughts on this?

In that case the sidecar file approach makes more sense I think?

johanpel:
> In that case the sidecar file approach makes more sense I think?

:+1:

https://github.com/rapidsai/quent/pull/230 now provides a path for that sidecar file to live and be written once from an instrumentation libraries' `Context`.

PR #233, opened 2026-06-16
fix!: tighten the `UiAnalyzer` timeline-filter contract

johanpel:
# Description

Make the timeline filter types a fixed part of the `UiAnalyzer` contract instead of application-specifiable associated types, and clarify a related instrumentation gotcha.

### Changes
- **Fixed contract:** removed the `TimelineGlobalParams`/`TimelineParams` associated types from `UiAnalyzer`. The timeline methods now take concrete `QueryFilter`/`OperatorFilter`, which live in `quent_query_engine_ui`. Dropped the now-redundant generic bounds across the server.
- **Rename:** `TaskFilter` → `OperatorFilter` everywhere (Rust, generated TS bindings, and frontend consumers); removed the stale `TaskFilter.ts`.
- **Tests:** reworked the `timeline_cache` test analyzer to the fixed contract — per-entry test variation (`series_offset`) moved onto `TestAnalyzer`, cache-key separation restored via distinct per-entry `resource_id`.
- **Docs:** the generated `{Name}Context::try_new` doc no longer suggests a random `Uuid::now_v7()` for the context id; it now states the id must equal the root entity's id and why. A `TODO` marks type-system enforcement for the upcoming refactor.

### Notes
- Frontend `pnpm typecheck` passes; Rust builds + `timeline_cache` tests green; clippy/fmt clean.


dhruv9vats (approved):
LGTM, thanks!

Issue #234, opened 2026-06-16
`quent-open`: bootstrap a quent viewer for quent telemetry data


mbrobbel:
I would like to propose to introduce a new binary crate `quent-open`.

The goal is to let users quickly spawn a quent viewer for their quent telemetry data.

This crate builds the required server and ui components on the fly by bootstrapping a small binary crate and running it. It will also spawn the browser.

Initially it would work with local files (you point it to a directory with telemetry data), but we can support other importer methods (via a trait) to allow users to add additional methods (via a runtime config). For example, if users collect benchmark run results with telemetry data using a remote service, an importer could take a benchmark run and fetch the corresponding telemetry data from the remote service to render locally. The runtime config would capture where `quent-open` can find the importer.

Initially we would also use the runtime config to map engines to their `UiAnalyzer` trait implementation. Later we can move that information to be included directly in the exported files or via a sidecar file (see #231), so that `quent-open` can fetch and build the matching implementation.

Later we can also provide this functionality as library if users want to embed this functionality.

mbrobbel:
This is now available https://github.com/rapidsai/quent/tree/main/crates/open

PR #235, opened 2026-06-16
feat(exporter): write build-time provenance to a model.qmi sidecar

mbrobbel:
Emit artifact provenance so downstream tools (e.g. quent-open) can locate and check out the producing crate and pick the Rust type to build a viewer, without a hand-maintained config.

Rather than embedding a header in every artifact (which forced per-format header machinery and header-skipping importers), the instrumentation Context writes a single `model.qmi` JSON sidecar into its per-context directory, once per context. With one exporter per observer planned (#229) this keeps provenance to one file per trace and leaves event streams pure.

- New `quent-build-info` crate: `BuildInfo`/`ModelInfo`/`ArtifactInfo` types, `ModelSource`, `model_info`, and a `build.rs` capturing the quent repo git (degrades to "unknown" without .git). `ArtifactInfo::write_sidecar` writes `model.qmi` as pretty JSON.
- `model!` generates a per-model `ModelSource` impl so out-of-repo crates record their own package and git (via `emit_source()` in their build.rs); in-repo models fall back to quent's git. Only a `T: ModelSource` bound is threaded — exporters/importers and call sites are otherwise untouched.
- The collector server writes the sidecar once per application when it persists collector-routed events.

Closes #231.
Contributes to #234.

### DIFF
--- a/crates/build-info/src/git.rs
+++ b/crates/build-info/src/git.rs
@@ -1,0 +1,131 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Git capture shared between this crate's `build.rs` (via `include!`) and the
+// `emit_source` build-script helper (via `mod git`). Kept dependency-free and
+// free of `//!` inner-doc comments so it is valid in both contexts.
+
+use std::path::Path;
+use std::process::Command;
+
+/// Raw git fields captured from a working tree. A field is `None` when it could
+/// not be determined (e.g. the source is not a git checkout), so the build never
+/// fails and absent provenance stays distinguishable from a real value.
+pub struct RawGit {
+    pub commit: Option<String>,
+    pub branch: Option<String>,
+    pub dirty: Option<bool>,
+    pub remote: Option<String>,
+    pub built_at: Option<String>,
+    /// Absolute path to the `.git` directory, if this is a git working tree.
+    pub git_dir: Option<String>,
+}
+
+fn run(dir: &Path, args: &[&str]) -> Option<String> {
+    let output = Command::new("git")
+        .args(args)
+        .current_dir(dir)
+        .output()
+        .ok()?;
+    if !output.status.success() {
+        return None;
+    }
+    let value = String::from_utf8(output.stdout).ok()?.trim().to_string();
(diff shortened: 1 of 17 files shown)


crates/build-info/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+}
+
+/// Per-model provenance hook, generated by `model!` for each event enum.
+///
+/// Lets exporters record the model's cargo package and the git of the crate that
+/// defines it — including out-of-repo crates — because the impl's `env!` /
+/// `option_env!` are evaluated in that crate. The Rust type path and model name
+/// are derived separately from [`std::any::type_name`], so no provenance has to
+/// be threaded through `Context`/`create_exporter` call sites.
```
johanpel: Nit: please explain what the trait represents rather than mentioning its use sites and impl details

crates/build-info/src/git.rs
```diff
@@ -0,0 +1,131 @@
...
+pub fn capture(dir: &Path) -> RawGit {
+    RawGit {
+        commit: run(dir, &["rev-parse", "HEAD"]),
+        branch: run(dir, &["rev-parse", "--abbrev-ref", "HEAD"]),
+        dirty: run(dir, &["status", "--porcelain"]).map(|status| !status.is_empty()),
+        remote: run(dir, &["remote", "get-url", "origin"]).map(sanitize_remote),
+        built_at: run(dir, &["log", "-1", "--format=%cI"]),
+        git_dir: run(dir, &["rev-parse", "--absolute-git-dir"]),
+    }
+}
```
johanpel, line 64: It might be nice to use [`gix`](https://github.com/gitoxidelabs/gitoxide) here. (Fine to do later)

johanpel (approved):
Looks good, this also sets up a path to thread the schema through later. Just two nits.

crates/build-info/src/git.rs
mbrobbel: I did consider this, but this is pretty large build dependency, and there is no real benefit over using `git` directly for this use case.

crates/build-info/src/lib.rs
mbrobbel: https://github.com/rapidsai/quent/compare/94e592f6a0a07f02dde9a3e38f236d6eeb91c8ee..38862827c2f840217712f081376850b9e099da3d

PR #217, opened 2026-06-11
feat: long entity pagination

johanpel:
# Description

WIP, do not review or merge yet.

This PR will add the option for the UI to limit the number of long entities that use a resource are transported to it in full from the analyzer.

## Related Issues

#214

Issue #214, opened 2026-06-10
Long entity display v1


johallar:
Single entity display in timelines is in a v0 POC state, this issue should capture work to refactor and improve to get it to a v1 state

PR #240, opened 2026-06-17
feat(analyzer): add entity listing endpoint and remove long entities from resource timeline endpoint

johanpel:
# Description

WIP

## Related Issues

Part of #214
Alternative to: #217


johanpel:
This is getting very stale, closing and will redo when needed

Issue #40, opened 2026-03-06
Application (and workload) history / indexing


johanpel:
Applications run workloads, where the application (e.g. a query engine cluster) can be long-lived, running MANY workloads (e.g. a query).

We're currently mostly focused on small (relatively speaking) benchmarks on short-lived applications in the sense that we always need to inspect the full data in order to get some metadata out that allows us to know more about an application/workload than just a UUID. This is useful for top-level exploration of all data that is in long-term telemetry storage, but the way it's built right now is that we're assuming a small amount of telemetry to be in the storage. 

We need some kind of history / indexing system for application instances and workloads, so we don't have to inspect data directly.

This will also allow for a distributed use case where data capture is completely decentralized e.g. in order to minimize network pressure at workload run-time in a cluster, except for a small message to a history service which tells it where the data is captured. At analysis-time, the history service can then retrieve the data ad hoc. Note this contrasts with the _typical_ approach of e.g. OTel where collection always happens (although theoretically it could be done decentralized) and pressure is put on the network.

PR #241, opened 2026-06-22
feat!: give each observer its own exporter

### LINKED ISSUE #229: Give each observer its own exporter

johanpel:
# Description

This PR is of mainly instrumentation infrastructure / support crates based on two issues:

1. Have one exporter per observer instead of one exporter per context (see #229)
2. Make the context id unrelated to the id of the root entity (e.g. the engine id in the query engine domain model). 

2 allows every context to write events without the need to synchronize on the root id or to consider the root entity as something special versus other entities. While this would be ideally propagated back to the modeling API, #191 will remove that in its current form entirely anyway, so it is left as is for now so upstream model definitions experience no breaking changes twice in a short period of time.

I initially tried to deal with 1 and 2 separately, but since this touches a lot of the same code across many layers it seemed best to tackle these in one pass.

The consequences are as follows. In this description, only {App}Context and {Entity}Observer/Handle are user-facing, plain Context and Observer/Handle mentions are not user facing, but are the model-agonistic backing structures for their generated counterparts.

- There is no longer a need for an "umbrella" event type representing all events of an application event model. This is for now still generated by the model macros because the analyzer side still relies on it. Removing that is intentionally not done in this PR to reduce the scope.
- The quent_instrumentation::Context (not user facing) is now mainly responsible for ensuring there is an async runtime that observers can work with, it no longer owns the exporter, and to provide a bridge from sync <-> async code.
- A generated {App}Context now concurrently writes the model provenance sidecar file, constructs (+connects) exporters and observers, and blocks until all of that is ready.
- Each Observer now manages its own Exporter (which lives in an async forwarder task that it cancels on drop)
- Each {Entity}Observer deals out {Entity}Handles for entity instances which hold an Arc to the inner Observer, which ensures everything is kept alive as long as a handle exists.

This means as long as there are any live handles, even when all Observers or Context is dropped, events can be emitted. Also, constructing or dropping Context / Observer / Handle can be done from both synchronous and asynchronous code as long as that code is not using a `current-thread` runtime (also raised in #226). This PR adds tests for all flavors of runtime environments for the Context to live in.

- Filesystem exporters now write in the following directory tree:

```
  - <context uuid>
    - <entity_name>
      - <uuid>.<extension>  (using a uuid here is just a non-enforced convention 
                             that mainly batching no-append file format exporters 
                             will benefit from as they can just generate a new uuid 
                             for each batch of events as file name).
    - <another entity name>
      - <uuid>.<extension>
      - <uuid>.<extension>
      - ...
    - model.qmi        (sidecar file) 
```

- For the query engine domain, since a (distributed) engine's events could now be spread out across multiple contexts / directories, listing engines is done by first scanning all engine events across the entire data source folder to obtain all engine ids. Then, all worker events are scanned to figure out which workers were spawned on behalf of those engines. For now it is assumed that these entities are directly tied to all event-producing processes of the engines, such that their instrumentation contexts together capture all events for an entire engine model. This is a rather brittle assumption that should not be relied on in follow-up work. I intend to address this with a generated importing stack that leverages the quent-ref-tree constraint after #191 is done, plus some sort of indexing service hinted at in #40.

## Related Issues

Closes #229

### Generated event type + stream name
```rust
pub enum FileStatsEvent { Checksum(Checksum), Decompressed(Decompressed) }

impl quent_model::EntityEvent for FileStatsEvent {
    const NAME: &'static str = "file_stats";   // exporter subdir / wire tag / ingest key
}
```

### `{App}Context` (the user entry point), the relevant slice
```rust
pub struct AppContext {
    file_stats: FileStatsObserver,     // ... one field per entity
    _inner: quent_model::Context,
}

impl AppContext {
    pub fn try_new(exporter: Option<…ExporterOptions>) -> Result<Self, …> {
        let inner = quent_model::Context::try_new(exporter)?;   // sync: resolve runtime only
        Self::assemble(inner)
    }

    // The single sync→async bridge: sidecar + every observer built concurrently,
    // blocked once.
    fn assemble(inner: quent_model::Context) -> Result<Self, …> {
        let (file_stats, …) = inner.block_on(async {
            let (_sidecar, file_stats, …) = quent_model::tokio::try_join!(
                async { inner.write_sidecar(<App as ModelSource>::model_info()).await; Ok(()) },
                inner.observer::<FileStatsEvent>(),     // ... one per entity
            )?;
            Ok((file_stats, …))
        })?;
        Ok(Self { file_stats: FileStatsObserver::new(file_stats), …, _inner: inner })
    }

    pub fn file_stats_observer(&self) -> FileStatsObserver { self.file_stats.clone() }   // cheap Arc clone
}
```

### `{Entity}Observer` facade → `{Entity}Handle`
```rust
pub struct FileStatsObserver { inner: Arc<quent_model::Observer<FileStatsEvent>> }

impl FileStatsObserver {
    pub fn new(observer: quent_model::Observer<FileStatsEvent>) -> Self { Self { inner: Arc::new(observer) } }
    pub fn send(&self, event: quent_model::Event<FileStatsEvent>) { self.inner.send(event); }  // pre-built (collector path)
    pub fn create(&self, id: Uuid) -> FileStatsHandle {
        FileStatsHandle { id, inner: self.inner.clone() }   // handle co-owns the observer
    }
}

pub struct FileStatsHandle { id: Uuid, inner: Arc<quent_model::Observer<FileStatsEvent>> }

impl FileStatsHandle {
    pub fn uuid(&self) -> Uuid { self.id }
    pub fn checksum(&self, event: Checksum) {
        self.inner.emit(self.id, FileStatsEvent::Checksum(event));   // Observer::emit -> EventSender -> forwarder
    }
    pub fn decompressed(&self, event: Decompressed) {
        self.inner.emit(self.id, FileStatsEvent::Decompressed(event));
    }
}
```

Emit path: `AppContext.file_stats_observer().create(id).checksum(..)` → `Observer::emit` → mpsc → forwarder task → exporter.

### Read/route side (collector + analyzer), same entity
```rust
// CollectorSink::ingest (server replays a received stream)
if entity == <FileStatsEvent as EntityEvent>::NAME {
    let e: Event<FileStatsEvent> = ciborium::from_reader(event)?;
    self.file_stats.send(e);                 // into the (already built) observer
    return Ok(());
}

// {Model}::import_events (analyzer reconstruction), per entity:
let path = dir.join(<FileStatsEvent as EntityEvent>::NAME);   // <ctx>/file_stats
if path.is_dir() {
    let importer = create_importer::<FileStatsEvent>(&FileSystem { format, path })?;
    streams.push(importer.map(|e| Event::new(e.id, e.timestamp, AppEvent::from(e.data))));
}
```

`AppEvent` (the umbrella enum) + its `From<{Entity}Event>` impls are still generated solely for this analyzer reconstruction; nothing on the capture path uses them.

### DIFF
--- a/crates/collector/client/src/lib.rs
+++ b/crates/collector/client/src/lib.rs
@@ -52,7 +52,11 @@ impl<T> Client<T>
 where
     T: Serialize + Send + 'static,
 {
-    pub async fn new(source_context_id: Uuid, address: String) -> CollectorResult<Client<T>> {
+    pub async fn new(
+        source_context_id: Uuid,
+        entity_type: &str,
+        address: String,
+    ) -> CollectorResult<Client<T>> {
         debug!("connecting to {address}");
         // Try to connect.
         // TODO(johanpel): figure out whether this can also go through health check
@@ -175,7 +179,9 @@ where
 
         debug!("opening stream ...");
 
-        // Identify this stream so the collector reproduces its events under the id.
+        // Identify this stream so the collector reproduces its events under the
+        // id, and tag it with the entity type so the collector routes the batch
+        // to the matching entity observer.
         let mut req = Request::new(ReceiverStream::new(grpc_receiver));
         req.metadata_mut().insert(
             "source-context-id",
@@ -183,6 +189,10 @@ where
                 .to_string()
                 .parse()
                 .expect("valid metadata value"),
+        );
+        req.metadata_mut().insert(
+            "entity-type",
+            entity_type.parse().expect("valid metadata value"),
         );
 
         let mut cloned_client = client.clone();
--- a/crates/collector/client/src/lib.rs
+++ b/crates/collector/client/src/lib.rs
@@ -25,8 +25,8 @@ use quent_collector_proto::{CollectEventRequest, collector_client::CollectorClie
 
 /// A sink for serialized per-entity event streams.
 pub trait CollectorSink {
-    /// Decode the serialized `event` belonging to the entity stream named
-    /// `entity`, and record it.
+    /// Ingest a serialized `event` belonging to the entity event stream named
+    /// `entity`.
     fn ingest(&self, entity: &str, event: &[u8]) -> Result<(), Box<dyn std::error::Error>>;
 }
 
--- a/crates/collector/client/src/lib.rs
+++ b/crates/collector/client/src/lib.rs
@@ -3,12 +3,12 @@
 
 //! A gRPC-based client that can send [`Event`]s to a collector.
 
+use std::sync::Mutex;
 use std::time::Duration;
 
 use quent_events::Event;
 use serde::Serialize;
 use tokio::{
-    runtime::Handle,
     select,
     sync::mpsc::{self, Receiver, Sender},
     task::JoinHandle,
@@ -40,6 +40,12 @@ pub enum CollectorError {
     Tonic(#[from] tonic::transport::Error),
     #[error("RPC error: {0}")]
     GRPC(#[from] Status),
+    #[error("invalid `{key}` metadata value: {source}")]
+    Metadata {
+        key: &'static str,
+        #[source]
+        source: tonic::metadata::errors::InvalidMetadataValue,
+    },
 }
 
 pub type CollectorResult<T> = std::result::Result<T, CollectorError>;
@@ -50,9 +56,10 @@ pub struct Client<T> {
     _grpc_client: CollectorClient<Channel>,
     event_sender: Sender<Event<T>>,
     cancellation_token: CancellationToken,
-    events_sender_handle: Option<JoinHandle<()>>,
-    events_collector_handle: Option<JoinHandle<()>>,
-    runtime_handle: Handle,
+    // `Mutex<Option<..>>` so the join handles can be taken from `&self` in
+    // `shutdown`; awaited there, never under the lock.
+    events_sender_handle: Mutex<Option<JoinHandle<()>>>,
+    events_collector_handle: Mutex<Option<JoinHandle<()>>>,
 }
 
 impl<T> Client<T>
@@ -195,11 +202,19 @@ where
             source_context_id
                 .to_string()
                 .parse()
-                .expect("valid metadata value"),
+                .map_err(|source| CollectorError::Metadata {
+                    key: "source-context-id",
+                    source,
+                })?,
         );
         req.metadata_mut().insert(
             "entity-type",
-            entity_type.parse().expect("valid metadata value"),
+            entity_type
+                .parse()
+                .map_err(|source| CollectorError::Metadata {
+                    key: "entity-type",
+                    source,
+                })?,
         );
 
         let mut cloned_client = client.clone();
@@ -212,10 +227,8 @@ where
             _grpc_client: client,
             event_sender,
             cancellation_token,
-            // Safety: this fn must be called from a tokio runtime.
-            runtime_handle: Handle::current(),
-            events_sender_handle: Some(events_sender_handle),
-            events_collector_handle: Some(events_collector_handle),
+            events_sender_handle: Mutex::new(Some(events_sender_handle)),
+            events_collector_handle: Mutex::new(Some(events_collector_handle)),
         })
     }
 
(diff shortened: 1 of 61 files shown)


crates/collector/client/src/lib.rs
```diff
@@ -43,16 +56,21 @@ pub struct Client<T> {
     _grpc_client: CollectorClient<Channel>,
     event_sender: Sender<Event<T>>,
     cancellation_token: CancellationToken,
-    events_sender_handle: Option<JoinHandle<()>>,
-    events_collector_handle: Option<JoinHandle<()>>,
-    runtime_handle: Handle,
+    // `Mutex<Option<..>>` so the join handles can be taken from `&self` in
+    // `shutdown`; awaited there, never under the lock.
```
mbrobbel, line 60: Why not use `&mut self` (or `self`) for `shutdown`?

crates/exporter/collector/src/lib.rs
```diff
@@ -4,44 +4,49 @@
...
 impl<T> CollectorExporter<T>
 where
-    T: Serialize + Send + 'static,
+    T: Serialize + Send + EntityEvent + 'static,
 {
+    /// `source_context_id` identifies this stream to the collector, which
+    /// reproduces the source's output under that id.
     pub async fn try_new(
-        options: CollectorExporterOptions,
+        address: String,
```
mbrobbel: Maybe consider a stronger type here?

crates/instrumentation/src/lib.rs
```diff
@@ -79,178 +78,335 @@ impl<T> EventSender<T> {
...
+                }
             }
         }
+    }
+
+    /// Write the model provenance sidecar file into the context's filesystem
+    /// exporter directory.
+    ///
+    /// If no filesystem exporter is configured, then this is a no-op.
+    pub async fn write_sidecar(&self, model: ModelInfo) {
```
mbrobbel: Shouldn't this be non-`pub` and used in `Context` construction?

crates/instrumentation/src/lib.rs
```diff
@@ -79,178 +78,335 @@ impl<T> EventSender<T> {
...
 
-    /// Identity of this context, generated on construction.
-    pub fn id(&self) -> Uuid {
-        self.id
+/// Provides an event pipeline to "observe" events of one *type* of entity `T`
+/// and export them.
+///
+/// Instrumented application code should not interact with this type directly
+/// unless they have a very special reason. Instead, it interacts with the
+/// generated observer only.
```
mbrobbel, line 317: Maybe add why or use `doc(hidden)`?

crates/instrumentation/src/lib.rs
```diff
@@ -79,178 +78,335 @@ impl<T> EventSender<T> {
...
-    /// Identity of this context, generated on construction.
-    id: Uuid,
-    handle: Option<Handle>,
-    events_sender: EventSender<T>,
-    exporter: Option<Arc<dyn Exporter<T>>>,
-    cancellation_token: CancellationToken,
-    forwarder_handle: Option<JoinHandle<()>>,
+/// The runtime an active context's observers run on.
+#[derive(Clone)]
+enum BackendRuntime {
```
mbrobbel, line 83: I would like to suggest always owning the `Runtime`, but something for a follow-up.

mbrobbel (approved):

crates/instrumentation/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/241/commits/bcb32d5a13790027b88d8a228d2b8d7a8793f2d0

crates/exporter/collector/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/241/commits/c456bd218a93685b2d75d43cc3acfd4ed693e9e2

crates/instrumentation/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/241/commits/c456bd218a93685b2d75d43cc3acfd4ed693e9e2

crates/collector/client/src/lib.rs
johanpel: Great question, this is due to the signature of the exporter trait fns that can be changed since now the exporters are fully owned by the forwarder tasks. I will address this in: https://github.com/rapidsai/quent/issues/245

crates/instrumentation/src/lib.rs
johanpel: https://github.com/rapidsai/quent/issues/246

dhruv9vats (approved):

Issue #196, opened 2026-06-02
Introduce resource concepts as constraint


johanpel:
With the introduction of `quent-schema` in #192 as a part of #191, we can now express the concept of resources as a constraint that is serialized into a schema.

Since resource capacities as usages can only be claimed throughout a finite time window, these usages are constrained to only be present in FSM states. Since usage relate to the providing resource entity, they are a type of entity reference.

Solving this issue thus depends on #194 and #195

johanpel:
This has been completed in #387 and #251

PR #251, opened 2026-06-24
feat(resource): add quent-resource constraint metadata types

johanpel:
# Description

This PR adds the `quent-resource` crate. It only adds the types that are serialized as opaque schema metadata. Validation will follow later.

## Related Issues

#196

### DIFF
--- a/crates/resource/src/lib.rs
+++ b/crates/resource/src/lib.rs
@@ -1,0 +1,72 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! The Quent built-in resource constraint.
+
+use quent_schema::{Annotations, Identifier};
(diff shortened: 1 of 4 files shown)


crates/resource/src/lib.rs
```diff
@@ -86,8 +50,68 @@ pub enum CapacityKind {
...
     /// A total quantity processed over a usage span, e.g. bytes sent over a
-    /// channel. Dividing it by the span's duration yields a perceived rate. The
-    /// true rate may be hidden from Quent (e.g. when the work is performed
-    /// asynchronously).
+    /// channel. Dividing it by the span's duration yields the **perceived**
+    /// rate.
     Rate,
 }
+
+pub type Capacities = indexmap::IndexMap<Identifier, Capacity>;
```
mbrobbel, line 58: `pub` type without a doc comment

mbrobbel (approved):

crates/resource/src/lib.rs
johanpel: will address in the follow-up

PR #254, opened 2026-06-25
feat(exporter): add an opt-in clap exporter arg group to quent-exporter

johanpel:
Several internal binaries (the fixed QE emitter, the simulator) each hand-rolled the same --exporter/--collector-address/--output-dir parsing. This adds a reusable ExporterArgs group behind a clap feature so that duplication collapses to one place — and framework users can opt into the same flags. The collector address is now typed as http::Uri, and the filesystem path is feature-gated so collector-only builds compile.

🤖 Generated with Claude Code (https://claude.com/claude-code)

### DIFF
--- a/crates/exporter/Cargo.toml
+++ b/crates/exporter/Cargo.toml
@@ -3,14 +3,19 @@ name = "quent-exporter"
 version.workspace = true
 edition.workspace = true
 publish.workspace = true
+
 [features]
 default = ["ndjson", "msgpack", "postcard", "collector"]
-ndjson = ["dep:quent-exporter-ndjson"]
-msgpack = ["dep:quent-exporter-msgpack"]
-postcard = ["dep:quent-exporter-postcard"]
+ndjson = ["dep:quent-exporter-ndjson", "filesystem"]
+msgpack = ["dep:quent-exporter-msgpack", "filesystem"]
+postcard = ["dep:quent-exporter-postcard", "filesystem"]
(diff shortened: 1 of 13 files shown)


crates/exporter/Cargo.toml
```diff
@@ -3,14 +3,19 @@ name = "quent-exporter"
...
-ndjson = ["dep:quent-exporter-ndjson"]
-msgpack = ["dep:quent-exporter-msgpack"]
-postcard = ["dep:quent-exporter-postcard"]
+ndjson = ["dep:quent-exporter-ndjson", "filesystem"]
+msgpack = ["dep:quent-exporter-msgpack", "filesystem"]
+postcard = ["dep:quent-exporter-postcard", "filesystem"]
 collector = ["dep:quent-exporter-collector", "dep:http"]
+clap = ["dep:clap", "dep:http"]
+# Internal marker: enabled by any filesystem format. Not meant to be set directly.
+filesystem = []
```
mbrobbel, line 14: You could move this to a build script (with `cargo::rustc-cfg`).

mbrobbel (approved):

crates/exporter/Cargo.toml
johanpel: https://github.com/rapidsai/quent/pull/254/commits/5b966bd643d13a0cd0afd3ea39da9d20c8c03855

PR #255, opened 2026-06-25
ci(experiment): don't check in generated TS files

johanpel:
# Description

Stop checking generated TypeScript bindings into the repository.

Bindings are generated into `ui/generated/ts-bindings` through the dedicated `quent-simulator-ui-bindings` package. This keeps each workflow self-contained:

- `pnpm build`, `pnpm typecheck`, and development scripts generate bindings automatically.
- Cargo builds with the `ui` feature generate bindings before building and embedding the UI.
- CI can generate bindings independently for fast comparison and only reports a binding diff when one exists.
- Cargo builds without the `ui` feature perform no binding or frontend work.

UI CI jobs now generate their own inputs instead of sharing a generated-bindings artifact.

Clean-state standalone Cargo and pnpm builds were verified with no generated bindings present. Consecutive embedded Cargo builds were also verified to avoid recompilation.

_Written by Codex._

# Example

The best way I could come up with surfacing whether TS bindings were changed was to make this check not skip (it will show skipped if there were no changes):

image

And then if you go to the UI summary tab, you can inspect them:

image

Any review comments based on those changes will still have to go to the Rust source of these bindings of course (unless we bump packages and the generators change).


ui/pnpm-workspace.yaml
```diff
@@ -13,3 +13,9 @@ allowBuilds:
   esbuild: true
   msw: true
   protobufjs: true
+
+overrides:
```
johallar: I'd prefer an update to the lockfile if possible, use `pnpm update <dep>` to see if we can bump version within semver bounds before overriding like this.

johallar:
for posterity: I need to make sure that the @quent packages are able to re-export the generated types correctly when built, and also add some "pre-dev" scripts to make sure types are generated so IDEs have types and typecheck works seamlessly locally.

johanpel:
Added some commits to make the bindings compile automatically as a part of `pnpm run dev`. 

First bindings take some time to compile, but without Rust changes it's fast now:

```
# First run
Compiling quent-simulator-server ...
Finished `dev` profile ... in 17.84s
VITE v7.3.6 ready in 5197 ms

# Subsequent run (without Rust changes)
Finished `dev` profile ... in 1.05s
VITE v7.3.6 ready in 2173 ms

# After modifying an exported Rust type
Compiling quent-simulator-ui ...
Compiling quent-simulator-analyzer ...
Compiling quent-simulator-server ...
Finished `dev` profile ... in 5.91s
VITE v7.3.6 ready in 631 ms
```

johanpel:
The best way I could come up with surfacing whether TS bindings were changed was to make this check not skip (it will show skipped if there were no changes):

image

And then if you go to the UI summary tab, you can inspect them:

image

Any review comments based on those changes will still have to go to the Rust source of these bindings of course (unless we bump packages and the generators change).

johanpel:
Tested this in commit [5cb5c3e](https://github.com/rapidsai/quent/pull/255/commits/5cb5c3ef0e48128721ef1573ef5b11376405b0da)

johallar (approved):
🙇 TYTY

mbrobbel (approved):

PR #256, opened 2026-06-25
refactor(analyzer): convert FsmCollection to associated-type form

johanpel:
# Description

Replace `FsmCollection<F, T>` with a single associated `type Fsm`, and drop the
uncalled `contains_fsm_type`. Lets a collection expose exactly one FSM type and
unblocks bounding generic code on it. No behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Related Issues

Part of #214, split off from #240

PR #258, opened 2026-06-25
feat(analyzer): entity list endpoint

johanpel:
# Description

Adds `POST /api/engines/{engine_id}/entities`: a paginated query listing a
query's FSM entities ranked by longest resource-usage span (tiebroken by UUID),
to back the UI's long-entities view. Additive — no timeline changes.

Request: optional scope (resource / group), query-relative window, filters
(`entity_type_name`, `min_usage_s`), sort, page, and a per-query `OperatorFilter`.
Contract lives in `quent-ui`; a generic `entities::list_entities` ranks any
`FsmCollection` FSM, with operator filtering applied by the analyzer (kept out of
the generic filter).

Functional test over the fixed scenario captures events in memory via
`CallbackExporter` and reconstructs them with a new
`quent-simulator-instrumentation::test_utils` helper.

Depends on #256 and #257 (merged).

## Related Issues

Part of #214, isolated from #240


examples/simulator/server/build.rs
```diff
@@ -22,5 +23,8 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     <BulkTimelineRequest<QueryFilter, OperatorFilter> as TS>::export_all(&cfg)?;
     <BulkTimelinesResponse as TS>::export_all(&cfg)?;
 
+    <EntityListRequest<QueryFilter> as TS>::export_all(&cfg)?;
```
johallar: I think we need to be able to filter by operator here. Not sure if the exact right location, but the FE will want to filter long entities by operator presumably.

johanpel: Should be threaded through now: https://github.com/rapidsai/quent/pull/258/changes#diff-b6b1e0299066cb084541802cb36e0c593cbdb98719ce7cc85a7b1fdd335f4e35R131-R137

crates/ui/src/entities/request.rs
```diff
@@ -0,0 +1,137 @@
...
+                model.resource(*resource_id)?;
+                Ok([*resource_id].into_iter().collect())
+            }
+            EntityScope::ResourceGroup {
+                resource_group_id,
+                resource_type_name,
+            } => {
+                let tree = ResourceTreeNode::try_new(model, *resource_group_id)?;
+                Ok(tree
+                    .iter_leaf_ids()
```
dhruv9vats, line 41: does this mean that only leaf resources should have valid usages ?

dhruv9vats (approved):
Thanks @johanpel

crates/ui/src/entities/request.rs
johanpel: Yes, this limitation is removed as a part of #191.

The resource concept will be unrelated to hierarchical trees of entities, also see:

https://github.com/rapidsai/quent/blob/main/crates/ref-tree/src/lib.rs
https://github.com/rapidsai/quent/blob/main/crates/resource/src/lib.rs

PR #261, opened 2026-06-26
feat: record a model's analyzer package in provenance

mbrobbel:
Adds an optional `analyzer_package` to `ModelInfo`/`model.qmi` and a defaulted `ModelSource::analyzer_package()`; `model!` gains an `analyzer_package: "<crate>"` directive recording the crate that provides a model's viewer.

Part 2/7 of the `quent-open` series (#234). Stacked on #260 — review/merge bottom-up (Files-changed includes #260 until it lands).

### DIFF
--- a/crates/model-macros/src/lib.rs
+++ b/crates/model-macros/src/lib.rs
@@ -71,6 +71,7 @@ pub fn derive_resizable_resource(input: TokenStream) -> TokenStream {
 ///     name: App,
 ///     root: Cluster,
 ///     entities: { Worker, Thread, Task },
+///     analyzer: "my-analyzer", // optional: crate providing the QuentViewer
 /// }
 /// ```
 #[proc_macro]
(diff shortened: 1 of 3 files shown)


crates/model-macros/src/lib.rs
```diff
@@ -70,6 +70,7 @@ pub fn derive_resizable_resource(input: TokenStream) -> TokenStream {
 ///         Worker,
 ///         Thread,
 ///         Task,
+///         analyzer_package: "my-analyzer", // optional: crate providing the QuentViewer
 ///     }
 /// }
```
johanpel, line 75: Perhaps we can make things a bit more explicit, something like:

```
model! {
  name: App,
  root: Cluster,
  entities: {
    Worker,
    Thread,
    ...,
  },
  analyzer: "my-analyzer"
}
```

What do you think?

mbrobbel: #269?

johanpel (approved):
👍

PR #264, opened 2026-06-26
feat(open): build and serve a viewer from the sidecar

mbrobbel:
`quent-open local <ctx>` generates a standalone wrapper crate git-pinned to the sidecar's quent + analyzer commits, `cargo build`s it (embedded UI → pnpm on first build), serves via axum, opens a browser. Clone-by-sidecar — no patching.

The bind host is configurable (`--host`, default `127.0.0.1`); the wrapper crate is generated with `quote` (`main.rs`) and `cargo-manifest` (`Cargo.toml`), and the built binary is located from cargo's JSON output.

> [!NOTE]
> Opening an artifact builds and runs the repositories its sidecar names. **Source-trust gating and provenance validation land in #266** (this PR has neither in isolation) — until then, only open artifacts you trust.

Part 5/7. Stacked on #263.


crates/open/src/main.rs
```diff
@@ -1,21 +1,28 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! `quent-open` opens local Quent benchmark artifacts in an application-specific
-//! viewer. See <https://github.com/rapidsai/quent/issues/234>.
+//! Open local Quent benchmark artifacts in an application-specific viewer.
```
johanpel:
```suggestion
//! Open local Quent artifacts in an application-specific viewer.
```

crates/open/src/spec.rs
```diff
@@ -0,0 +1,260 @@
...
+        let analyzer_package =
+            info.model
+                .analyzer_package
+                .clone()
+                .ok_or_else(|| OpenError::NoAnalyzer {
+                    model: info.model.name.clone(),
+                })?;
+        Ok(Self {
+            root: root.to_path_buf(),
+            format: detect_format(root)?,
```
johanpel, line 113: Probably not something we want to do in this PR since I don't think it's a trivial change, but we may want to leave it up to the `QuentViewer` impl to detect the formats. `import_events` would just receive one URI argument.

crates/open/src/spec.rs
```diff
@@ -0,0 +1,260 @@
...
+        info
+    }
+
+    fn ctx_with_stream(name: &str, file: &str) -> tempfile::TempDir {
+        let dir = tempfile::tempdir().unwrap();
+        let entity = dir.path().join(name);
+        std::fs::create_dir_all(&entity).unwrap();
+        std::fs::write(entity.join(file), b"").unwrap();
+        dir
+    }
```
johanpel, line 200: This hardcodes the assumption we want to test into the test input. Let's base this on an actual Context + exporter write.

crates/open/src/wrapper.rs
```diff
@@ -0,0 +1,210 @@
...
+        );
+        assert_eq!(analyzer["rev"].as_str().unwrap(), "analyzercommit");
+    }
+
+    #[test]
+    fn main_rs_names_viewer_and_format() {
+        let main = main_rs(&spec());
+        assert!(main.contains("use quent_simulator_analyzer::Viewer;"));
+        assert!(main.contains("FileSystemFormat::Msgpack"));
+        assert!(main.contains("analyzer_service_router::<")); // turbofish may wrap
```
johanpel:
```suggestion
```

crates/open/src/spec.rs
mbrobbel: Agree. I'll do this in a follow-up to avoid changing the trait in this PR.

crates/open/src/spec.rs
mbrobbel: I'll do this after #273, making it a lot easier to write an integration test.

johanpel (approved):

PR #265, opened 2026-06-26
feat(open): discover all exports and serve one viewer per model

mbrobbel:
`local` recursively discovers all context dirs, groups them by build spec (one viewer per distinct model), builds in parallel, and streams each URL as it's ready; opens a browser only when there's a single viewer.

Part 6/7. Stacked on #264.

### DIFF
--- a/crates/open/src/viewer.rs
+++ b/crates/open/src/viewer.rs
@@ -164,8 +164,13 @@ async fn cargo_build(crate_dir: &Path) -> Result<PathBuf> {
         .await?;
     let status = child.wait().await?;
     if !status.success() {
+        // Compiler diagnostics are rendered into the JSON stream; cargo's own
+        // output (build scripts, the final error summary) is on stderr in the
+        // log. Include both in full so the cause is never truncated away.
+        let mut detail = rendered_diagnostics(&json);
+        detail.push_str(&std::fs::read_to_string(&log_path).unwrap_or_default());
         return Err(OpenError::Build {
-            status: format!("{status}; last output:\n{}", log_tail(&log_path)),
+            status: format!("{status}\n{detail}"),
         });
     }
     wrapper_executable(&json).ok_or_else(|| OpenError::Build {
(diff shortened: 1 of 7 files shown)


crates/open/src/viewer.rs
```diff
@@ -138,13 +186,69 @@ fn wrapper_executable(stdout: &[u8]) -> Option<PathBuf> {
...
+        // canonical paths but one link name. Keep the first and skip the rest
+        // rather than failing the whole viewer on the colliding link.
+        if !linked.insert(name.to_owned()) {
+            eprintln!(
+                "warning: ignoring duplicate context id `{}` at {}",
+                name.to_string_lossy(),
+                context.display()
+            );
+            continue;
+        }
```
johanpel: I'm not sure I'm following this, how could a tree hold two copies with the same context ID?

crates/open/src/viewer.rs
```diff
@@ -117,7 +165,7 @@ async fn cargo_build(crate_dir: &Path) -> Result<PathBuf> {
     let status = child.wait().await?;
     if !status.success() {
         return Err(OpenError::Build {
-            status: status.to_string(),
+            status: format!("{status}; last output:\n{}", log_tail(&log_path)),
```
johanpel: I would rather include the full message to not risk losing the cause. If users find it too long they can always pipe to tail themselves.

crates/open/src/viewer.rs
mbrobbel: [6127a9f](https://github.com/rapidsai/quent/pull/265/commits/6127a9f059367cf698665b21e269f044b5409afe)

crates/open/src/viewer.rs
mbrobbel: [6127a9f](https://github.com/rapidsai/quent/pull/265/commits/6127a9f059367cf698665b21e269f044b5409afe)

johanpel (approved):

PR #266, opened 2026-06-26
feat(open): validate provenance and gate untrusted sources

mbrobbel:
Opening an artifact builds+runs the git sources its sidecar names. Validate provenance (hex commit, https/ssh remote, cargo-name package — blocks injection) and gate builds on a trust allowlist (built-ins + `<config>/quent/open/trusted` + `--trust`/`--trust-all`) with confirm-on-unknown (refused non-interactively).

Part 7/7. Stacked on #265.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #269, opened 2026-06-29
feat(model-macros)!: structured `model!` syntax (name/root/entities)

mbrobbel:
Make `model!` explicit and order-flexible: labeled `name:` / `root:` / optional `entities: { … }` fields, instead of `Name { root: …, <bare component list> }`. Separating entities from directives reads more clearly and removes the parse ambiguity a bare list creates.

From @johanpel's review of #261.

**BREAKING:** all `model!` call sites migrate. Updated the in-repo sites (query-engine, readme, simulator) + the compile-fail tests. The quent-open stack (#261–#266) will rebase onto this; #261 then adds `analyzer:` on the new syntax.

### DIFF
--- a/crates/model-macros/src/lib.rs
+++ b/crates/model-macros/src/lib.rs
@@ -63,14 +63,14 @@ pub fn derive_resizable_resource(input: TokenStream) -> TokenStream {
 
 /// Composes a model's entities into a model type and event enum.
 ///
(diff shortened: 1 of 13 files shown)


johanpel (approved):
Looks good, as we add more fields making this explicit is much clearer IMO!

PR #270, opened 2026-06-29
feat(yaml): add minimal YAML event-model source lowering to a Schema

johanpel:
# Description

Adds `quent-yaml`: a YAML source format that parses into a `quent_schema::Schema`.
Field types are structural (built-ins, records, `list`/`option`, a plain `ref`);
constraints/metadata pass through opaquely; lowering collects all diagnostics in
one run. Purely YAML → Schema — no codegen-aware checks. Public API:
`parse_from_str` / `parse_from_file` → `Parsed { schema, warnings }`. The
instrumentation-build example now sources its model from `model.yaml`.

## Related Issues

Part of #355

This is a simple first commit to scaffold the `quent-yaml` crate.
Subsequent PRs will introduce features that make it get to feature parity with quent-model and quent-model-macros. Once we reach that those will be removed.


crates/instrumentation-build/example/build.rs
```diff
@@ -1,32 +1,37 @@
...
+    println!("cargo:rerun-if-changed={}", model.display());
+
+    // YAML source -> quent_schema::Schema. Errors carry file:line:column.
+    let parsed = match quent_yaml::parse_from_file(&model) {
+        Ok(parsed) => parsed,
+        Err(e) => {
+            eprintln!("{e}");
+            std::process::exit(1);
+        }
+    };
```
mbrobbel: This returns `Result` so:
```suggestion
    let parsed = quent_yaml::parse_from_file(&model)?;
```

crates/yaml/src/bin/check.rs
```diff
@@ -0,0 +1,109 @@
...
+        failed |= !check(file, warnings);
+    }
+    if failed {
+        ExitCode::FAILURE
+    } else {
+        ExitCode::SUCCESS
+    }
+}
+
+fn check(file: &str, warnings: bool) -> bool {
```
mbrobbel: Consider returning a `Result`?

crates/yaml/src/bin/check.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+/// Render one diagnostic, inserting a source line and caret under the column
+/// when it is located, so the caret stays directly under the message.
+///
+/// Builds on the diagnostic's own `Display`: the caret block is spliced in
+/// after the header line, before any `help:` line, rather than re-formatting
+/// the header and help here.
+fn render(src: &str, diagnostic: &Diagnostic) -> String {
```
mbrobbel: Maybe introduce a type that wraps this and then provide this via a `Display` impl?

crates/yaml/src/ast.rs
```diff
@@ -0,0 +1,185 @@
...
+/// A payload is a plain scalar. Anything structured, like a list or a nested
+/// mapping, is rejected while reading.
+pub(crate) type AnnotationMap = IndexMap<String, Option<String>>;
+
+/// A whole model file.
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(crate) struct Model {
+    /// The format version. Only `1` is supported.
+    pub(crate) quent: u32,
```
mbrobbel: Is this the version of this yaml format?

crates/yaml/src/diag.rs
```diff
@@ -0,0 +1,141 @@
...
+    /// A dotted semantic path, e.g. `entities.Engine.events.started`, from
+    /// lowering.
+    Path(String),
+    /// The model as a whole, with no finer location.
+    Whole,
+}
+
+/// A single problem in a model source.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Diagnostic {
```
mbrobbel, line 24: For follow-up: there are some nice crates for this on crates.io.

crates/yaml/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+    Io(#[from] std::io::Error),
+    #[error("{0}")]
+    Invalid(Diagnostics),
+}
+
+/// Parse and lower model source text into a validated [`Schema`].
+///
+/// Diagnostics name the source `source`, or leave it unnamed when `None`. See
+/// [`parse_from_file`] to read and name a file.
+pub fn parse_from_str(src: impl AsRef<str>, source: Option<&str>) -> Result<Parsed, Error> {
```
mbrobbel, line 44: Nit: static function of `Parsed`?

crates/yaml/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+                        .to_string(),
+                ),
+            )
+        })
+        .collect();
+    Ok(Parsed { schema, warnings })
+}
+
+/// Read a model file and parse it via [`parse_from_str`], naming it by path.
+pub fn parse_from_file(path: impl AsRef<Path>) -> Result<Parsed, Error> {
```
mbrobbel, line 105: Same here. But this is just style preference.

mbrobbel (approved):

crates/yaml/src/ast.rs
johanpel: Yes, but now that I'm revisiting this perhaps we should not set it to 1 but something like "alpha" so it is implied that it's still unstable for now and may break relatively often.

crates/instrumentation-build/example/build.rs
johanpel: https://github.com/rapidsai/quent/pull/270/commits/5a0b096c5f937b8f9c1817c2f54a63a45da08b39

crates/yaml/src/bin/check.rs
johanpel: https://github.com/rapidsai/quent/pull/270/commits/8cc656a71de550677da2435510a5686473fb30f6

crates/yaml/src/bin/check.rs
johanpel: https://github.com/rapidsai/quent/pull/270/commits/931941fc1619274514fad32bb01043fff14e06fa

crates/yaml/src/ast.rs
johanpel: https://github.com/rapidsai/quent/pull/270/commits/dbf4d4963c499d84a6ce0e32e5a3fc5250aea281

Will follow up with the same for the constraint crates for now as they all plug .v1

crates/yaml/src/diag.rs
johanpel: Something like: https://github.com/rust-lang/annotate-snippets-rs ?

mbrobbel: Or: 
- https://crates.io/crates/miette
- https://crates.io/crates/ariadne
- https://crates.io/crates/codespan-reporting

PR #277, opened 2026-06-30
feat(open)!: detect the artifact format inside the viewer importer

mbrobbel:
Follow-up to #264 review (johanpel): let the analyzer detect the artifact's serialization format instead of threading it through `quent-open`.

`QuentViewer::import_events` and the `model!`-generated `import_events` now take just the context directory and detect the format from the on-disk streams (new `quent_exporter::FileSystemFormat::detect`). The server's `index_query_engines` detects per-context too. The generated viewer wrapper enables all exporter formats so any artifact opens, and `quent-open` drops `Format`/`detect_format` and the format component of its cache key.

**Breaking:** `QuentViewer::import_events` / `Model::import_events` drop the `format` argument; `index_query_engines` drops its `format` parameter. Downstream `QuentViewer` impls (e.g. sirius) drop the `format` param and call `Model::import_events(dir)`.

**Compatibility:** `quent-open` builds the wrapper against the quent/analyzer commits pinned in `model.qmi` and now emits the one-argument calls, so it can only open artifacts whose pinned commits already include this change. Artifacts pinned to earlier commits (from the pre-release window) won't build a viewer — an accepted consequence of this breaking change; versioning the sidecar/codegen against the pinned API is left as future work.

Off `main`, independent of the merged quent-open work (#265) and the open stack (#266/#273/#276). Whichever of this PR and that stack lands second needs a rebase — both touch `quent-open`.

### DIFF
--- a/crates/open/src/spec.rs
+++ b/crates/open/src/spec.rs
@@ -1,11 +1,12 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Build a [`ViewerSpec`] from a context's `model.qmi`: pinned git sources,
-//! analyzer package, and artifact format for generating/building a viewer.
+//! Build a [`ViewerSpec`] from a context's `model.qmi`: the pinned git sources
+//! and analyzer package needed to generate/build a viewer. (The serialization
+//! format is detected at import time by the analyzer, not here.)
 
 use std::hash::{Hash, Hasher};
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
 use quent_build_info::{ArtifactInfo, BuildInfo, SIDECAR_FILE_NAME};
 use walkdir::WalkDir;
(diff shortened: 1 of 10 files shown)


crates/open/src/spec.rs
```diff
@@ -1,11 +1,12 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Build a [`ViewerSpec`] from a context's `model.qmi`: pinned git sources,
-//! analyzer package, and artifact format for generating/building a viewer.
+//! Build a [`ViewerSpec`] from a context's `model.qmi`: the pinned git sources
+//! and analyzer package needed to generate/build a viewer. (The serialization
+//! format is detected at import time by the analyzer, not here.)
```
johanpel:
```suggestion
//! Build a [`ViewerSpec`] from a context's `model.qmi`: the pinned git sources
//! and analyzer package needed to generate/build a viewer.
```

johanpel (approved):
Nice cleanup, thanks.

Issue #49, opened 2026-03-10
Higher-level plan operator filtering


johanpel:
If you're at some higher-level plan and select an operator, it should be possible to highlight all activity on the timeline related to any child operators residing in lower plans.

Right now the analyzer is only capable of doing this for lowest-level plans.

PR #289, opened 2026-07-01
feat(query-engine): widen OperatorFilter to accept multiple operator IDs

johanpel:
# Description

operator_id: Option<Uuid> → operator_ids: Vec<Uuid> (empty = no filter), so a timeline request can scope to a set of operators instead of one. Updates simulator analyzer filtering, cache tests, and UI call sites accordingly. No behavior change for single-operator callers.

Written by Claude Sonnet 5.

## Related Issues

#49


johallar (approved):
🙇

PR #295, opened 2026-07-02
feat(ui): auto-expand info panel when DAG node is selected

### LINKED ISSUE #281: UI: Operator node drill-down detail panel
## Summary

Clicking an Operator node in the plan DAG currently shows no detail. We need an inline detail panel that opens on click, showing:

- `instance_name` and `type_name`
- `custom_attributes` (key-value pairs, e.g. join strategy, node_id, partition count)
- Per-port statistics (rows in/out, bytes) when available
- Parent operators (logical → physical linkage)

## Context

Reported during review of rapidsai/cudf#22337 (cudf-polars Quent integration). The plan DAG structure is correct and nodes are declared, but clicking a node reveals nothing useful to the user.

The design intent is progressive disclosure: the plan DAG shows the shape of the query, and clicking a node reveals engine-specific detail without leaving the plan view. This is analogous to how Polars Cloud's query profiler shows per-operator stats inline.

## Design Notes

- Panel should appear inline (sidebar or overlay), not navigate away from the plan view
- Should handle the case where custom_attributes is empty gracefully
- Per-port stats (rows, bytes) are not yet emitted by cudf-polars but the panel should be designed to accommodate them when they arrive

cmatzenbach:
# Description

In order to make the info panel more discover-able, auto-expand the info panel whenever a user selects a DAG node. If a user clicks outside of the node into blank space on the DAG the panel will automatically collapse (so users can do that instead of having to close the panel via the interaction button there).

## Related Issues

Closes #281 

## Testing

1. Click DAG node, panel should auto-expand
2. Click on DAG outside of a node, panel should collapse
3. All other interactions remain the same

## Screenshots

https://github.com/user-attachments/assets/29f6bfab-74f6-4e77-a1c6-36784f97c8c9

### DIFF
--- a/ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
+++ b/ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
@@ -13,10 +13,8 @@ export const DAGNodeInfoPanel = () => {
   const [isExpanded, setIsExpanded] = useState(false);
 
   useEffect(() => {
-    if (!selectedNodeData) {
-      setIsExpanded(false);
-    }
-  }, [selectedNodeData]);
+    setIsExpanded(!!selectedNodeData);
+  }, [selectedNodeData?.nodeId]);
 
   return (
     <div className="border-t bg-card flex-shrink-0">


johallar (approved):
👍

PR #306, opened 2026-07-02
feat(ui): layout direction control

### LINKED ISSUE #280: UI: Flip DAG layout so source nodes (e.g. Scan) appear at the bottom
## Summary

The query plan DAG in the Quent UI currently renders top-to-bottom (root at top, leaf/source nodes like Scan at the bottom of visual space but arrows pointing down). For query profiling, the conventional and intuitive direction is bottom-to-top: data sources (Scan, DataFrameScan) appear at the bottom and data flows upward toward the result.

## Context

Reported during review of rapidsai/cudf#22337 (cudf-polars Quent integration).

The edge direction in the trace is already correct: edges go `child.out → parent.in` (i.e. source to sink), so Scan is the source. The issue is in `ui/packages/@quent/components/src/dag/layout.ts` which uses `d3-dag`'s `sugiyama` layout, which is top-to-bottom by default.

## Suggested Fix

In `layout.ts`, either:
- Configure `sugiyama` for bottom-to-top layout, or
- Reverse the edge direction before passing to d3-dag and re-reverse node positions after

## References
- `ui/packages/@quent/components/src/dag/layout.ts`

cmatzenbach:
# Description

Changes the default layout to be from bottom-to-top, as opposed to top-to-bottom based on comment in linked issue. Adds a dropdown under the "Settings" tab that allows users to control the layout direction of the DAG and switch back if desired.

## Related Issues

Resolves: #280

## Testing

1. Ensure query plans now flow from bottom to top by default
2. Go into "Settings" and switch plan direction, ensuring it renders accordingly
3. Ensure everything looks normal and other settings function as expected in both direction modes

## Screenshots

Screenshot_20260706_124344

Screenshot_20260706_124415

### DIFF
--- a/ui/packages/@quent/components/src/dag/DAGChart.tsx
+++ b/ui/packages/@quent/components/src/dag/DAGChart.tsx
@@ -38,6 +38,7 @@ import {
   useEffectiveHighlightedNodeIds,
   useSetSelectedNodeData,
   useSetDagDisplayedNodeIds,
+  useSelectedDagLayoutDirection,
 } from '@quent/hooks';
 import { calculateLayout, NODE_LAYOUT_WIDTH } from './layout';
 import type { DAGData } from '../services/query-plan/types';
@@ -276,6 +277,7 @@ const FlowLayout = ({
   const setDagDisplayedNodeIds = useSetDagDisplayedNodeIds();
   const setSelectedNodeData = useSetSelectedNodeData();
   const selectedNodeIds = useSelectedNodeIds();
+  const [layoutDirection] = useSelectedDagLayoutDirection();
   const hasUserInteracted = useRef(false);
 
   // Sync controlled selectedNodeIds into the atom when provided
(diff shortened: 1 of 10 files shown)


johallar (approved):
Functionality is great!

One design nit, the alignment irks my eyeballs a little bit:
Image

Maybe we can mess with having labels on top of the selects now that we're not as vertically challenged? Or open to other layout options. Feel free to merge and we can work on the design of the settings pane as we add more.

cmatzenbach:
> Functionality is great!
> 
> One design nit, the alignment irks my eyeballs a little bit: Image
> 
> Maybe we can mess with having labels on top of the selects now that we're not as vertically challenged? Or open to other layout options. Feel free to merge and we can work on the design of the settings pane as we add more.

Yeah I agree, now that we have the space we can improve this. Let's revisit this in a follow-up.

PR #314, opened 2026-07-06
feat(ui): zoom + pan UI test

johallar:
# Description
Adds a simple zoom/pan timeline e2e test which fails with echarts 6 upgrade, will use to validate


johallar:
Fixed by https://github.com/rapidsai/quent/pull/447

Issue #250, opened 2026-06-24
Make serde opt-in across instrumentation stack


johanpel:
#249 will place serde derives on generated event types by default because the current instrumentation run-time support structures require it, because all our current exporters require it. Because other types of exporters could exist, this should become opt-in rather than the default.

PR #321, opened 2026-07-07
feat(instrumentation)!: remove serialize bound from generic instrumentation path

johanpel:
# Description

Lifts the `Serialize`/`Deserialize` bounds off the generic `Context`, the
generated `{App}Context`, and event types so they apply only when an exporter
actually serializes. A non-serializing exporter (e.g. callback) now needs no
`Serialize` on its event types.

### What changed
- **Provider traits.** New `ExporterProvider<T>` (async `create_exporter`) and
  `ImporterProvider<T>` (sync `create_importer`). `ResolvedExporterOptions`
  implements `ExporterProvider` directly, split into a serde impl
  (`T: Serialize`, for filesystem/collector builds) and a bound-free
  callback-only impl.
  _Why: the old shared construction path (a generic `Context`/free
  `create_exporter`) forced `Serialize` onto every event type. A provider trait
  scopes the bound to the impl that actually serializes, so the callback impl
  can carry none._

- **Pure-host `Context`.** The runtime host no longer builds exporters; the
  generated context builds each entity's exporter via a provider and hosts it
  through `Context::observer`. `try_active`→`try_new`, `observer_with`→`observer`.
  _Why: while `Context` built exporters it had to name `Serialize`, propagating
  the bound to the generated `{App}Context`. Removing construction from it is
  what drops the bound from the generic and generated contexts._

- **`serde` is now optional.** Serde-using code is feature-gated per backend;
  callback-only builds compile with serde absent.
  _Why: opt-in means a build that serializes nothing must not compile — or
  depend on — serde at all._

- **Per-backend `Options` restructure.** The filesystem exporter/importer move
  into a `filesystem` submodule; each backend owns an `Options` type with its
  own `resolve`/`create_exporter`/`create_importer`.
  _Why: co-locating each backend's provider impl with its options is what lets
  the `Serialize` bound (and the serde feature gate) sit on only the
  serializing backends, not a shared enum-wide impl._

- **Crate rename `quent-exporter*` → `quent-io*`.**
  _Organizational, not required for the serde change: the crate now covers both
  export and import, so the name reflects that._

- **Instrumentation module split** into `context.rs`, `observer.rs`,
  `sidecar.rs`.
  _Organizational, not required for the serde change: `lib.rs` shrank once
  `Context` became pure-host, so the pieces were separated._

### Notes
- Feature unification caveat: a non-`Serialize` model only compiles in a build
  with serde off everywhere.

## Related Issues

Solves #250


johanpel:
closing this to split it up a bit more: https://github.com/rapidsai/quent/pull/324

PR #323, opened 2026-07-08
feat(analyzer): surface per-transition state attributes and show task stats in timeline tooltips

felipeblazing:
## What

Typed FSM state payloads (e.g. `input_bytes` on a task's `Computing` state) survived analyzer ingestion as `TransitionEvent.data: T` but were dropped at every boundary on the way to the UI — `TransitionEvent::attributes()` returned `empty()` and `quent_ui::FsmTransition` had no attributes field. This PR surfaces them end-to-end and renders them on hover in the timeline.

### Backend
- New `ExtractAttributes` (value-level, one `Attribute` per declared field) and `ToAttributeValue` (field → `Value` conversion) traits in `quent_model::analyze`.
- `state!` generates extraction for inline and external attribute modes (inline `instance_name` is excluded — it's surfaced via `ExtractInstanceName`); `#[derive(Attributes)]` generates it for attribute structs so nested structs and `Vec<Struct>` fields compose (`Vec` of a local struct is handled syntactically in the macro since the orphan rule forbids the impl downstream); `fsm!` dispatches via a new `TransitionInfo::attributes()` (defaulted to empty so resource FSMs are untouched).
- **Lazy conversion** (per review): nothing is materialized at ingestion. `TransitionEvent` keeps only the typed data; the `Transition` trait's `attributes()` returns an owned `Vec<Attribute>` computed on demand, so the typed→dynamic conversion runs only where a UI payload is built (`try_to_ui_fsm`, after long-entity filtering) — FSM instances that are filtered out never pay for it, in cycles or memory.
- `quent_ui::FsmTransition` gains two fields (per review): `attributes` — recorded by the application's instrumentation — and `derived_attributes` — computed by its analyzer (e.g. a per-span rate) — so users can tell where a value came from. ts-bindings regenerated.

### Contract: analyzers synthesize, the UI renders verbatim
The UI contains **no application-specific semantics**. Each application's analyzer converts recorded attributes lazily and appends whatever derived values it wants shown (a processing rate, a resolved operator/pipeline chain, …) to `derived_attributes` in its UI converter. The UI keeps only domain-agnostic behavior:
- attribute rows with name-based unit formatting (`*_bytes` → IEC bytes; `bytes_per_sec` keys → SI `B/s`, checked first since they also match the bytes heuristic),
- derived attributes rendered under a muted "derived" subsection,
- a span `duration` row (any span has one),
- wrapped rendering for long string values so operator chains stay readable,
- defensive unwrapping of `Value`'s externally-tagged serde encoding (`{"U64": n}`), and tolerance of responses from servers built before this change.

The simulator analyzer demonstrates the pattern in `try_to_ui_fsm`: recorded attributes via the lazy `transition.attributes()`, plus a derived `bytes_per_sec` from `Computing.input_bytes` over the span; the simulator task's `Computing` state gains an `input_bytes` attribute as the in-repo end-to-end vehicle.

## Testing
- `cargo test --workspace` (109 test binaries) and `cargo clippy --workspace --all-targets` clean; unit tests for inline extraction and `TransitionInfo` dispatch.
- UI: `pnpm typecheck` clean; 472 vitest tests pass including coverage for recorded/derived propagation onto marks, tooltip rendering (`input_bytes → 1.40 GiB`, derived `bytes_per_sec → 2.00 GB/s`, wrapped pipeline chain), and the `Value` unwrap/format helpers.
- Verified live against Sirius TPC-H SF300 2-GPU telemetry: hovering a computing span shows recorded `[current_operator_id, input_bytes]` and derived `[bytes_per_sec, pipeline]`. Existing datasets work unchanged — only the read path changed.

Note for downstream consumers: `FsmTransition.{attributes,derived_attributes}` are required fields, so hand-rolled converters (e.g. Sirius's `try_to_ui_fsm`) need a small update; companion PR: sirius-db/sirius#1112.


crates/ui/src/lib.rs
```diff
@@ -205,6 +206,10 @@ pub struct FsmTransition {
     pub usages: Vec<FsmUsage>,
     /// The timestamp in seconds relative to an epoch.
     pub timestamp: TimeSec,
+    /// Attribute key-value pairs carried by this state. Applications may
+    /// also synthesize derived attributes here (e.g. a per-span processing
+    /// rate) — the UI renders whatever arrives.
+    pub attributes: Vec<Attribute>,
```
johanpel: Could we split this out over two fields such that in the UI the distinction between "event attributes" and "derived attributes" can be made?

I think this is useful because this helps application devs to understand where certain values are computed (either in their application's instrumentation code, or in the analyzer, respectively).

crates/analyzer/src/fsm/events.rs
```diff
@@ -134,6 +136,7 @@ impl<T: TransitionInfo> FsmEventsBuilder<T> {
             timestamp: event.timestamp,
             state_name,
             usages,
+            attributes: state.attributes(),
```
johanpel: IIUC we're eagerly performing the conversion of all statically typed attributes to dynamically types ones here for **every** FSM instance, including the types that are bulky and typically heavily filtered before conversion to something UI-compatible (e.g. in Sirius, task, data batch, etc.).

This takes a lot of cycles and expands the memory footprint of loaded data significantly, while most results remain unused.

I wonder if it's possible to do this conversion more lazily, only in the analyzer, as a part of a conversion to an `quent_ui::FsmTransition` only.

(We should probably do the same for usages, not in this PR ofcourse)

examples/simulator/analyzer/src/task.rs
```diff
@@ -47,10 +47,27 @@ impl TaskExt for Task {
...
     fn try_to_ui_fsm(&self, epoch: TimeUnixNanoSec) -> AnalyzerResult<FiniteStateMachine> {
-        let transitions = self
-            .transitions()
+        let raw = self.transitions();
+        let transitions = raw
             .iter()
-            .map(|t| {
+            .enumerate()
+            .map(|(i, t)| {
+                let mut attributes = t.attributes.clone();
```
johanpel: This is the place where statically typed -> dynamically typed attribute conversion should happen only

johanpel (commented):
Nice work so far, some comments:

crates/ui/src/lib.rs
felipeblazing: Why is that an important distinction to make. If the trait is computed in the application, input_bytes for example vs derived like Rate (bytes /  seconds?)

johanpel: Because in my experience, when developers see unexpected values in profiling tools, they rightfully apply a dose of healthy skepticism that motivates them to figure out exactly how the value is determined. This often leads them to inspect the code determining that value, so we can help them a little by making that distinction.

crates/analyzer/src/fsm/mod.rs
```diff
@@ -20,9 +20,17 @@ pub mod runtime;
...
-    /// this transition.
-    fn attributes(&self) -> impl Iterator<Item = &Attribute>;
+    /// Return the arbitrary key-value attributes associated with this
+    /// transition.
+    ///
+    /// May be computed on demand — implementations backed by statically
+    /// typed state data convert it lazily so that FSM instances that are
+    /// filtered out before display never pay for the conversion. Call this
+    /// only when the attributes will actually be used (e.g. when building a
+    /// UI payload).
```
johanpel: Nit: let's only describe the contract, not the implementation
```suggestion
```

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -190,6 +190,60 @@ export function isBytesStat(name: string): boolean {
...
+
+/**
+ * Format an attribute value for display, inferring units from the key name.
+ * The bytes-rate check precedes the bytes check — `bytes_per_sec` matches both.
+ */
+export function formatAttributeValue(key: string, value: unknown): string {
+  const v = unwrapAttributeValue(value);
+  if (v == null) return '—';
+  if (typeof v === 'number' || typeof v === 'bigint') {
+    const n = Number(v);
```
johallar: This won't work if `v` > `Number.MAX_SAFE_INTEGER`. This is currently a shortcoming of the UI that we need to address, putting in an issue now. eg. the DAG detail panel + formatter utils assume everything is a `Number`. This will be kind of a broad change so don't need to fix here

johallar: https://github.com/rapidsai/quent/issues/351

ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
```diff
@@ -160,25 +179,69 @@ function buildBarSegments(
   return { segments, overlayPct };
 }
 
-function ActiveMarksSection({
-  marks,
-}: {
-  marks: { label: string; stateName: string; color: string }[];
-}) {
+/** Values longer than this wrap onto their own line. */
+const INLINE_VALUE_MAX_CHARS = 32;
```
johallar, line 183: I have other ideas to table-ify this as we add more and more information, but I can follow up when [refactoring the entities view](https://github.com/rapidsai/quent/issues/214), better to have this be out there + useful in the short term.

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -190,6 +190,60 @@ export function isBytesStat(name: string): boolean {
...
+  'Struct',
+  'List',
+]);
+
+/**
+ * Unwrap an attribute `Value` to a plain JS value. The Rust `Value` enum
+ * serializes externally tagged (`{"U64": 5}`) while the generated TS type is
+ * untagged — handle both shapes.
+ */
+export function unwrapAttributeValue(value: unknown): unknown {
```
johallar: There is some overlap between this and `unwrapTaggedValue`, and I believe `unwrapTaggedValue` handles all possible `Value` types, but best to consolidate here since this package can be imported anywhere. cc: @cmatzenbach

crates/analyzer/src/fsm/mod.rs
```diff
@@ -20,9 +20,13 @@ pub mod runtime;
 pub trait Transition: Timestamp {
     /// Return the unique name of the state this transition leads to.
     fn name(&self) -> &str;
-    /// Return an iterator over arbitrary key-value attributes associated with
-    /// this transition.
-    fn attributes(&self) -> impl Iterator<Item = &Attribute>;
+    /// Return the key-value attributes of this transition.
```
johanpel: Nit:
```suggestion
    /// Return the key-value attributes of this transition converted to dynamically-typed ones.
```

domains/query_engine/tests/fixed/src/lib.rs
```diff
@@ -524,6 +524,8 @@ fn execute_tasks(ctx: &SimulatorContext) {
         ts!(
             t_c,
             task.computing(
+                "",
```
johanpel: I'm confused by the appearance of the `&str` argument here, what is it for, and where did this come from? (This might be some pre-existing problem surfacing)

johanpel (approved):
Rust changes look good to me, thanks @felipeblazing

johallar (approved):

johallar:
Thanks for the changes @felipeblazing!

PR #324, opened 2026-07-08
refactor(instrumentation)!: relocate serde bounds behind exporter providers

johanpel:
# Description

Lifts the `Serialize` bound off the generic `Context`, the generated `{App}Context`, and event types so it applies only when an exporter actually serializes. A callback exporter now needs no `Serialize` on its event types.

Changes:

- Adds **`ExporterProvider<T>`** trait, implemented by:
  - `ResolvedExporterOptions`, a serde impl (`T: Serialize`) for filesystem/collector
  - A serde bound-free impl for the callback exporter. 
  - Scopes the bound to the impl that serializes.
  - This can later be leveraged to support e.g. Narrow-based exporters.
- **Pure-host `Context`.** 
  - `Context::observer` no longer builds the exporter or requires `T: Serialize`. 
  - It takes a pre-built `Box<dyn Exporter<T>>`.
  - `try_new` drops `ModelInfo`/`ExporterOptions` and just adopts an `id`. 
  - Exporter construction (via a provider) and the sidecar write move to the generated context.
  - This is what sheds `Serialize` from the generic and generated contexts.
- **`serde` is optional**, gated per backend; callback-only builds compile without it.
- **Instrumentation split** into `context.rs`/`observer.rs`/`sidecar.rs` (organizational).

## Example of changes to generated context

```rust
impl AppContext {
    pub fn try_new(
        exporter: Option<quent_model::exporter::ExporterOptions>,
    ) -> Result<Self, Box<dyn std::error::Error>> { /* -> try_with_id */ }

    pub fn try_with_id(
        id: quent_model::uuid::Uuid,
        exporter: Option<quent_model::exporter::ExporterOptions>,
    ) -> Result<Self, Box<dyn std::error::Error>> { /* resolve + write_sidecar -> build(id, resolved) */ }

    // non-generic: ResolvedExporterOptions is itself the provider
    fn build(
        id: quent_model::uuid::Uuid,
        resolved: quent_model::exporter::ResolvedExporterOptions,
        let exporter = <quent_model::exporter::ResolvedExporterOptions
            as quent_model::exporter::ExporterProvider<TaskEvent>>::create_exporter(&resolved).await?;
        // ... other EntityEvent types
    }
}
```

## Related issues

Next for #250, one final cosmetic/cleanup incoming after this


crates/instrumentation/src/observer.rs
```diff
@@ -0,0 +1,217 @@
...
+use uuid::Uuid;
+
+/// Wrapper around an optional channel sender.
+///
+/// When the inner sender is `None` (i.e. the noop exporter is selected), `send`
+/// is a no-op that avoids any channel or event-forwarding overhead.
+pub struct EventSender<T> {
+    tx: Option<UnboundedSender<Event<T>>>,
+    /// Flag shared across clones to prevent potentially massive log spam from
+    /// subseQUENT sender errors after the first.
```
mbrobbel, line 28: 👏

crates/instrumentation/src/observer.rs
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+}
+
+impl<T> EventSender<T> {
+    /// Returns a noop sender that silently drops all events.
+    pub fn noop() -> Self {
+        Self {
+            tx: None,
+            disable_error_log: Arc::new(AtomicBool::new(false)),
```
mbrobbel: It doesn't matter but I guess:
```suggestion
            disable_error_log: Arc::new(AtomicBool::new(true)),
```

crates/instrumentation/src/observer.rs
```diff
@@ -0,0 +1,217 @@
...
+///
+/// Generated code constructs and shares this type. Instrumented application
+/// code uses the generated observer and its per-instance entity handles
+/// instead. Those manage the shared ownership and flush-on-last-drop this type
+/// relies on, so holding or dropping it directly can lose or prematurely flush
+/// events.
+#[doc(hidden)]
+pub struct Observer<T>
+where
+    T: Send + EntityEvent + 'static,
```
mbrobbel: Just curious why you need `T: Send` here?

mbrobbel (approved):

crates/instrumentation/src/observer.rs
johanpel: Thanks, indeed we don't need any bounds here since we factored out the things that do need this into `spawn_forwarder`: https://github.com/rapidsai/quent/pull/324/commits/ba134a7a19d1360838acf404b16cd37c33ad1c37

crates/instrumentation/src/observer.rs
johanpel: https://github.com/rapidsai/quent/pull/324/commits/ba134a7a19d1360838acf404b16cd37c33ad1c37

PR #341, opened 2026-07-08
refactor(instrumentation)!: Rename `quent-exporter` → `quent-io` and unify exporter options

johanpel:
# Description

Follow-up to #324. Two commits: a mechanical crate rename/restructure, then a small design change to the options type. Reviewable commit-by-commit.

### 1. Rename + restructure (`26ed640`)

Renames the `quent-exporter*` crate family to `quent-io*` (the crate covers both export and import). Moves the filesystem exporter/importer into a `filesystem` submodule with per-backend `Options` types, and adds a synchronous `ImporterProvider` trait replacing the free `create_importer` function. All consumers (the `quent_model::io` re-export, the `model!` macro, the codegen bridges, domains, examples) are updated to the new names and module paths. No behavior change.

### 2. Single `ExporterOptions` (`ba28f8f`)

Collapses `ExporterOptions` and the near-identical `ResolvedExporterOptions` into one `ExporterOptions`. The context id is now data rather than a type-state: each backend's `Options` carries a `context_id` (private field, defaulted nil via a `new(..)` constructor), and `resolve(id) -> Resolved` becomes a consuming `with_context_id(id) -> Self`. The filesystem and collector exporters reject a nil context id at construction, and the filesystem output directory (`root/<context_id>`) is computed lazily.

## Related Issues

Final PR in a series due to #250.

This is mostly an API cleanup for naming and importer/exporter consistency.


crates/exporter/src/lib.rs
johanpel, line 1: Meat of the change is here, the amount of config guards here is somewhat confusing.

crates/io/src/lib.rs
johanpel, line 1: `crates/exporter/src/lib.rs` is now split out across some modules that are feature gated in their entirety instead here:

crates/io/src/lib.rs
```diff
@@ -0,0 +1,140 @@
...
+/// filled in by [`Self::with_context_id`]; their exporters reject a nil id.
+#[derive(Debug, Clone)]
+pub enum ExporterOptions {
+    #[cfg(filesystem)]
+    FileSystem(FileSystemExporterOptions),
+    #[cfg(feature = "collector")]
+    Collector(CollectorExporterOptions),
+    #[cfg(feature = "callback")]
+    Callback(EventCallback),
+}
```
johanpel, line 51: The second change is that there is now just one variant of this, but its constituents that need a context id now carry it in their fields, and then their respective exporters do whatever they want with it instead of awkwardly trying to unify the behavior at this level.

crates/model-macros/src/model_macro.rs
```diff
@@ -311,17 +311,17 @@ pub fn expand(input: TokenStream) -> syn::Result<TokenStream> {
...
             pub fn try_with_id(
                 id: quent_model::uuid::Uuid,
-                exporter: Option<quent_model::exporter::ExporterOptions>,
+                exporter: Option<quent_model::io::ExporterOptions>,
             ) -> Result<Self, Box<dyn std::error::Error>> {
                 match exporter {
                     None => Ok(Self::noop(id)),
                     Some(options) => {
-                        let resolved = options.resolve(id);
+                        let options = options.with_context_id(id);
```
johanpel: This still happens, but now exporter providers themselves will reject nil context ids where applicable.

crates/io/collector/src/lib.rs
```diff
@@ -5,19 +5,56 @@
...
+where
+    T: Send + EntityEvent + 'static,
+    T: serde::Serialize,
+{
+    async fn create_exporter(&self) -> ExporterResult<Box<dyn Exporter<T>>> {
+        if self.context_id.is_nil() {
+            return Err(ExporterError::Other(
+                "collector exporter requires a context id; call `with_context_id` first".into(),
+            ));
+        }
```
johanpel: e.g. here

crates/io/src/filesystem/exporter.rs
```diff
@@ -0,0 +1,82 @@
...
+where
+    T: Send + EntityEvent + 'static,
+    T: Serialize,
+{
+    async fn create_exporter(&self) -> ExporterResult<Box<dyn Exporter<T>>> {
+        if self.context_id.is_nil() {
+            return Err(ExporterError::Other(
+                "filesystem exporter requires a context id; call `with_context_id` first".into(),
+            ));
+        }
```
johanpel: or here for filesystem exporters

johanpel (commented):
Guidance for non-cosmetic changes:

crates/io/types/src/lib.rs
```diff
@@ -29,15 +53,32 @@ impl From<std::io::Error> for ExporterError {
...
+/// Result type for importers.
 pub type ImporterResult<T> = std::result::Result<T, ImporterError>;
 
+/// A source of one entity's events.
+pub trait Importer<T>: Iterator<Item = Event<T>> {}
+
+/// Provides an importer instance for `T`.
+///
+/// Synchronous, unlike [`ExporterProvider`]: importer construction opens files
+/// with blocking I/O and does no async work.
```
mbrobbel: What if the importer is async?

crates/io/types/src/lib.rs
```diff
@@ -29,15 +53,32 @@ impl From<std::io::Error> for ExporterError {
...
+pub trait ImporterProvider<T>
+where
+    T: Send + EntityEvent + 'static,
```
mbrobbel: `T: Send`?

crates/io/types/src/lib.rs
```diff
@@ -6,6 +6,30 @@
...
+    ///
+    /// Calling [`Self::push`] will result in an error after calling this.
+    async fn shutdown(&mut self) -> ExporterResult<()>;
+}
+
+/// Provides an exporter instance for `T`.
+#[async_trait::async_trait]
+pub trait ExporterProvider<T>
+where
+    T: Send + EntityEvent + 'static,
```
mbrobbel: Why `+ 'static` here and not for `T` in `Exporter`?

crates/io/types/src/lib.rs
```diff
@@ -6,6 +6,30 @@
...
+where
+    T: Send + EntityEvent,
+{
+    /// Export one event.
+    async fn push(&mut self, event: Event<T>) -> ExporterResult<()>;
+
+    /// Make a best-effort to flush any buffered events, then release any internal resources.
+    ///
+    /// Calling [`Self::push`] will result in an error after calling this.
+    async fn shutdown(&mut self) -> ExporterResult<()>;
```
mbrobbel: Maybe consume `self` here?

crates/io/types/src/lib.rs
```diff
@@ -6,6 +6,30 @@
 use quent_events::{EntityEvent, Event};
 use thiserror::Error;
 
+/// A sink for one entity's event stream.
+#[async_trait::async_trait]
+pub trait Exporter<T>: Send
```
mbrobbel: Why `Send`?

crates/instrumentation/tests/collector_roundtrip.rs
```diff
@@ -84,15 +82,14 @@ fn collector_client_flushes_all_events_on_drop() {
...
     {
         let observer = ctx
             .block_on(async {
                 let exporter =
-                    <ResolvedExporterOptions as ExporterProvider<TestEvent>>::create_exporter(
-                        &resolved,
-                    )
-                    .await?;
+                    <ExporterOptions as ExporterProvider<TestEvent>>::create_exporter(&options)
+                        .await?;
```
mbrobbel: Staring at this; why not just:
```rust
let ctx = Context::try_new(id)?;
let observer = ctx.observer::<TestEvent>(CollectorExporterOptions::new(address)).await;
```

This avoids the error on `create_exporter` because you don't need `.with_context_id`, because you already have it in the context. Or am I missing something?

crates/io/types/src/lib.rs
johanpel: No importers are async today. Changing this would result in non-trivial changes to the analyzer crates so if we ever need this we can do it later.

crates/io/types/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/341/commits/54aa509d5357599cee7ae6bbdbe4a7a04f6c406b

crates/io/types/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/341/commits/54aa509d5357599cee7ae6bbdbe4a7a04f6c406b

crates/io/types/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/341/commits/54aa509d5357599cee7ae6bbdbe4a7a04f6c406b

crates/io/types/src/lib.rs
johanpel: Added a comment explaining this in: https://github.com/rapidsai/quent/pull/341/commits/54aa509d5357599cee7ae6bbdbe4a7a04f6c406b:

```
/// This is `Send` because generated application contexts first attempt to
/// create all exporters (asynchronously). If any errors occur, these are
/// immediately surfaced through the blocking context creation API. If no errors
/// occur, the exporters are then moved into their respective forwarder tasks.
```

Of course we could change this behavior in a follow-up if desired.

crates/instrumentation/tests/collector_roundtrip.rs
johanpel: I was going in circles with this a bit, because some exporters (actually just one) don't require a context id, and so I didn't want to include it as an argument of `ExporterProvider::create_exporter`.

But as you noticed this just shifts the complexity, as the burden of ensuring there is a context id for exporters that need it is placed on the code leveraging the trait to create exporters, where besides knowing the trait it also needs to reason about the internals of the exporter you're using, which is way less ideal.

Shifted this back again in: https://github.com/rapidsai/quent/pull/341/commits/68ec02871c9b7e9519203faf681c80809dbf4f52

mbrobbel (approved):

PR #345, opened 2026-07-08
chore(deps): bump react to match react-dom

johallar:
# Description
Fork of https://github.com/rapidsai/quent/pull/336, which also bumps react to match react-dom version

## Related Issues
https://github.com/rapidsai/quent/pull/336


johanpel (approved):

johallar:
Not sure the catalog approach is what we want yet, just taking a minute to read a bit more. Otherwise will merge and or open another PR with diff updates.

johallar:
We're going with the catalog approach to centralize versions within our workspace, and I will add to the coderabbit review instructions to check that if we bump catalog dependencies, we should also mirror that in peer dependencies for the individual packages

PR #346, opened 2026-07-09
test(open): gate quent-open backward compatibility in CI

mbrobbel:
# Description

Follow-up to #343: artifacts pin the quent revision they were built from, so `quent-open` must keep opening sidecars written by older quent revisions. #341 broke that and nothing flagged it before merge — this adds a regression gate so the next such break fails in the PR that introduces it, and the compatibility path ships in that same PR instead of after the fact.

**The test** (`opens_artifacts_built_by_a_previous_quent_commit`, in `crates/open/src/viewer.rs`): synthesizes the sidecar an older quent wrote for the in-repo simulator model, pinned to `QUENT_OPEN_COMPAT_COMMIT`, then runs the production path — sidecar → discovery → spec → wrapper generation → `cargo build`, including the legacy `quent-exporter` retry — and asserts the viewer binary is produced. It is `#[ignore]`d in normal test runs since it fetches git sources and compiles a full viewer.

**The CI job** (`open-compat` in `rust.yml`): runs the test with the previous commit — the PR's base commit, or the parent of the pushed commit on `main` — against `${{ github.server_url }}/${{ github.repository }}`, so it also works on forks. This catches both manifest-level breakage (crate/feature renames, like #341) and generated-`main.rs` API breakage, since the wrapper actually compiles against the pinned revision.

When this job fails on a PR, that PR breaks `quent open` for existing artifacts: add a compatibility path (like the `quent-io` → `quent-exporter` retry in `build_one`) in the same PR, rather than merging the breakage.

Verified locally end-to-end: the test builds a viewer for a sidecar pinning `6a7ffce` (current `main~1`) from the real GitHub remote and passes; it is skipped in a plain `cargo test`.

Versioning-based detection (recording a bumped quent version in sidecars so quent-open can select the wrapper shape deterministically, see the note in #343) is intentionally left for a follow-up; this gate is independent of how the compatibility path is implemented.

## Related Issues

Follow-up to #343 / #341.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -59,6 +59,42 @@ jobs:
       - run: pixi run cargo test --workspace --all-features --locked --all-targets
       - run: pixi run cargo build --workspace --all-features --locked --release
 
+  # Regression gate for `quent-open` backward compatibility: build a viewer for
+  # a sidecar pinning the previous quent commit (the PR's base commit, or the
+  # parent of the pushed commit), with the quent-open of this tree. A failure
+  # means this change breaks `quent open` for existing artifacts — add a
+  # compatibility path in the same PR instead of merging the breakage.
+  open-compat:
+    name: quent-open compatibility
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+          # Picking the previous commit on push builds needs the parent of HEAD.
+          fetch-depth: 2
+      - uses: prefix-dev/setup-pixi@5185adfbffb4bd703da3010310260805d89ebb11 # v0.9.6
+        with:
+          cache: true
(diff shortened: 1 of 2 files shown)


mbrobbel:
Validation that this gate catches breakage — a seeded break ([`7ea9b99`](https://github.com/mbrobbel/quent/commit/7ea9b994c91afdc9c6ef5ab8041b62f00c2b6848), since force-pushed away) made the generated wrapper `main.rs` call `index_query_engines_v2`, which no quent revision provides:

- **Seeded run**: every job stayed green — including the full Rust job with all unit tests — and only `quent-open compatibility` failed, in 2m17s, with the pinned revision's actual rustc error (``E0432: unresolved import `...::index_query_engines_v2` ``) in the test output: https://github.com/rapidsai/quent/actions/runs/29006229531/job/86078342287
- **Clean run**: on the reviewed commit the job passes in ~2m40s, building the viewer against the PR's base commit: https://github.com/rapidsai/quent/actions/runs/29005357460/job/86075540575

Also validated locally for the manifest-level class (a hypothetical second rename, `quent-io` → `quent-io2`): the test fails fast (0.18s) after the output shows the `quent-exporter` retry genuinely running and both names failing resolution — i.e. a repeat of #341 is flagged in the PR that introduces it.

johanpel (approved):

PR #348, opened 2026-07-09
ci: add CodeRabbit review configuration

johanpel:
# Description

Add a repo-root .coderabbit.yaml customizing CodeRabbit for quent's layout (Rust crates, TS/React ui, proto, docs, workflows). markdownlint is disabled since Markdown is already linted in CI via rumdl.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -3,9 +3,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 reviews:
-  profile: chill
-  high_level_summary: true
-  high_level_summary_in_walkthrough: true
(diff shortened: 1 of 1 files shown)


johanpel:
@coderabbitai review

johanpel:
@coderabbitai review

johanpel:
Seems like coderabbit won't work with its own modified config, probably for security reasons. In this PR it's probably just showing its default behavior which is rather noisy and should be surpressed by the proposed config. So we'll notice whether this minimal config is not too noisy after merge.

mbrobbel (approved):

PR #350, opened 2026-07-09
feat(ui): zoom timeline with mouse wheel and trackpad pinch

felipeblazing:
## Summary

The timeline view now zooms with plain mouse wheel, trackpad two-finger scroll, and trackpad pinch (delivered by browsers as ctrl+wheel), centered on the pointer's position on the time axis. Previously zooming required shift+wheel, and plain wheel was passed through to the page for scrolling.

- **Resource timeline rows** (`Timeline.tsx`): the inside dataZoom uses `zoomOnMouseWheel: true`, so all wheel/pinch input zooms; ECharts anchors the zoom at the pointer. The minimum-zoom-window guard now applies to all wheel input and suppresses the browser fallback (page scroll / pinch page-zoom) at the limit — ECharts would otherwise convert a blocked zoom into a pan.
- **Operator gantt** (`OperatorGanttChart.tsx`): plain wheel stays reserved for scrolling the operator rows (its only scroll mechanism); trackpad pinch (ctrl+wheel) zooms alongside the existing shift+wheel.

The zoom controller bar already zoomed on plain wheel; all charts stay in sync via the existing connect group and zoom atom.

Note the trade-off: with wheel bound to zoom, scrolling the page requires the pointer to be outside a timeline chart (e.g. over the tree columns).

## Test plan

- [x] `pnpm exec vitest run packages/@quent/components/src/lib/timeline.utils.test.ts` (48 passed)
- [x] eslint clean on both changed files
- [x] Manually verified against the simulator server (`quent-simulator-server` + `quent-simulator` + `pnpm dev`): wheel and pinch zoom center on the pointer, ruler/controller/rows stay in sync, zoom-in stops at the minimum window, gantt rows still scroll with plain wheel

### DIFF
--- a/ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
+++ b/ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
@@ -282,6 +282,17 @@ export function OperatorGanttChart({
           filterMode: 'none',
           xAxisIndex: [0],
         },
+        {
+          // Trackpad pinch arrives as ctrl+wheel; plain wheel stays reserved
+          // for scrolling the operator rows (see the wrapper wheel handler).
+          type: 'inside',
+          zoomOnMouseWheel: 'ctrl',
+          moveOnMouseMove: false,
+          moveOnMouseWheel: false,
+          throttle: 30,
+          filterMode: 'none',
+          xAxisIndex: [0],
+        },
       ],
     }),
     [gridOptions, startTimeMs, xAxisMax, yAxisCategories, customSeriesData, renderItem]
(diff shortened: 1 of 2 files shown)


johallar:
@felipeblazing I need to add a hint somewhere, but you can zoom on the individual resource timelines by holding the shift key currently. I much prefer this UX to having to move your mouse to a specific spot to scroll the resource tree

johallar:
Oh i see: 
> Previously zooming required shift+wheel, and plain wheel was passed through to the page for scrolling.

I do still much prefer a "hot key to zoom" on mouse wheel/trackpad/pinch, but I think we need to make it much more obvious that that's how you do it.

PR #353, opened 2026-07-10
perf(io): add event batch export via Exporter::drain_events

johanpel:
# Description

This PR adds `Exporter::drain_events` (default drains through single event pushes).

The observer forwarder task now drains the unbounded channel with `recv_many` and exports a batch per call instead of one call per event. It reuses one buffer which is reserved with `Exporter::batch_size_hint` before each `recv_many`.

For now only the msgpack/ndjson/postcard exporters override `drain_events` to export an event batch into one write buffer and then issue a single write. Callback and collector keep the default. I think the collector could leverage this too to shave off some overhead but that is a bit more work.

### DIFF
--- a/crates/io/msgpack/src/lib.rs
+++ b/crates/io/msgpack/src/lib.rs
@@ -14,7 +14,7 @@ use tokio::{
     fs::{File, OpenOptions},
     io::{AsyncWriteExt, BufWriter},
 };
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 use uuid::Uuid;
 
 /// File extension for MessagePack event files.
@@ -87,7 +87,7 @@ where
                     batch.extend_from_slice(&(payload.len() as u32).to_be_bytes());
                     batch.extend_from_slice(&payload);
                 }
-                Err(e) => error!("unable to serialize event: {e}"),
+                Err(e) => warn!("unable to serialize event: {e}"),
             }
         }
         writer.write_all(batch).await?;
--- a/crates/io/msgpack/src/lib.rs
+++ b/crates/io/msgpack/src/lib.rs
@@ -35,6 +35,8 @@ pub struct MsgpackExporterOptions {
 pub struct MsgpackExporter {
     /// `None` once [`shutdown`](Exporter::shutdown) has flushed and released it.
     writer: Option<BufWriter<File>>,
+    /// Framing buffer reused across [`drain_events`](Exporter::drain_events).
+    batch: Vec<u8>,
 }
 
 impl MsgpackExporter {
@@ -50,6 +52,7 @@ impl MsgpackExporter {
             .await?;
         Ok(Self {
             writer: Some(BufWriter::new(file)),
+            batch: Vec::new(),
         })
     }
 }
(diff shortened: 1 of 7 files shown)


johanpel:
Some benchmarks across varying event types, that I may add in a follow-up later:

image

crates/io/types/src/lib.rs
```diff
@@ -18,7 +19,28 @@ pub trait Exporter<T>: Send {
...
+    /// Exports every event in `events`, in order, leaving it empty.
+    ///
+    /// The default forwards each event to [`push`](Self::push). Override this
+    /// to amortize per-event overhead.
+    async fn drain_events(&mut self, events: &mut Vec<Event<T>>) -> ExporterResult<()>
+    where
+        T: Send + 'async_trait,
+    {
+        for event in events.drain(..) {
+            self.push(event).await?;
```
9prady9: Is breaking out of the batch on the first `push` error intentional here? Since `events.drain(..)` is dropped on the `?` early-return, every remaining event in the batch is discarded (up to `batch_size_hint`, 256). The old forwarder pushed events individually and logged a `warn` then continued, so a single failure only dropped that one event.

crates/io/msgpack/src/lib.rs
```diff
@@ -68,6 +71,29 @@ where
...
+        // write. A record that fails to serialize is logged and skipped so one
+        // bad event does not drop the batch.
+        batch.clear();
+        for event in events.drain(..) {
+            match rmp_serde::to_vec(&event) {
+                Ok(payload) => {
+                    batch.extend_from_slice(&(payload.len() as u32).to_be_bytes());
+                    batch.extend_from_slice(&payload);
+                }
+                Err(e) => error!("unable to serialize event: {e}"),
```
9prady9: Consistency nit: a serialize failure is logged at `error!` and the event skipped here, whereas `push` returns `Err` for the same failure (surfaced as a `warn` by the forwarder). Could we make the log level consistent across the batch and single-event paths? Same applies to the ndjson and postcard `drain_events` overrides.

crates/io/types/src/lib.rs
```diff
@@ -18,7 +19,28 @@ pub trait Exporter<T>: Send {
     /// Export one event.
     async fn push(&mut self, event: Event<T>) -> ExporterResult<()>;
 
-    /// Make a best-effort to flush any buffered events, then release any internal resources.
+    /// Suggested max events per [`drain_events`](Self::drain_events) batch.
+    ///
+    /// Callers may pass fewer in [`drain_events`], or ignore it completely.
+    fn batch_size_hint(&self) -> NonZeroUsize {
+        NonZeroUsize::new(256).unwrap()
```
9prady9: Nit: this constructs the `NonZeroUsize` and `unwrap`s on every call. Would a static/const constant work here instead (e.g. `const { NonZeroUsize::new(256).unwrap() }` or a module-level `const`), to make it a compile-time value and drop the runtime unwrap?

crates/io/types/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/353/commits/54ad451da07f832001526cbe368cb944128f75ab

crates/io/msgpack/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/353/commits/54ad451da07f832001526cbe368cb944128f75ab

crates/io/types/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/353/commits/54ad451da07f832001526cbe368cb944128f75ab

9prady9 (approved):

PR #370, opened 2026-07-13
chore: add missing instrumentation build example `Cargo.lock`

mbrobbel:
Fixes the failing Rust workflow step:
```
pixi run cargo clippy --manifest-path crates/instrumentation-build/example/Cargo.toml --all-targets --locked -- -D warnings
```


johanpel (approved):
Ah, it seems /merge works even if CI fails. I'll go figure out if we can make the bot disallow this somehow.

Issue #355, opened 2026-07-10
YAML-based application event model definition


johanpel:
The Rust proc macros in `quent-model-macros` are still from the PoC state of the project. They were introduced by the desire to leverage the Rust type system as much as possible to express a model and purely through the Rust compiler. While this is still a longer-term desire, it is also rather complicated (for various reasons that are too detailed to go into here) and they are scheduled for removal.

To make things easier, we set into motion a schema-based Quent migration, where we're going to change the way the application event model is declared,. The first parser will be YAML-based.

johanpel:
The current YAML DSL is at feature parity with the legacy `quent-model` and `quent-model-macros` crates now. See #270 and #383 and further explained in #681 

Imports in the YAML DSL are considered nice-to-haves for now

PR #383, opened 2026-07-13
feat(yaml): scaffold quent-yaml with diagnostics and a spanned YAML tree

johanpel:
# Description

First slice of a YAML source format for application event models. Adds the `quent-yaml` crate with its two foundations, and nothing yet built on top of them:

- **Located diagnostics** (`diag.rs`): a `Diagnostic` (file, line, column, semantic path, message, optional help) and a `Sink` that accumulates them so one load reports every problem instead of stopping at the first. Suggestions use `strsim`.
- **An owned YAML tree** (`tree.rs`) built from the saphyr parser's event stream rather than a deserializer or a ready-made document tree — those drop source positions, silently keep only the last of a duplicated key, and lose a scalar's original text. The event stream keeps every node's position, lets duplicate keys survive until they can be reported, and preserves scalar text. Structural problems abort with one located diagnostic: syntax errors, tags, merge keys, directives, unknown-anchor aliases, and multiple documents. Guards against resource exhaustion: a nesting-depth cap (a deeply nested block document would otherwise overflow the tree's recursive drop) and a byte budget on anchor/alias reuse (Billion Laughs). Strips a leading byte-order mark.

The format itself — mapping access, the type mini-language, lowering to a `quent_schema::Schema`, and the load API — follows in later PRs in this stack. `#![allow(dead_code)]` is intentional for this slice; the consumers land next.

## Related Issues

Part 1/6 of #355

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2697,6 +2697,7 @@ name = "quent-yaml"
 version = "0.1.0"
 dependencies = [
  "saphyr-parser",
+ "strsim",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3095,9 +3095,9 @@ dependencies = [
 
 [[package]]
 name = "saphyr-parser"
-version = "0.0.9"
+version = "0.0.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


johanpel:
closing this in favor of experimenting a bit more in https://github.com/rapidsai/quent/pull/270/changes

PR #384, opened 2026-07-13
feat(timeline): pan time window with two-finger horizontal scroll

felipeblazing:
## Summary

Adds two-finger horizontal trackpad scrolling to pan the time window across the timeline view. When the pointer is over a timeline, a horizontal swipe slides the visible window left/right through time.

## How it works

A capture-phase, non-passive `wheel` listener on the Timeline chart:
- Detects a predominantly horizontal swipe (`|deltaX| > |deltaY|`), ignoring vertical scroll and shift+scroll zoom.
- Calls `preventDefault()` to suppress browser horizontal page scroll.
- Converts the pixel delta into a percentage of the current visible span and dispatches an ECharts `dataZoom` action, clamped to `[0, 100]`.

The action propagates through the existing ECharts connect group → `TimelineController` fires `onZoomChange` → `zoomRangeAtom` updates, so all stacked timelines, the controller bar, and the ruler stay in sync. Data re-fetches use the existing 150ms zoom debounce.

## Behavior preserved
- Vertical two-finger scroll → normal page scroll (unchanged).
- Shift + scroll → zoom (unchanged).

## Testing
Verified locally against the simulator backend with the Vite dev server: zoom in, then two-finger horizontal swipe pans the window; all charts and the ruler track together.

### DIFF
--- a/ui/packages/@quent/components/src/timeline/Timeline.tsx
+++ b/ui/packages/@quent/components/src/timeline/Timeline.tsx
@@ -370,6 +370,45 @@ export function Timeline({
       }
     });
 
+    // Two-finger horizontal trackpad scroll → pan the time window.
+    // Must be a capture-phase, non-passive listener so it fires before ECharts'
+    // child canvas sees the event, and so preventDefault() is permitted.
+    dom.addEventListener(
+      'wheel',
+      e => {
+        if (e.deltaX === 0 || Math.abs(e.deltaX) <= Math.abs(e.deltaY)) return;
+        if (instance.isDisposed?.()) return;
+
+        e.preventDefault();
+
+        const opt = instance.getOption() as {
+          dataZoom?: Array<{ start?: number; end?: number }>;
+        };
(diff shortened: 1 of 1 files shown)


ui/packages/@quent/components/src/timeline/Timeline.tsx
```diff
@@ -370,6 +370,49 @@ export function Timeline({
       }
     });
 
+    // Two-finger horizontal trackpad scroll → pan the time window.
```
cmatzenbach, line 373: One thing we're trying to do when reviewing is clean up AI comments, as they're often overly verbose to the point of being confusing. I'd say either just keep this one line, or shorten the other two to "Utilizes `preventDefault()` so that listener fires before ECharts' child canvas sees the event"

felipeblazing: done

cmatzenbach (approved):
Looks great, fix worked perfectly

PR #385, opened 2026-07-14
feat(nvtx): Phase 1 — NVTX capture foundation (events, injection cdylib, bridge + e2e)

9prady9:
## Phase 1 — NVTX Capture Foundation

Adds a Linux/macOS-64 NVTX ingestion path so Quent can capture NVTX events emitted by
GPU-accelerated libraries (libcudf, cuCascade, …) and flow them into the standard event
pipeline — driven by a deterministic, GPU-less in-repo test app.

Draft: opening early for design/API review. Follow-ups (see *Deferred* below) are intentional
scope, not omissions.

### Review groups (3 crate seams — one commit each)

1. **`quent-nvtx-events`** — verbatim, Quent-agnostic NVTX vocabulary. `NvtxEvent` covering the
   core surface (push/pop + start/end ranges, marks, domain lifecycle, registered strings,
   category/thread naming, resource create/destroy); the CORE `nvtxEventAttributes` payload union
   captured **undecoded**. No Quent-internal deps (kept separable for potential NVTX upstreaming).
2. **`quent-nvtx-injection`** — the cdylib NVTX loads via `NVTX_INJECTION64_PATH`. Exports
   `InitializeInjectionNvtx2`, one-shot CORE/CORE2 callback-table fill, verbatim conversion.
   `catch_unwind` at every C-ABI boundary; caller strings copied in-callback; size-bounded
   attribute reads. Bindings are **committed + hermetic** (default/offline builds compile no C and
   fetch no NVTX headers); regeneration is feature-gated with NVTX pinned to a git rev and
   allow-listed in `deny.toml`.
3. **`quent-nvtx`** — the bridge + self-configuring capture cdylib. A bounded lock-free
   `ArrayQueue` ring with drop-and-count (the app thread never blocks) drained off-thread into an
   `EventSender`; each event carries a capture timestamp. An ELF `.init_array` constructor installs
   an ndjson sink at `dlopen` (off the loader lock); `.fini_array` flushes at exit with a bounded
   watchdog join so a stuck flush can't wedge process exit. Includes a **Quent-free deterministic
   NVTX test app** and a **GPU-less subprocess e2e harness**.

### How capture attaches

`NVTX_INJECTION64_PATH=…/libquent_nvtx.so` + `QUENT_NVTX_OUTPUT_DIR=<dir>` → NVTX `dlopen`s the
cdylib → `.init_array` installs the sink → push/pop/etc. callbacks land timestamped `NvtxEvent`s in
a real `.ndjson` file. The instrumented app needs **no code changes** and links nothing from Quent.
(The self-configuring cdylib is required because `dlopen` is `RTLD_LOCAL`: a hook installed in an
app-linked copy is a different, unreachable static from the one NVTX's callbacks read.)

### Testing (no GPU required)

```
# the e2e test needs the capture cdylib built first
pixi run cargo build -p quent-nvtx --features e2e
pixi run cargo test  -p quent-nvtx --features e2e
pixi run cargo test  -p quent-nvtx-injection -p quent-nvtx-events
```

The e2e subprocess test asserts every captured kind, the payload union round-tripping verbatim,
per-event timestamps, and cross-thread `RangeStart`/`RangeEnd` pairing. Default/offline builds stay
hermetic (no `libclang`, no NVTX fetch).

### Deferred (intentional scope)

- **Classic default-domain + wide-char (`*W`) NVTX APIs** are diagnosed-but-not-captured (a one-time
  runtime warning + doc note). v1 target libraries are domain-scoped; target is Linux-only.
- **True byte-verbatim strings** — non-UTF-8 NVTX strings are currently captured lossily with a
  one-time diagnostic; a `Vec<u8>` vocabulary change is deferred.
- Payload **extension** (schemas/enums/binary `nvtxPayloadData_t`) is deferred to a later phase
  (libcudf emits none today).

### Constraints

Linux/macOS 64-bit only (`compile_error!` elsewhere) — NVTX injection relies on
weak-symbol override / `NVTX_INJECTION64_PATH`.


9prady9:
Right now, all three commits are reviewable on their own. However, given the size of this PR, I am wondering if this should be split up.

integrations/nvtx/injection/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "quent-nvtx-injection"
```
mbrobbel, line 5: I suggest we upstream a feature to nvtx-sys that adds the required items (behind a `tools` cfg feature) to the bindgen allowlist, instead of introducing this crate here.

integrations/nvtx/events/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "quent-nvtx-events"
+version.workspace = true
+edition.workspace = true
+publish.workspace = true
+
+[dependencies]
+serde.workspace = true
+uuid = { workspace = true, features = ["serde"] }
```
mbrobbel, line 9: This is unused?

integrations/nvtx/events/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "quent-nvtx-events"
```
mbrobbel, line 2: Same suggestion here; let's figure out how to upstream the generic `subscriber` parts to the `nvtx` crate.

integrations/nvtx/injection/build.rs
```diff
@@ -0,0 +1,113 @@
...
+        .header("wrapper.h")
+        .clang_arg(format!("-I{}", include_dir.display()))
+        .raw_line("// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.")
+        .raw_line("// SPDX-License-Identifier: Apache-2.0")
+        .raw_line("//")
+        .raw_line("//! Committed bindgen output for the NVTX injection ABI (D-14).")
+        .raw_line("//!")
+        .raw_line("//! Generated by `build.rs` under the `regenerate-bindings` feature from the")
+        .raw_line("//! pinned NVIDIA/NVTX headers (`c/include/nvtx3/nvToolsExt.h`, rev v3.5.0 /")
+        .raw_line("//! 7d113f290f89eeeae9c957011c497101f3948d9e). Do NOT edit by hand; re-run")
```
mbrobbel, line 50: This won't update automatically?

mbrobbel (commented):
+1 for splitting this up

integrations/nvtx/events/Cargo.toml
9prady9: good catch, removing it.

integrations/nvtx/injection/Cargo.toml
9prady9: Agreed the bindgen allowlist items belong upstream in nvtx-sys behind a tools feature. Let's pursue that in parallel and switch quent over once it lands — keeping this crate separable will work in favor of that — rather than blocking this capability on the upstream round-trip.

integrations/nvtx/events/Cargo.toml
9prady9: agree, we can follow same approach as injection crate.

integrations/nvtx/injection/build.rs
9prady9: That is by design actually - the bindings are committed so default/CI builds stay hermetic (no libclang, no NVTX fetch). Regeneration is a manual `--features regenerate-bindings` step pinned to the NVTX `rev`, so a bump is a deliberate, reviewable commit rather than silent drift.

9prady9:
First PR https://github.com/rapidsai/quent/pull/386 extracted from this PR's changes. @mbrobbel @johanpel

integrations/nvtx/injection/Cargo.toml
mbrobbel: Let's fork nvtx + apply the patch and use that here until we land the changes upstream. This avoids introducing all this code here.

integrations/nvtx/injection/build.rs
mbrobbel: Yes, but the docs don't mention updating this bit.

9prady9:
Split into 3 PRs linked to this PR.

Issue #76, opened 2026-03-26
Consume NVTX ranges


johanpel:
In order to get more insight into GPU-accelerated data processing support libraries, including `libcudf`, `cuCascade`, and more, Quent should be able to consume NVTX ranges to form traces. There is already a "trace" type entity into which this can be molded, or we can adjust it to better fit NVTX's semantics.

There are two challenges:
1. AFAICT there are no Rust bindings for NVTX that expose its "injection" API (also see the example: https://github.com/NVIDIA/NVTX/tree/dev/tools/sample-injection).
2. It is non-trivial to have multiple NVTX consumers, if those consumers solely rely on the `NVTX_INJECTION64_PATH` environment variable to inject themselves into a process.

9prady9:
## Minimal reproduction: static injection is DSO-local

Sirius is a DuckDB extension that depends on libcudf. Both contain NVTX 3.2.0 header state. Inspection of the built binaries shows a separate local `nvtxGlobals_v3` and `InitializeInjectionNvtx2_fnptr` in `sirius.duckdb_extension` and in `libcudf.so`.

We reproduced that topology with the real Quent injection code, a plugin DSO, and a dependent DSO standing in for libcudf. The plugin installed Quent's hook and linked the documented strong `InitializeInjectionNvtx2_fnptr`. Each DSO then emitted one `nvtxMarkA`.

Results (`after_plugin`, then `after_dependency`):

```text
static injection, RTLD_LOCAL:  1, 1
static injection, RTLD_GLOBAL: 1, 1
NVTX_INJECTION64_PATH:         1, 2
```

Static injection worked for NVTX calls compiled into the plugin, but not for calls compiled into the dependent DSO. Changing `RTLD_LOCAL` to `RTLD_GLOBAL` made no difference.

### Why the original Sirius run also missed Sirius events

The successful static MRE retained the strong injection shim in the final plugin DSO. The original Sirius build did not. Its intermediate `libtelemetry_bridge.a` contained both a strong initialized `InitializeInjectionNvtx2_fnptr` and `InitializeInjectionNvtx2`, but the final `sirius.duckdb_extension` contained only the null NVTX placeholder and no `InitializeInjectionNvtx2` implementation.

[...]

PR #386, opened 2026-07-14
feat(nvtx): add quent-nvtx-events — verbatim NVTX event vocabulary

9prady9:
## `quent-nvtx-events` — verbatim NVTX event vocabulary

First of three PRs splitting the NVTX capture foundation out of #385 (per the request to split it up). Stack:

1. **`quent-nvtx-events`** — this PR (vocabulary types)
2. `quent-nvtx-injection` — cdylib NVTX loads via `NVTX_INJECTION64_PATH` (follows once this lands)
3. `quent-nvtx` — bridge + deterministic test app + capture e2e (follows injection)

### What this adds

A Quent-agnostic, **verbatim** vocabulary of NVTX events — the raw shapes NVTX hands us, with no interpretation:

- `NvtxEvent` covering the core surface: push/pop + start/end ranges, marks, domain lifecycle, registered strings, category/thread naming, resource create/destroy.
- `NvtxMessage` / `NvtxColor` / `NvtxEventAttributes`, and the core `nvtxEventAttributes` payload union captured **undecoded** (`NvtxPayload`).
- Handle resolution and payload decoding are deliberately *not* done here — a later analysis stage resolves handles from the event stream.

### Pure vocabulary, separable by design

- **No Quent-internal dependencies.** `serde` is the only dependency and it's **optional** (default-on `serde` feature), so a consumer that only needs the in-memory types doesn't pull it — compiles with and without the feature.
- **No pipeline concerns here.** Entity naming and the event wrapper live in the bridge crate (PR 3), not in this crate — keeping it a clean, upstreamable data vocabulary.

> Re upstreaming the generic vocabulary into `nvtx` / `nvtx-sys`: agreed in principle and being pursued in parallel — the crate is kept dependency-free precisely so that move is clean. `nvtx-sys` is currently producer-only (no consumer/subscriber surface), so upstreaming is a parallel track rather than a blocker for landing capture.

### Tests

None — the crate is pure serde-derived data types, so there's nothing crate-specific to assert beyond what serde itself guarantees (round-trip tests removed per review).

### DIFF
--- a/integrations/nvtx/events/src/attributes.rs
+++ b/integrations/nvtx/events/src/attributes.rs
@@ -1,0 +1,57 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Captured subset of `nvtxEventAttributes_t`.
+//!
+//! These types mirror the raw NVTX attribute members verbatim (message, color,
+//! category, payload). Handles are never resolved at capture time — the analyzer
+//! resolves registered strings, domains, and categories from the event stream in
+//! a later phase.
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+use crate::payload::NvtxPayload;
+
+/// A message attached to an NVTX event.
+///
+/// Registered messages keep only their raw handle, never resolved at capture
+/// time. The analyzer maps [`NvtxMessage::RegisteredHandle`] back to its string
+/// from the captured [`RegisterString`](crate::NvtxEvent::RegisterString) events.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub enum NvtxMessage {
+    /// An immediate string message, copied verbatim at capture.
+    String(String),
+    /// A handle to a previously registered string; resolved in the analyzer.
+    RegisteredHandle(u64),
+}
+
+/// A verbatim NVTX color attribute: the raw `nvtxColorType_t` tag paired with
+/// the raw color value (e.g. `NVTX_COLOR_ARGB`).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub struct NvtxColor {
+    /// Raw `nvtxColorType_t` tag.
+    pub color_type: i32,
+    /// Raw color value (e.g. packed ARGB).
+    pub value: u32,
+}
+
+/// Captured subset of `nvtxEventAttributes_t`.
+///
+/// Only the members Quent reconstructs from are retained (`category`, `color`,
+/// `message`, `payload`); all are stored verbatim with no capture-time
+/// resolution or decoding.
+#[derive(Debug, Clone, PartialEq, Default)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub struct NvtxEventAttributes {
+    /// Raw category id (`0` = none). Namespaced by domain in the analyzer.
+    pub category: u32,
+    /// Optional color attribute.
+    pub color: Option<NvtxColor>,
+    /// Optional message (immediate string or registered handle).
+    pub message: Option<NvtxMessage>,
+    /// Optional payload union value from the core `nvtxEventAttributes`.
+    pub payload: Option<NvtxPayload>,
+}
--- a/integrations/nvtx/events/src/lib.rs
+++ b/integrations/nvtx/events/src/lib.rs
@@ -1,0 +1,126 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Verbatim, Quent-agnostic NVTX event vocabulary.
+//!
+//! Every downstream NVTX crate speaks this shared contract: the injection cdylib
+//! produces [`NvtxEvent`]s and the bridge forwards them into Quent. Events are
+//! captured **verbatim** — every handle (domain / category / resource /
+//! registered-string id) is a raw integer, and no name resolution or payload
+//! decoding happens at capture time. Handles are resolved from the event stream
+//! by a later analysis stage.
+//!
+//! The crate deliberately depends on nothing Quent-internal (optionally only
+//! `serde`, behind the default `serde` feature) so it stays cleanly separable and
+//! could be offered upstream to the NVTX Rust crates later. Adapting these events
+//! into Quent's pipeline — entity naming, the event wrapper — is the bridge
+//! crate's responsibility, not this crate's.
+
+mod attributes;
+mod payload;
+
+pub use attributes::{NvtxColor, NvtxEventAttributes, NvtxMessage};
+pub use payload::{NvtxPayload, NvtxPayloadValue, PayloadExtensionEvent};
+
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
+
+/// A verbatim NVTX core event.
+///
+/// Every variant mirrors one core NVTX call kind. Handles are raw integers,
+/// captured with no resolution. The default (NULL) domain is represented as a
+/// `domain` of `0`.
+#[derive(Debug, Clone, PartialEq)]
+#[cfg_attr(feature = "serde", derive(Deserialize, Serialize))]
+pub enum NvtxEvent {
+    /// `nvtxDomainRangePushEx` — open a nested (per-thread) range.
+    RangePush {
+        /// Raw domain handle (`0` = default domain).
+        domain: u64,
+        /// Captured event attributes (message, color, category, payload).
+        attributes: NvtxEventAttributes,
+    },
+    /// `nvtxDomainRangePop` — close the most recent push on this thread.
+    RangePop {
+        /// Raw domain handle (`0` = default domain).
+        domain: u64,
+    },
+    /// `nvtxDomainRangeStartEx` — open a process-wide range keyed by id.
+    RangeStart {
+        /// Raw domain handle (`0` = default domain).
+        domain: u64,
+        /// Raw `nvtxRangeId_t` correlating start and end.
+        range_id: u64,
+        /// Captured event attributes (message, color, category, payload).
+        attributes: NvtxEventAttributes,
+    },
+    /// `nvtxDomainRangeEnd` — close the range with the matching id.
+    RangeEnd {
+        /// Raw domain handle (`0` = default domain).
+        domain: u64,
+        /// Raw `nvtxRangeId_t` correlating start and end.
+        range_id: u64,
+    },
+    /// `nvtxDomainMarkEx` — an instantaneous marker.
+    Mark {
+        /// Raw domain handle (`0` = default domain).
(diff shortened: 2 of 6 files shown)


integrations/nvtx/events/src/attributes.rs
```diff
@@ -0,0 +1,54 @@
...
+//! stream in a later phase.
+
+use serde::{Deserialize, Serialize};
+
+use crate::payload::NvtxPayload;
+
+/// A message attached to an NVTX event.
+///
+/// Registered messages keep only their raw handle — never resolved at capture
+/// time (D-01). The analyzer maps [`NvtxMessage::RegisteredHandle`] back to its
```
johanpel: What is 

> D-01

?

integrations/nvtx/events/src/attributes.rs
```diff
@@ -0,0 +1,54 @@
...
+/// resolution or decoding.
+#[derive(Debug, Clone, PartialEq, Default, Deserialize, Serialize)]
+pub struct NvtxEventAttributes {
+    /// Raw category id (`0` = none). Namespaced by domain in the analyzer.
+    pub category: u32,
+    /// Optional color attribute.
+    pub color: Option<NvtxColor>,
+    /// Optional message (immediate string or registered handle).
+    pub message: Option<NvtxMessage>,
+    /// Optional CORE payload union value.
```
johanpel: > CORE

What does this stand for (assuming it's an abbreviation)?

integrations/nvtx/events/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+//! Verbatim, Quent-agnostic NVTX event vocabulary.
+//!
+//! This crate is the shared contract every downstream NVTX crate speaks: the
+//! injection cdylib produces [`NvtxEvent`]s and the bridge wraps them for the
+//! Quent pipeline. Events are captured **verbatim** — every handle
+//! (domain/category/resource/registered-string id) is a raw integer, and no name
+//! resolution or payload decoding happens at capture time (D-01). The analyzer
+//! resolves handles from the event stream in a later phase.
+//!
+//! # Separability (D-03)
```
johanpel: What is

> D-03

?

integrations/nvtx/events/src/attributes.rs
```diff
@@ -0,0 +1,54 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Captured subset of `nvtxEventAttributes_t`.
+//!
+//! These types mirror the raw NVTX attribute members verbatim (message, color,
+//! category, payload). Handles are never resolved at capture time (D-01) — the
+//! analyzer resolves registered strings, domains, and categories from the event
+//! stream in a later phase.
+
+use serde::{Deserialize, Serialize};
```
johanpel, line 12: We might want to feature gate serde support since not all exporters require it since #250 was solved

integrations/nvtx/events/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+
+pub use attributes::{NvtxColor, NvtxEventAttributes, NvtxMessage};
+pub use payload::{NvtxPayload, NvtxPayloadValue, PayloadExtensionEvent};
+
+use serde::{Deserialize, Serialize};
+
+/// The entity-name contract for an event-stream payload type.
+///
+/// Mirrored locally to keep this vocabulary crate free of Quent-internal
+/// dependencies (D-03 separability). It is structurally identical to the Quent
```
johanpel: What is 
> D-03
?

I'll stop flagging these as it seems to be a pattern, but let's make sure these comments don't refer to things invisible to other developers.

integrations/nvtx/events/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+/// pipeline's `quent_events::EntityEvent`; the NVTX bridge adapts
+/// [`NvtxEventKind`] into that pipeline.
+pub trait EntityEvent {
+    /// The name of the entity producing these events.
+    const NAME: &'static str;
+}
```
johanpel: If this crate is trying to be Quent-agnostic and (close to) upstreamable, then I don't think it is necessary to include this trait here at all.

integrations/nvtx/events/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+
+impl From<NvtxEvent> for NvtxEventKind {
+    fn from(event: NvtxEvent) -> Self {
+        Self(event)
+    }
+}
+
+impl EntityEvent for NvtxEventKind {
+    const NAME: &'static str = "NvtxEvent";
+}
```
johanpel: For the same reason as above, I think this should be a concern of the bridging code only.

integrations/nvtx/events/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+    fn nvtx_message_variants_round_trip() {
+        for message in [
+            NvtxMessage::String("immediate".to_owned()),
+            NvtxMessage::RegisteredHandle(0xABCD),
+        ] {
+            let json = serde_json::to_string(&message).expect("serialize");
+            let back: NvtxMessage = serde_json::from_str(&json).expect("deserialize");
+            assert_eq!(back, message);
+        }
+    }
```
johanpel: This seems a but superfluous as it tests serde + serde_json more than any functionality our sources here capture. I'm sure those crates have stellar coverage already.

integrations/nvtx/events/src/lib.rs
```diff
@@ -0,0 +1,291 @@
...
+        let kind = NvtxEventKind::from(NvtxEvent::RangePop { domain: 0 });
+        // Transparent newtype: serializes identically to the inner event.
+        let kind_json = serde_json::to_string(&kind).expect("serialize kind");
+        let event_json =
+            serde_json::to_string(&NvtxEvent::RangePop { domain: 0 }).expect("serialize event");
+        assert_eq!(kind_json, event_json);
+
+        let back: NvtxEventKind = serde_json::from_str(&kind_json).expect("deserialize");
+        assert_eq!(back, kind);
+    }
```
johanpel: Same as above were we to remove the trait from this crate

integrations/nvtx/events/src/payload.rs
```diff
@@ -0,0 +1,148 @@
...
+    #[test]
+    fn deferred_payload_extension_vocabulary_round_trips() {
+        let event = PayloadExtensionEvent::BinaryPayload {
+            schema_id: 9,
+            bytes: vec![1, 2, 3, 4],
+        };
+        let json = serde_json::to_string(&event).expect("serialize");
+        let back: PayloadExtensionEvent = serde_json::from_str(&json).expect("deserialize");
+        assert_eq!(back, event);
+    }
```
johanpel: Same as above

johanpel (commented):
Thanks for the PR, some comments:

integrations/nvtx/events/src/lib.rs
9prady9: Yeah, no need flag all of them. I usually run a pass to remove some internal flags. Looks like these got added while extracting this PR from the other one #385 

Leaving this but marking all other similar comments as resolved.

johanpel:
From a code perspective this looks fine to me but I think we need to reconsider adding it as a Quent crate and follow the suggestion here: https://github.com/rapidsai/quent/pull/385#discussion_r3577309193

9prady9:
> From a code perspective this looks fine to me but I think we need to reconsider adding it as a Quent crate and follow the suggestion here: [#385 (comment)](https://github.com/rapidsai/quent/pull/385#discussion_r3577309193)

I think we should keep this and look into that in parallel. There is no need to block the follow tasks and pursue this which can be delayed with fork creation and setting up other things.

Keeping this as a separate crate lets us remove this at later point quickly and switch over to the nvtx-sys upstream.

mbrobbel:
> > From a code perspective this looks fine to me but I think we need to reconsider adding it as a Quent crate and follow the suggestion here: [#385 (comment)](https://github.com/rapidsai/quent/pull/385#discussion_r3577309193)
> 
> I think we should keep this and look into that in parallel. There is no need to block the follow tasks and pursue this which can be delayed with fork creation and setting up other things.
> 
> Keeping this as a separate crate lets us remove this at later point quickly and switch over to the nvtx-sys upstream.

I'm curious, with a fork we are not blocked, so why not do that now?

9prady9:
> > > From a code perspective this looks fine to me but I think we need to reconsider adding it as a Quent crate and follow the suggestion here: [#385 (comment)](https://github.com/rapidsai/quent/pull/385#discussion_r3577309193)
> > 
> > 
> > I think we should keep this and look into that in parallel. There is no need to block the follow tasks and pursue this which can be delayed with fork creation and setting up other things.
> > Keeping this as a separate crate lets us remove this at later point quickly and switch over to the nvtx-sys upstream.
> 
> I'm curious, with a fork we are not blocked, so why not do that now?

If and when I do any changes to these crates, one would have to jump through couple of extra steps to achieve that. Instead if move/upstream them later, I can play with them quickly and upstream them once they are upstream ready. Potentially less work for me with the trade off being some extra bunch of lines in the repository temporarily.

I am confident it won't stay in the same shape/form as it is today given how things change as we progress through next 4 phases listed as sub-issues in #76

johanpel (approved):
> If and when I do any changes to these crates, one would have to jump through couple of extra steps to achieve that. Instead if move/upstream them later, I can play with them quickly and upstream them once they are upstream ready. Potentially less work for me with the trade off being some extra bunch of lines in the repository temporarily.

To make sure we are on the same page (perhaps this is clear already), the suggestion of @mbrobbel is to create a personal fork NVTX, add the Rust injection wrappers there, then use a [Cargo git dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).

You can iterate quickly on the fork without upstreaming anything to NVTX until you feel it's done. Only takes a bump of the git rev to synchronize changes. Less work in the end; when you're ready to upstream to NVTX, just open a PR from the fork, get it merged and update the git dep, no code migration needed. And you can leverage NVTX' existing build script to generate sys bindings.

Ultimately your choice, approving to unblock merging using the current approach if you prefer that.

9prady9:
> > If and when I do any changes to these crates, one would have to jump through couple of extra steps to achieve that. Instead if move/upstream them later, I can play with them quickly and upstream them once they are upstream ready. Potentially less work for me with the trade off being some extra bunch of lines in the repository temporarily.
> 
> To make sure we are on the same page (perhaps this is clear already), the suggestion of @mbrobbel is to create a personal fork NVTX, add the Rust injection wrappers there, then use a [Cargo git dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).
> 
> You can iterate quickly on the fork without upstreaming anything to NVTX until you feel it's done. Only takes a bump of the git rev to synchronize changes. Less work in the end; when you're ready to upstream to NVTX, just open a PR from the fork, get it merged and update the git dep, no code migration needed. And you can leverage NVTX' existing build script to generate sys bindings.
> 
> Ultimately your choice, approving to unblock merging using the current approach if you prefer that.

That's true, it shifts the task to host this code temporally. However, I have to shift to different repository (locally of course) whenever I want edit the code from those crates. Yes, it is few simple steps, nevertheless some extra hops to achieve it compared to just editing the file in the sibling folder.

Hence my hesitation to do it right away. Lets do it in place and handle the shifting part later.

9prady9:
Created ticket for upstreaming task https://github.com/rapidsai/quent/issues/388

Issue #371, opened 2026-07-13
Phase 1: Capture Foundation


9prady9:
**What this delivers:** the ability to capture NVTX activity from a running app into Quent's pipeline as raw events.

**In scope**
- Attach to an app with no code changes and capture its NVTX calls faithfully
- A deterministic in-repo test app + a test harness that runs in CI without a GPU
- Capture stays off the app's hot path

**Out of scope (later phases)**
- Turning raw events into readable ranges
- Showing anything in the UI
- Running alongside Nsight/other tools
- Decoding payload data

**Deliverables**
- Injection library that captures raw NVTX events
- In-repo NVTX test app + CI harness

**Done when**
- [x] All common NVTX calls captured as raw events (ranges, marks, domains, named strings, categories, named threads, resources)
- [ ] Payload/extension data captured verbatim
- [x] Events timestamped + reach an exporter
- [x] CI green with no GPU
- [ ] No app stalls under heavy load

---
Part of #76 · Phase 1 of 5

9prady9:
Addressed via #386 #391 #402 and #403

PR #391, opened 2026-07-15
feat(nvtx): add quent-nvtx-injection cdylib (separable NVTX FFI)

9prady9:
## `quent-nvtx-injection` — the NVTX injection cdylib

PR 2 of 3, following #386 (`quent-nvtx-events`, merged). The bridge + capture e2e follows this one.

### What it is

- The Quent-agnostic cdylib NVTX loads via `NVTX_INJECTION64_PATH`. Exports `InitializeInjectionNvtx2`, fills the CORE/CORE2 callback tables, and converts each callback into the `quent-nvtx-events` vocabulary **verbatim** (raw handles, no resolution/decoding).
- Depends on nothing Quent-internal except `quent-nvtx-events`. Linux 64-bit only (`compile_error!` elsewhere).

### Safety & correctness

- **Panic containment** — every `extern "C"` callback runs inside `catch_unwind`; a panic never crosses the C ABI. Synthesized handles/ids and push/pop levels are returned correctly even if conversion panics.
- **Per-image init** — NVTX calls `InitializeInjectionNvtx2` once per NVTX-using image (exe + each instrumented `.so` has its own `nvtxGlobals`), so callbacks are installed on **every** call, not just the first — otherwise later images are silently no-op'd.
- **Bounded reads** — attribute structs are read only up to their declared `size`; the payload union is read at the tagged member's exact width (no uninitialized upper bytes).
- **String copy-in** — immediate `const char*` messages are copied into owned `String`s before returning; registered strings keep only their raw handle.
- **Faithful push/pop** — range push/pop return NVTX's 0-based per-thread, per-domain nesting level.
- **Bounds-checked table writes** — callbacks are written only into in-range, non-null NVTX slots.

### Generated bindings

- `build.rs` runs `bindgen` over the NVTX ABI into `$OUT_DIR` on every build — nothing checked in, no `bindings.rs` to keep in sync (resource/range-id types folded into the allowlist; no hand-written `convert::abi`).
- Headers + `libclang` come from the pixi-pinned `nvtx-c` / `libclang` packages (`[target.linux-64.dependencies]`), so `pixi run cargo …` is hermetic (`build.rs` reads `CONDA_PREFIX`). Bumping the captured NVTX version is a one-line `nvtx-c` bump — no regen feature, NVTX git dep, or `deny.toml` allow-git.

### Captured surface

- **CORE2 (domain-scoped ASCII):** mark, range start/end/push/pop, domain / register-string / name-category / resource.
- **CORE (classic default-domain ASCII):** `nvtxMarkA`/`Ex`, `nvtxRangePushA`/`Ex`, `nvtxRangePop`, `nvtxRangeStartA`/`Ex`/`nvtxRangeEnd`, `nvtxNameCategoryA`, `nvtxNameOsThreadA` — captured on the default domain (`0`). No vocabulary change: `quent-nvtx-events` already models the default domain as `domain 0`.
- **Wide-char (`*W`/Unicode):** subscribed with warn-once stubs — Unicode capture deferred, but such a call emits a one-time diagnostic and keeps range nesting/ids valid instead of silently no-op'ing.

### Tests

- 17 unit tests covering conversion and nesting logic (verbatim capture per kind, size-bounded reads, string copy-in, member-width payloads, per-thread/per-domain nesting, default-domain `*A` conversions). No NVTX headers or GPU required beyond the pixi-pinned `nvtx-c`.

### Upstreaming

- The injection ABI allowlist / consumer surface is a candidate to contribute to `nvtx-sys` behind a `tools` feature (currently producer-only) — parallel track, not a blocker.


deny.toml
```diff
@@ -54,4 +54,8 @@ unknown-registry = "deny"
 # not in the allow list is encountered.
 unknown-git = "deny"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
-allow-git = []
+# NVIDIA/NVTX is the pinned header source for `quent-nvtx-injection`'s
+# `regenerate-bindings` feature (D-13). Fetched only when regenerating the
```
mbrobbel: D-13

integrations/nvtx/injection/build.rs
```diff
@@ -0,0 +1,129 @@
...
+
+//! Build script for `quent-nvtx-injection`.
+//!
+//! Default/CI builds do nothing but declare the header dependency and
+//! `include!` the committed `src/bindings.rs` — no bindgen, no libclang,
+//! no NVTX git fetch. Bindgen runs ONLY under the `regenerate-bindings` feature;
+//! the `static-injection` feature compiles the strong-symbol C shim.
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    println!("cargo::rerun-if-changed=wrapper.h");
```
mbrobbel: I guess we only want to invalidate when re-generating the bindings?
```suggestion
    #[cfg(feature = "regenerate-bindings")]
    println!("cargo::rerun-if-changed=wrapper.h");
```

integrations/nvtx/injection/src/callbacks.rs
```diff
@@ -0,0 +1,194 @@
...
+    });
+    handle as usize as nvtxResourceHandle_t
+}
+
+/// CORE2 `DomainResourceDestroy` subscriber.
+pub(crate) extern "C" fn on_domain_resource_destroy(resource: nvtxResourceHandle_t) -> c_int {
+    let _ = std::panic::catch_unwind(|| {
+        init::dispatch(convert::resource_destroy(resource as usize as u64));
+    });
+    0
```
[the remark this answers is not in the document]
9prady9: A real table installation is covered by the bridge's subprocess e2e which is gng to be handled in the follow up PR - last PR in phase 1 i.e. #371

integrations/nvtx/injection/src/lib.rs
```diff
@@ -0,0 +1,24 @@
...
+//! installs the CORE/CORE2 callback tables one-shot. Push/pop calls are
+//! converted to verbatim [`NvtxEvent`](quent_nvtx_events::NvtxEvent)s and handed
+//! to a sink-agnostic `Fn(NvtxEvent)` hook installed via
+//! [`install_hook`](crate::install_hook). This crate depends on nothing
+//! Quent-internal except `quent-nvtx-events`, so it stays separable/upstreamable.
+
+// Linux 64-bit only. NVTX injection relies on the ELF weak-symbol /
+// NVTX_INJECTION64_PATH mechanism; Windows and 32-bit are out of scope.
+#[cfg(not(all(target_os = "linux", target_pointer_width = "64")))]
+compile_error!("quent-nvtx-injection supports Linux 64-bit only");
```
johanpel, line 17: Since the crate is now part of the workspace default-members this will cause builds to error out on other platforms so we might wan to remove it from there.

This may be sensible in general for integrations (to make any of them opt-in).

integrations/nvtx/README.md
```diff
@@ -0,0 +1,114 @@
...
+- [How capture works](#how-capture-works)
+- [NVTX bindings (committed bindgen output)](#nvtx-bindings-committed-bindgen-output)
+- [Upstreaming](#upstreaming)
+
+## Crates
+
+| Crate | Path | Role |
+|-------|------|------|
+| `quent-nvtx-events` | `events/` | The verbatim, Quent-agnostic NVTX event **vocabulary** (`NvtxEvent` + attribute/payload types). Pure Rust — no C, no NVTX headers, and nothing Quent-internal — so it can be offered upstream to the NVTX Rust crates later. |
+| `quent-nvtx-injection` | `injection/` | The Quent-agnostic injection **cdylib** NVTX loads. Exports `InitializeInjectionNvtx2`, fills the NVTX callback tables, converts each call into a verbatim `NvtxEvent`, and hands it to a sink-agnostic `Fn(NvtxEvent)` hook. This is the only crate bound to the NVTX C ABI. |
```
mbrobbel, line 30: Nit: if these are Quent-agnostic, we should maybe remove the `quent-` prefix?

integrations/nvtx/injection/src/init.rs
```diff
@@ -0,0 +1,392 @@
...
+    };
+    use NvtxCallbackIdCore2 as Cb;
+
+    subscribe!(
+        table,
+        size,
+        Cb::NVTX_CBID_CORE2_DomainMarkEx,
+        callbacks::on_domain_mark_ex,
+        extern "C" fn(nvtxDomainHandle_t, *const nvtxEventAttributes_t) -> c_int
+    );
```
johanpel: Looking at the pinned headers shouldn't this be:

```suggestion
    subscribe!(
        table,
        size,
        Cb::NVTX_CBID_CORE2_DomainMarkEx,
        callbacks::on_domain_mark_ex,
        extern "C" fn(nvtxDomainHandle_t, *const nvtxEventAttributes_t)
    );
```
since it returns void?

There may be more of these.

deny.toml
9prady9: damn, despite repetitve efforts, these keep creeping up 🤦🏾‍♂️

integrations/nvtx/README.md
9prady9: Let me handle that rename in a follow up PR which purely rename only so that its easy on review.

integrations/nvtx/injection/src/init.rs
9prady9: https://github.com/rapidsai/quent/pull/391/commits/62bfae2bac8005e09ad1a0e7ede342e192f7aa90

integrations/nvtx/injection/build.rs
9prady9: https://github.com/rapidsai/quent/pull/391/commits/2442359443ecf537e9cf71ee369208dc6d4b7446

integrations/nvtx/injection/src/lib.rs
9prady9: I removed them from default members, we still build and test on CI. It's just not used from simple build or test commands.

integrations/nvtx/injection/src/init.rs
```diff
@@ -0,0 +1,392 @@
...
+/// This layer subscribes **only** the domain-scoped ASCII CORE2 surface (mark,
+/// range start/end/push/pop, domain/register-string/name-category/resource) plus
+/// CORE `NameOsThreadA`. The classic default-domain API
+/// (`nvtxRangePushA`/`nvtxMarkA`/`nvtxRangeStartA`/`nvtxRangeStartEx`) and every
+/// wide-char (`*W` / Unicode) variant are intentionally **NOT** subscribed yet:
+/// coverage is deferred because the target is Linux-only and the v1 target
+/// libraries (libcudf, cuCascade) are domain-scoped. An app that only uses the
+/// non-domain or wide-char surface will produce an empty/partial capture; a
+/// one-time runtime diagnostic on the unsupported message path (see
+/// [`crate::convert`] `decode_message`) surfaces that gap.
```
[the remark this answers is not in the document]
9prady9: Keeping the runtime diagnostic here: at init we can't know whether an app will actually use an unsupported/wide-char surface, so a proactive init-time notice would fire for every app regardless. Part 1 (silencing the NVTX_MESSAGE_UNKNOWN no-message sentinel) makes the runtime signal accurate, which is what we can do correctly today.

integrations/nvtx/injection/src/init.rs
```diff
@@ -0,0 +1,392 @@
...
+/// Decrement this thread's open-range depth for `domain` and return the 0-based
+/// level of the range being ended (NVTX `nvtxDomainRangePop` return semantics).
+/// An unbalanced pop saturates at level `0`.
+pub(crate) fn range_pop_level(domain: u64) -> c_int {
+    RANGE_DEPTH.with(|depth| {
+        let mut map = depth.borrow_mut();
+        let level = map.entry(domain).or_insert(0);
+        *level = (*level - 1).max(0);
+        *level
+    })
```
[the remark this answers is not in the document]
9prady9: https://github.com/rapidsai/quent/pull/391/commits/51ce309af60e33962737308055f9a2ef7db4e0e6

integrations/nvtx/injection/build.rs
```diff
@@ -0,0 +1,132 @@
...
+
+    let include_dir = match std::env::var("NVTX_INCLUDE_DIR") {
+        Ok(dir) => PathBuf::from(dir),
+        Err(_) => match locate_nvtx_include() {
+            Ok(dir) => dir,
+            Err(e) => {
+                // The `regenerate-bindings` feature is switched on incidentally by
+                // any `--all-features` build/clippy (e.g. CI). Regeneration is a
+                // manual maintenance action that needs the NVTX headers (via
+                // `NVTX_INCLUDE_DIR` or the `nvidia-nvtx` git-dep). When they are
```
johanpel: If we can avoid manual regeneration and checking in generated sources, we should. I think we can pin the NVTX headers (nvtx-c) and libclang in Pixi, then have build.rs always generate the bindings into OUT_DIR. This keeps the bindings synchronized with the headers without modifying tracked sources and we wouldn't need all the feature flag machinery here.

integrations/nvtx/injection/src/convert.rs
```diff
@@ -0,0 +1,929 @@
...
+/// committed bindings, which would require libclang and rewrite the committed
+/// artifact.
+///
+/// These resource/range-id ABI types are intentionally hand-maintained
+/// and are **NOT** part of the regenerated `src/bindings.rs`. Running
+/// `build.rs` under `regenerate-bindings` will not (re)produce them — that is
+/// expected, not a sign the bindings are incomplete. Keep this module in sync
+/// with the NVTX C headers by hand on an NVTX `rev` bump.
+#[allow(nonstandard_style, dead_code)]
+pub(crate) mod abi {
```
johanpel: Related to the previous comment, but could these types also be added to the bindgen allowlist? Is there a bindgen limitation that requires these declarations to remain manual?

integrations/nvtx/injection/src/convert.rs
```diff
@@ -0,0 +1,929 @@
...
+/// hot capture path.
+fn warn_lossy_once() {
+    static WARNED: AtomicBool = AtomicBool::new(false);
+    if !WARNED.swap(true, Ordering::Relaxed) {
+        eprintln!(
+            "quent-nvtx: a non-UTF-8 NVTX string was captured lossily (invalid bytes replaced \
+             with U+FFFD); byte-verbatim capture is deferred. This warning fires once."
+        );
+    }
+}
```
johanpel, line 547: Nit: perhaps it would be better to get the client app to install a callback for this situation so they can deal with it however they please

integrations/nvtx/injection/build.rs
mbrobbel: I also prefer this approach

integrations/nvtx/injection/src/convert.rs
9prady9: No hard bindgen limitation — they'd generate fine if added to the allowlist.

They're hand-written on purpose: the allowlist is kept deliberately narrow to the injection surface we actually use, and these few extra range-id/resource types are small and stable. Hand-declaring them avoids widening the committed bindings (and re-running regeneration) just for a handful of types, and — same as the committed-bindings rationale above — keeps the crate self-contained for the eventual upstream move.

If we switch to always-generating, these would just fold into the allowlist and the manual block goes away.

integrations/nvtx/injection/build.rs
9prady9: Committing the generated bindings is deliberate — default and CI builds stay hermetic (no libclang toolchain or NVTX headers needed just to compile the crate). It also keeps the crate self-contained, which is exactly what we want since these crates are slated to be upstreamed (to nvtx-sys) and then removed from quent: a checked-in artifact lifts cleanly to the upstream home and leaves no libclang/header build dependency behind in quent's toolchain when we drop them. Generating into OUT_DIR would instead make libclang + NVTX headers a hard build requirement for every contributor, CI job, and downstream consumer.

integrations/nvtx/injection/src/convert.rs
9prady9: Agreed in principle. The eprintln! is a placeholder default; a client-installable diagnostics callback is the better shape, but it's easier to design once we have a real consumer exercising this path — the bridge's subprocess e2e / test-app in the follow-up PR (#371). Deferring this to that PR so the callback API is designed against actual usage rather than in the abstract.

integrations/nvtx/injection/build.rs
mbrobbel: Bindgen via `build.rs` is what `nvtx-sys` does, so we'll have to add those dependencies when we move to that anyways?
And not all downstream consumers will use this crate.

integrations/nvtx/injection/src/convert.rs
mbrobbel: >If we switch to always-generating, these would just fold into the allowlist and the manual block goes away.

+1 for this

integrations/nvtx/injection/build.rs
johanpel: > Committing the generated bindings is deliberate — default and CI builds stay hermetic (no libclang toolchain or NVTX headers needed just to compile the crate).

Why is this a problem when our CI builds through Pixi?

> It also keeps the crate self-contained, which is exactly what we want since these crates are slated to be upstreamed (to nvtx-sys) and then removed from quent

This is another reason why it was previously suggested to create a personal fork of NVTX.

> a checked-in artifact lifts cleanly to the upstream home and leaves no libclang/header build dependency behind in quent's toolchain when we drop them.

This seems strange, why is removing two lines of pixi toml so hard when we already have to remove hundreds of lines of generated code when we migrate?

> Generating into OUT_DIR would instead make libclang + NVTX headers a hard build requirement for every contributor, CI job, and downstream consumer.

We have exactly one supported build environment which is Pixi based.

9prady9: A bit of context on why it started as committed bindings: the goal was hermetic, reproducible builds — no libclang toolchain or NVTX header fetch needed just to compile the crate — plus keeping it self-contained for the eventual upstream move. I considered that to be a safe approach for now before the dust settles down on this feature and I have to touch pixi for this.

I personally favor the above approach. I am fine with autogen approach too.

With libclang + nvtx-c pinned in pixi, bindings always track the pinned headers, there's no checked-in generated source to drift, and it lets us drop the regenerate-bindings.

I'll switch build.rs to generate into OUT_DIR which will also fold the manual abi types into the allowlist.

I would strongly recommend against using personal forks as a dependency — for known reasons that's discouraged in open-source projects, and we already pin official NVTX and an alternative approach, so we don't need one.

mbrobbel: Please let's reconsider forking and upstreaming instead, so we can skip all this work.

9prady9: There is an active issue for upstreaming effort already https://github.com/rapidsai/quent/issues/388 I will pick that up once the phase 1 is complete while handling phase 2.

Personal forks is not a good approach, it creates a bottleneck with whoever creates the fork. I have seen this happen quite a lot to go that route again. If we handle upstreaming in parallel, this problem becomes obsolete, hopefully soon.

integrations/nvtx/injection/src/convert.rs
9prady9: should be handled in https://github.com/rapidsai/quent/pull/391/commits/7cea8d20950f82ff3f081c1ba9cb4acabe78cd0e

johanpel (approved):

PR #393, opened 2026-07-15
feat: DAG data-flow timeline — per-operator distribution protocol + playhead UI

felipeblazing:
## What

Adds a generic **data-flow-over-time** view to the query-plan DAG: a time slider (playhead) under the DAG that shows, for each operator at time T, how much data is in each lifecycle state (e.g. queued / packaged / processing) and where it resides (an application-defined dimension, e.g. memory tier), animated as you scrub.

quent stays fully application-agnostic: the protocol is *per operator, a binned timeline of a distribution over (FSM state × downstream-declared dimension) for downstream-declared measures*. All semantics (which FSM, what the dimension keys mean, units) are declared by the downstream analyzer and rendered verbatim.

## Backend

- **`crates/analyzer`**: `DistributionTimelineBuilder` — span-weighted aggregation over opaque `(series, measure, state, dimension)` keys, reusing the existing `KeyedAggregator`.
- **`crates/ui`**: `DistributionTimelineRequest` / `DistributionDecl` (dimension keys + measures with `QuantitySpec` refs, `entity_type_name` referencing the bundle's `fsm_types`) / `DistributionSeries` DTOs (ts-rs exported).
- **`domains/query_engine`**: `UiAnalyzer::data_flow_timeline` with a **default `Unsupported` impl** (existing analyzers compile unchanged; the UI hides the view), and `POST /api/engines/{id}/timeline/data-flow` (one request returns all operators of a query; scrubbing is a pure client-side bin lookup). Response shape is `combine_chunks`-compatible for future chunked caching.
- **Simulator**: full reference implementation (dimension = memory resource instance used by the state, measures = task count/bytes) + functional tests against the fixed scenario with hand-computed bins.

## UI

- **`DagPlayhead`**: plain-DOM scrubber under the DAG (drag, keyboard, play/pause) that also drives a synced crosshair on the resource timelines.
- **`NodeFlowBar`** on every DAG node: a stacked state bar over a dimension (tier) bar, scaled against the fetched window's max so scrubbing shows genuine accumulation; width-gated in-segment value labels (2–3 significant digits + unit, only when they fit); a dual-measure totals line (`121 | 5.4GB`).
- Controls: feature toggle, bar measure select, **bar-label measure select** (labels can show counts while widths show bytes, or vice versa), and **dimension-key filter chips** (e.g. show only GPU-1); selection consistently filters bars, labels, totals, scale, and the info-panel state × dimension matrix.
- Node info panel gains the full state × dimension matrix at the playhead time; legend gains state/dimension groups.
- Perf: only the tiny bar components subscribe to the per-bin frame; a scrub tick re-renders ~200 mini-bars, not 200 DAG nodes.

## Testing

- `cargo test` across the workspace (new aggregation unit tests + simulator functional tests).
- 564 vitest tests (19 files) green; typecheck, lint, production build green.
- Verified end-to-end against the simulator server and against Sirius (companion PR in sirius-db/sirius) at TPC-H SF1/SF300 on 2 GPUs.

## Notes for reviewers

- Bin values are span-weighted (time-weighted averages per bin, consistent with all existing timelines), so fractional counts are expected; the UI labels values as "during this bin".
- `docs/domains/query_engine/README.md` documents the new protocol section.
- Includes a merge of latest `main` (clean, no conflicts).


felipeblazing:
Addressed the CodeRabbit review in the two commits just pushed:

**Rust** (`refactor(query-engine): address PR #393 review feedback`):
- `domains/query_engine/ui`: `data_flow` module is now private with `DataFlowTimelineBinned`/`DataFlowTimelineResponse` re-exported from the crate root; all consumers (analyzer trait, server handler, simulator analyzer/build script, functional tests) import from the root.
- Simulator `data_flow_timeline` now declares only dimension keys actually observed in the requested query's tasks (no engine-wide metadata leakage), and the no-memory sentinel key grows (`none` → `none_` → …) until it cannot collide with a real memory resource instance name — collision-free in both the series data and the decl.

**UI** (`refactor(ui): address PR #393 review feedback`):
- `DagPlayhead` stops playback and hides the synced crosshair whenever the overlay is disabled or bin metadata disappears (it stays mounted while rendering null, so the interval previously kept ticking invisibly). Covered by a new fake-timer test asserting broadcasts stop and position is preserved on re-enable.
- `DAGChart` layout effect now guards against stale async results with a cancelled flag — only the latest `calculateLayout` invocation commits nodes/edges.
- `DAGLegend` conditional classes merged via `cn()`.
- One component per file: `DataFlowMatrix` and `ColorDot` extracted from `DAGNodeInfoPanel`, `SegmentValueLabel` extracted from `NodeFlowBar`.

**Skipped (1)**: the `@` alias import in `DagPlayhead` — the alias is defined only in the app root tsconfig (`ui/src`); `@quent/components` has no `paths` mapping and all 25+ sibling files in the package import relatively, so switching would break the package typecheck. Kept the package's dominant convention.

All suites green: `cargo test --workspace` (116 targets), 565 vitest tests, typecheck/lint/build.

examples/simulator/analyzer/src/lib.rs
```diff
@@ -774,6 +793,183 @@ impl UiAnalyzer for SimulatorUiAnalyzer {
...
+        let query_id = request.app_params.query_id;
+        let epoch = self.query_engine_model().query_epoch(query_id)?;
+        let config = request.config.try_into_binned_span(epoch)?;
+
+        // Which of the declared measures to compute; empty means all.
+        let want =
+            |name: &str| request.measures.is_empty() || request.measures.iter().any(|m| m == name);
+        let want_tasks = want(MEASURE_TASKS);
+        let want_bytes = want(MEASURE_BYTES);
+        if !want_tasks && !want_bytes {
```
johanpel: If either of these is valid we accept unknown measures, should we error instead?

docs/domains/query_engine/README.md
```diff
@@ -189,6 +189,31 @@ act as [Resource Groups][resource-group], forming a hierarchy through which
...
+- **Dimension**: an opaque, small, enumerable key set declared per response
+  (e.g. where an entity's data resides), with display names and stable order.
+- **Measures**: named weights (e.g. an entity count, resident bytes) with a
+  quantity spec reference for unit formatting.
+
+Bin values are span-weighted (an entity in a state for a fraction of a bin
+contributes that fraction), matching all other timelines. Analyzers that do
+not provide the feature return `Unsupported` — the default implementation —
+and the UI hides the view.
+
```
johanpel: The intention of this doc is to specify and explain the query engine domain event model, but not to explain the UI or any implementation details (those are not in the current docs/ at all). Let's find a better place for this or remove and create an issue so we can address it when we have some introductory sections / scaffolding in place.

crates/analyzer/src/timeline/binned/distribution.rs
```diff
@@ -0,0 +1,194 @@
...
+pub struct DistributionKey<'a, S> {
+    /// Opaque series the sample belongs to (e.g. an operator id downstream).
+    pub series: S,
+    /// The measure this weight contributes to (e.g. an entity count).
+    pub measure: &'a str,
+    /// The FSM state name during the span.
+    pub state: &'a str,
+    /// Application-defined dimension key (opaque to the aggregation).
+    pub dimension: &'a str,
+}
```
johanpel: I think it would be nice to make this more generic:

```suggestion
/// Identity of one aggregation cell of a distribution timeline.
#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub struct DistributionKey<S, M, St, D> {
    /// Opaque series the sample belongs to (e.g. an operator id downstream).
    pub series: S,
    /// The measure this weight contributes to (e.g. an entity count).
    pub measure: M,
    /// The FSM state name during the span.
    pub state: St,
    /// Application-defined dimension key (opaque to the aggregation).
    pub dimension: D,
}
```
?

This way the `DistributionTimelineBuilder` would only require `Eq + Hash` bounds and we wouldn't always need stringification.

crates/analyzer/src/timeline/binned/categorical.rs
```diff
@@ -0,0 +1,194 @@
...
+            SpanNanoSec::try_new(2000, 3000).unwrap(),
+            1.0,
+        )?;
+
+        let timeline = builder.build();
+        let bins = timeline.data.get(&key(1, "count", "a", "x")).unwrap();
+        assert_eq!(bins[..], [0.0; 10]);
+        Ok(())
+    }
+}
```
johanpel, line 171: The tests here seem a bit superfluous as they test functionality of the inner aggregator more than any functionality expressed in this source. If these tests do not exist on the aggregator yet we should move them there.

crates/analyzer/src/timeline/binned/distribution.rs
```diff
@@ -0,0 +1,194 @@
...
+//! A distribution timeline describes, per opaque series (e.g. an operator in a
+//! query engine), how some weighted quantity (a "measure", e.g. an entity
+//! count) is distributed over the states of a finite state machine and an
+//! application-defined dimension (e.g. which resource holds the entity's
+//! data), for each time bin of a window.
+//!
+//! This module is application-agnostic: series, measures, states, and
+//! dimension keys are all opaque to the aggregation. Downstream analyzers
+//! decide what they mean and are expected to keep dimension keys a small
+//! enumerable set.
```
johanpel: Nit: I was thrown off a bit about what this does because "distribution" might imply a probability distribution, but the values within each bin are absolute bin time weighted measures and not normalized to sum to 1. Perhaps we could name this "categorical" or something?

domains/query_engine/ui/src/data_flow.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Distribution series keyed by operator id.
+    pub operators: HashMap<Uuid, DistributionSeries>,
+}
+
+/// Response for a data-flow distribution timeline request.
+#[derive(TS, Debug, Clone, Serialize)]
+pub enum DataFlowTimelineResponse {
+    /// This analyzer does not provide data-flow distributions; the UI hides
+    /// the corresponding view.
+    Unsupported,
```
johanpel: Perhaps we can add an Unsupported variant to the AnalyzerError instead? That could be leveraged by other endpoints too.

johanpel (commented):
Inspected the Rust side. Thanks @felipeblazing, this is a good start and an exciting feature for the QE parts.

felipeblazing:
Thanks for the review @johanpel — all six points addressed in the latest pushes:

- **Unknown measures now always error**: any requested measure not in the declared set is rejected with `InvalidArgument` naming the offender, even alongside valid ones (simulator + test; mirrored in the Sirius implementation).
- **Docs**: removed the protocol section from the domain event-model doc; filed #407 to give analyzer/UI protocol documentation a proper home once scaffolding exists.
- **Generic key**: `CategoricalKey<S, M, St, D>` with `Eq + Hash` bounds only — no forced stringification (added a test exercising fully non-string key components).
- **Tests moved down**: the span-weighting / zero-duration / out-of-window cases now live on `KeyedAggregator` where that behavior is implemented; the categorical layer keeps only key-identity tests.
- **Naming**: renamed the whole layer "distribution" → **"categorical"** (`CategoricalTimelineBuilder`, `CategoricalTimelineRequest/Decl/Series`), with the module docs now explicit that bin values are absolute time-weighted quantities, not normalized shares.
- **`AnalyzerError::Unsupported`**: added as you suggested and reusable by any endpoint; the data-flow endpoint returns `DataFlowTimelineBinned` directly, the trait default errs `Unsupported`, the server maps it to HTTP **501**, and the UI probes on status (501 ⇒ feature hidden, no retries). The response-level `Unsupported` variant is gone.

Full suites green: `cargo test --workspace` (116 targets), fmt/clippy `--all-features --locked -D warnings`, 583 UI tests, typecheck/lint/build. The companion Sirius PR (sirius-db/sirius#1187) is updated to the new API.

johanpel (approved):
Thanks for addressing all the comments, approving Rust changes.

PR #394, opened 2026-07-15
fix(ui): add bigint display support

### LINKED ISSUE #351: bug: DAG detail panel + entities in tooltip do not handle `BigInt`
### Summary

https://github.com/rapidsai/quent/blob/main/ui/packages/%40quent/utils/src/formatters.ts#L115

Change here when fixed: 
https://github.com/rapidsai/quent/pull/323/changes#diff-44d0636bc11898c5427414456211477ebf2e926b88b11267f105b8afbf75a7edR237

Replace `StatValue` with the ts-binding's `Value` everywhere

### Version / commit

main

it('throws when passed a BigInt instead of a number', () => {
  expect(() => formatWithPrefix(1024n as unknown as number, 'B', 'Iec')).toThrow(TypeError);
});

### Expected behavior

BigInt is handled when formatting numbers/units for either DAG nodes/edges, entity data, anywhere a `ts-bindings/Value` is formatted

### Environment

- OS and architecture:
- Rust: `rustc --version`
- Cargo: `cargo --version`
- Node.js: `node --version`
- pnpm: `pnpm --version`
- Browser and version:
- Docker / Docker Compose version:
- Using `pixi shell`? yes/no

### Rust details

_No response_

### UI details

_No response_

### Screenshot or video

_No response_

### Additional logs or context

_No response_

cmatzenbach:
# Description

Previously the DAG stats panel and tooltips did not support `BigInt`, so they would be coerced into strings and not go through the formatting functions. This adds `BigInt` support to the formatting functions (and opens up pathways so `BigInt` values can access them) so that they will be treated like other numeric values.

Other changes:
- Moved `isNumericValue` into `@quent/utils` to use as a general utility function, avoiding repeated logic
- Fix pivot to include `BigInt` (coerced into numbers) rather than dropping them
- Fix DAG node/edge coloring so `BigInt` fields keep continuous coloring rather than falling back to categorical
- Added tests covering `BigInt` paths

Note: Since we are coercing `BigInts` to `numbers` in several areas, I had claude do an analysis on the safety of each of these coercions in terms of what it affects in the application. Here is a chart showing each place and its risk factor:

| Site | Purpose | Precision loss impact |
| :--- | :---: | ---: |
| dagFieldProcessing (node/edge coloring) | value normalized to (v-min)/(max-min) ∈ [0,1] for a gradient |  None — a color gradient can't show 15-digit precision anyway |
| getSortValue (pivot sort) | ordering only | Negligible — only mis-orders values within ~1 of each other above 2⁵³; cosmetic |
| pivot aggregation (sum/mean/min/max/stdev) | float math on number[]  | Approximate — sums of huge values drift in low digits, but results are displayed scaled/rounded |
| formatWithPrefix Si/Iec, formatDuration, _mbs, ratios | scaled display ("8.0 PiB", "1.5 M") | Invisible — output is ~3 sig figs, so lost low digits never appear |
| formatNumber / formatNumberWithMaxFractionDigits integer branch |  exact full-digit display with grouping | Negligable - unless a displayed stat can carry an exact value above 9 quadrillion |

## Related Issues

Closes: #351 

## Testing

A set of unit tests were added to ensure all new `BigInt` paths are verified. 

If you have a ndjson file with `BigInt` stats (I do not), that would be the best test scenario


ui/packages/@quent/utils/src/formatters.test.ts
```diff
@@ -252,6 +252,32 @@ describe('formatWithPrefix (Iec)', () => {
   });
 });
 
+// ---------------------------------------------------------------------------
+// formatWithPrefix — bigint values
+// ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
cmatzenbach: This is wrong - `test.tsx` is only used for tests that test react components _directly_, which this file does not.

ui/packages/@quent/components/src/pivot-table/buildPivotedRows.test.ts
```diff
@@ -156,6 +156,25 @@ describe('buildPivotedRows row clustering', () => {
     expect(fordHybrid?.aggs.get('price')?.count).toBe(2);
   });
 
+  it('aggregates bigint stat values (large U64/I64) without throwing', () => {
```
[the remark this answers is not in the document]
cmatzenbach: This is wrong - test.tsx is only used for tests that test react components directly, which this file does not.

ui/packages/@quent/components/src/pivot-table/utils.test.ts
```diff
@@ -161,6 +161,12 @@ describe('formatStatValue', () => {
   it('converts a string value using String()', () => {
     expect(formatStatValue('hello', 'label')).toBe('hello');
   });
+
+  it('formats a bigint value (large U64/I64 stat) via the field formatter', () => {
+    // Byte fields scale; a raw String() would show unformatted digits.
+    expect(formatStatValue(1073741824n, 'spill_bytes')).toBe('1.00 GiB');
+    expect(formatStatValue(1500n, 'output_rows')).toBe('1.50 k');
+  });
```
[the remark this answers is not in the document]
cmatzenbach: This is wrong - test.tsx is only used for tests that test react components directly, which this file does not.

johallar:
One change I want to make for this issue is to use the typescript binding for `Value` instead of defining our own `StatValue`. This should keep the UI type safe + bound to the API contract in the future

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -113,15 +113,17 @@ const IEC: readonly [number, string][] = [
...
   symbol: string,
   prefixSystem: PrefixSystem,
   decimals: number = 1
 ): string {
-  if (value === 0) return symbol ? `0 ${symbol}` : '0';
+  const num = typeof value === 'bigint' ? Number(value) : value;
 
-  const abs = value < 0 ? -value : value;
-  const sign = value < 0 ? '-' : '';
+  if (num === 0) return symbol ? `0 ${symbol}` : '0';
```
johallar, line 132: bigint 0 does not triple equal number 0 😒 `0 === 0n -> false`

loose equality or another check for `num === 0n` would fix here

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -113,15 +113,17 @@ const IEC: readonly [number, string][] = [
...
 
 export function formatWithPrefix(
-  value: number,
+  value: number | bigint,
   symbol: string,
   prefixSystem: PrefixSystem,
   decimals: number = 1
 ): string {
-  if (value === 0) return symbol ? `0 ${symbol}` : '0';
+  const num = typeof value === 'bigint' ? Number(value) : value;
```
johallar, line 130: Converting `BigInt` to `Number` is going to be a problem if above the max safe integer value

cmatzenbach:
> One change I want to make for this issue is to use the typescript binding for `Value` instead of defining our own `StatValue`. This should keep the UI type safe + bound to the API contract in the future

Agreed - went ahead and derived that from the BE-generated `DynamicValue` binding. Kept `StatValue` itself as it's a domain term, already used in multiple places (pre-PR) and includes the optional `null` value.

ui/packages/@quent/utils/src/formatters.ts
cmatzenbach: That was acknowledged and addressed that in the table in the PR description above; the precision loss doesn't affect the output here. The rounding done by `formatWithPrefix` already throws away far more precision than the `BigInt` -> `number` conversion would, so the precision loss is invisible in the final output.

The place where the precision loss _would_ matter is in `formatNumber`, which is handled differently. Instead we are passing `BigInt` to `Intl.NumberFormat` to ensure no conversion happens.

ui/packages/@quent/utils/src/formatters.ts
cmatzenbach: Probably outdated now based on your following comment, but yes, `BigInt` is coerced to ensure all values are `number` types here

ui/packages/@quent/utils/src/formatters.ts
johallar: Ah I see, I think the case where this matters most is `prefixSystem === "None"`, which would call `toFixed` on the coerced Number. If there was precision loss on the `BigInt` conversion to `Number` that would still be shown here.

Since we have the `BigInt` already, I think we want to do the scaling for units using the full precision if possible. Here's a failing unit test for the "None" case:

```
  it('preserves unprefixed bigint precision above Number.MAX_SAFE_INTEGER', () => {
    expect(formatWithPrefix(9007199254740993n, 'Hz', 'None', 0)).toBe(
      '9007199254740993 Hz'
    );
  });
```

There's one other case where losing the precision can cause it to roll into another prefix. Less concerned about this, but an edge case that could be fixed:
```
  it('normalizes a rounded mantissa into the next prefix', () => {
    expect(formatWithPrefix(999999999999999n, 'Hz', 'Si', 1)).toBe('1.0 PHz');
  });
```

ui/packages/@quent/components/src/pivot-table/utils.ts
```diff
@@ -173,7 +173,7 @@ export function getSortValue(
   if (!isAgg) {
     const v = row.values.get(stat);
     if (v === undefined) return null;
-    return isNumericValue(v) ? v : null;
+    return isNumericValue(v) ? Number(v) : null;
```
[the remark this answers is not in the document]
johallar: This too, but if you want I can go back and do the pivot table stuff

ui/packages/@quent/components/src/services/query-plan/dagFieldProcessing.ts
```diff
@@ -19,11 +19,12 @@ export function computeNodeColoring(
   });
   if (!entries.length) return null;
 
-  if (entries.every(e => typeof e.value === 'number')) {
-    const nums = entries.map(e => e.value as number);
+  // handle potential bigints
+  if (entries.every(e => isNumericValue(e.value))) {
```
johallar, line 23: This doesn't seem like an equal swap since `isNumericValue` will return true for `Number` _and_ `BigInt` now, so it would end up coercing the values on the line below.

johallar: oo nevermind, I see the table explanation now, I think that's OK

ui/packages/@quent/components/src/pivot-table/utils.ts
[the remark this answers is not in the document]
johallar: lol, comment intended for @cmatzenbach

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -329,20 +336,23 @@ function formatSiCount(value: number, decimals = 2): string {
...
  * Order: duration (ns) → bytes → row/batch counts → throughput → ratios → default table number.
  */
-export function inferFieldFormatter(fieldName: string): (value: number) => string {
-  if (fieldName.endsWith('_ns')) return v => formatDuration(v / 1e6);
-  if (isBytesStat(fieldName)) return v => formatBytes(v, 2);
-  if (isCountStat(fieldName)) return v => formatSiCount(v, 2);
-  if (fieldName.endsWith('_mbs')) return v => `${v.toFixed(1)} MB/s`;
-  if (
+export function inferFieldFormatter(fieldName: string): (value: number | bigint) => string {
+  let format: (value: number) => string;
```
johallar: This should be `number | bigint` shouldn't it?

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -329,20 +336,23 @@ function formatSiCount(value: number, decimals = 2): string {
...
+  if (fieldName.endsWith('_ns')) format = v => formatDuration(v / 1e6);
+  else if (isBytesStat(fieldName)) format = v => formatBytes(v, 2);
+  else if (isCountStat(fieldName)) format = v => formatSiCount(v, 2);
+  else if (fieldName.endsWith('_mbs')) format = v => `${v.toFixed(1)} MB/s`;
```
johallar: I think this + the ratio/fraction block below is problematic for `BigInts`, they don't have a `toFixed` method

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -329,20 +336,23 @@ function formatSiCount(value: number, decimals = 2): string {
...
     fieldName.endsWith('_fpr') ||
     fieldName.endsWith('_selectivity') ||
     fieldName.endsWith('_rate')
   )
-    return v => `${(v * 100).toFixed(1)}%`;
-  return v => formatNumberWithMaxFractionDigits(v, 4);
+    format = v => `${(v * 100).toFixed(1)}%`;
+  else format = v => formatNumberWithMaxFractionDigits(v, 4);
+
+  return value => format(typeof value === 'bigint' ? Number(value) : value);
```
johallar: This coercion is saving the formatters above from type errors, but prevents the bigint from ever reaching `formatWithPrefix`

ui/packages/@quent/utils/src/formatters.ts
cmatzenbach: Ah that's a great catch. Added a new path for `BigInt` values when `prefixSystem` is disabled so they are no longer coerced. 

Also included one other small change AI had found earlier related to the Terra prefix - when the divided mantissa rounds up to 1000, we now step up to the next prefix instead of displaying 1000.0 THz.

ui/packages/@quent/utils/src/formatters.ts
cmatzenbach: Assuming you're talking about the `format` variable, no - the `BigInt` conversion is handled at line 368 in the returned wrapper: `value => format(typeof value === 'bigint' ? Number(value) : value)`

ui/packages/@quent/utils/src/formatters.ts
cmatzenbach: These lambdas take `number` - the conversion from `BigInt` to number happens at line 368 before format is ever called. By the time any of these run, `v` is already a number, so `.toFixed()` works fine.

ui/packages/@quent/utils/src/formatters.ts
cmatzenbach: So I had looked at this before, but the formatters in this function - durations, ratios, percentages, counts - all do float arithmetic that doesn't make sense on a `BigInt` anyway. The only formatter where preserving `BigInt` precision matters is `formatBytes`, which calls `formatWithPrefix`. However, byte counts from the API are unlikely to exceed `Number.MAX_SAFE_INTEGER` in practice (that's what, 9 petabytes?), and even if they did, `formatBytes` applies IEC prefix scaling which rounds to 1 decimal place, so any precision loss in the `Number()` conversion would be invisible in the output.

johallar: This makes sense, but we are making a big assumption based on the field name as to what its probable values are at this point. The fact is that any of these `DynamicValue` type statistics can be 64 bit numbers, and it's going to depend on the engine developer who is implementing the instrumentation what these values are.

It would feel better to _really_ cover our bases and handle all the (numeric) types instead of leaning on the `Number` coercion IMO.

eg. fields with `_ns` that are nanosecond epoch timestamps are going to exceed `MAX_SAFE_INTEGER` and display incorrect values when coerced to `Number`, but also if any of the other name based bucketing is not correct we could end up in that same situation

johallar: Or actually...maybe if serde serializes these values with the type we can just have two code paths?

cmatzenbach: Ok, here's another idea that addresses some values that could be valid bigints:
```
 export function inferFieldFormatter(fieldName: string): (value: number | bigint) => string {
 return (value: number | bigint): string => {                                                                                                                                                                                                                                                      
    const num = typeof value === 'bigint' ? Number(value) : value;                                                                                                                                                                                                                                  
    if (fieldName.endsWith('_ns')) return formatDuration(num / 1e6);                                                                                                                                                                                                                                       
    if (isBytesStat(fieldName)) return formatBytes(value, 2);                                                                                                                                                                                                                                       
    if (isCountStat(fieldName)) return formatSiCount(num, 2);                                                                                                                                                                                                                                       
    if (fieldName.endsWith('_mbs')) return `${num.toFixed(1)} MB/s`;                                                                                                                                                                                                                                
    if (                                                                                                                                                                                                                                                                                            
       fieldName.endsWith('_ratio') ||                                                                                                                                                                                                                                                               
       fieldName.endsWith('_fraction') ||                                                                                                                                                                                                                                                            
       fieldName.endsWith('_fpr') ||                                                                                                                                                                                                                                                                 
       fieldName.endsWith('_selectivity') ||                                                                                                                                                                                                                                                         
       fieldName.endsWith('_rate')                                                                                                                                                                                                                                                                   
     )                                                                                                                                                                                                                                                                                               
       return `${(num * 100).toFixed(1)}%`;                                                                                                                                                                                                                                                          
    return typeof value === 'bigint' ? formatNumber(value) : formatNumberWithMaxFractionDigits(num, 4);                                                                                                                                                                                             
 };
 ```
The `inferFieldFormatter` function now handles bigints with two explicit paths rather than blindly coercing everything to Number upfront:
  - formatBytes receives the original bigint, which `formatWithPrefix` already handles natively
  - The default case uses `formatNumber(value)` for bigints, which formats via `Intl.NumberFormat`
  - All other branches (_ns, counts, ratios, _mbs) still convert to `Number` since their math is float-only

Now, for the paths do still coerce to `Number`, here's my thought - there's really a limit to how far you can go here. For the ratio/percentage branch, if a value arrives as a bigint and you multiply it by 100 and call `toFixed`, the output is almost certainly wrong semantically too - a bigint named `_ratio` is a misuse of the field naming convention, and no amount of bigint-safe math will make the display meaningful. If the engine developer names a field `_ratio` and emits a u64, the display is wrong regardless - bigint safety doesn't fix that semantic mismatch. Let me know what you think.

ui/packages/@quent/components/src/pivot-table/utils.ts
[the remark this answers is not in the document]
cmatzenbach: @johallar Ok, just pushed up a fix for this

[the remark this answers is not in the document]
johallar: Nice, I see the agg fixes, but the line that this comment is on will still affect sorting won't it?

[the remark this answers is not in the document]
cmatzenbach: Ah right, should've mentioned it - I was thinking we could address this in a follow-up PR, as that would require cascading changes to `gradientBg` and the column range comparisons in `PivotedStatsTable.tsx`. Here would be the three areas where we are currently accepting precision loss, which would require that refactor:

1. `getSortValue` line 176 - sort order for non-aggregated rows with large bigints
2. `getSortValue` agg path - agg.sum, agg.min, agg.max come from `PivotedRowAgg` which stores `number | null`, so even though we now compute them more accurately as bigints, they're converted to number before being stored and returned here
3. `gradientBg` color scaling - uses the number sort value to position a value on a color gradient, so large bigints that lost precision in `getSortValue` get slightly wrong colors

I'm not quite sure how large those changes would be, so I thought keeping it separate might make sense.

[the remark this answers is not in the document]
johallar: 👍 that works! Let's just track in an issue so we don't lose track of it

johallar (approved):

ui/packages/@quent/components/src/pivot-table/utils.ts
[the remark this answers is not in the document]
cmatzenbach: Done - logged in #475

Issue #352, opened 2026-07-10
Make schema builders amendable


johanpel:
Schema builders are currently one-shot. They take fully built children, and constraint crates cannot contribute to a schema under construction. Rework them into a tree of builders where every nested builder stays reachable and amendable until a top-level schema build().

johanpel:
Solved in #360

PR #397, opened 2026-07-16
feat(yaml): FSM entities via an fsms: block

johanpel:
WIP

# Description

Adds FSM support to the YAML DSL: a top-level `fsms:` block declares an entity's events as states, cardinality derived from the topology, lowered through a reusable `FsmEntityBuilder` in `quent-fsm`. Also adds an off-by-default `any_event` codegen option emitting an `AnyEvent` decoder for type-erased events, shown in the example.

## Related Issues

Part of #352.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2784,6 +2784,7 @@ version = "0.1.0"
 dependencies = [
  "indexmap",
  "quent-constraints",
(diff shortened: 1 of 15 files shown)


johanpel:
Going to split this out as there are some weakly related parts that are easy to split out

PR #402, opened 2026-07-17
feat(nvtx): app-integrated in-process NVTX capture

9prady9:
Wires captured NVTX events into a Quent pipeline, completing the foundational NVTX stack (#386 `quent-nvtx-events`, #391 `quent-nvtx-injection`).

### What it is

The application drives capture: it owns its Quent `Context` and exporter, annotates its code with the NVTX Rust API, and links `quent-nvtx-injection` with its `static-injection` feature so NVTX initializes injection in-process at the first NVTX call — no cdylib, no `NVTX_INJECTION64_PATH`.

- **`quent-nvtx-bridge`** — `NvtxEventEntity`, a `#[serde(transparent)]` newtype over `NvtxEvent` implementing Quent's `EntityEvent` (the orphan-rule adapter; the only crate depending on Quent internals).
- **`quent-instrumentation`** — adds `Observer::sender()`, a cloned `EventSender` so the `'static` injection hook can emit into an app-owned observer that still flushes on drop.
- **`quent-nvtx-injection`** — `static-injection` links the strong-symbol shim with `+whole-archive`, so the strong `InitializeInjectionNvtx2` overrides NVTX's weak no-op.
- **`quent-nvtx-example`** — runnable wiring plus its test.

### Using it

```rust
let observer = ctx.block_on(async { ctx.observer::<NvtxEventEntity>(opts).await })?;
let sender = observer.sender();
quent_nvtx_injection::install_hook(move |e| {
    sender.send(Event::new_now(session, NvtxEventEntity::from(e)));
})?;
// ... nvtx::mark!(...), nvtx::range!(...) ...
drop(observer); // flush
```

### Tests

`cargo test -p quent-nvtx-example` runs the example against a temp dir and asserts every core NVTX kind round-trips through ndjson. No GPU. Linux 64-bit only.


integrations/nvtx/instrumentation/c/sample_app.cpp
johanpel, line 1: Wouldn't it be simpler to just use NVTX' Rust API to produce these events in an e2e test?

integrations/nvtx/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+fn build_ndjson_observer(
+    session: Uuid,
+    dir: PathBuf,
+) -> Result<Observer<NvtxEventEntity>, Box<dyn std::error::Error>> {
+    let ctx = Context::try_new(session)?;
+    let options = FileSystemExporterOptions::new(FileSystemFormat::Ndjson, dir);
+    // `Context::observer` builds the exporter itself (via the provider) on the
+    // context's runtime, so construction errors surface here on the drain thread.
+    ctx.block_on(async { ctx.observer::<NvtxEventEntity>(options).await })
+}
```
johanpel: This creates a new Quent context. An application would typically have one Quent context configured by the client application, including the exporters. This is now hardcoded to ndjson which should typically just be used for debugging.

integrations/nvtx/instrumentation/src/lib.rs
```diff
@@ -0,0 +1,462 @@
...
+                // Without a drain the ring is never emptied and every captured
+                // event silently drop-and-counts forever, so surface it.
+                diag!("quent-nvtx: exporter unavailable; capture disabled (all events will be dropped)");
+                return;
+            };
+            drain_loop(&drain_ring, &observer, &drain_shutdown);
+            // `observer` drops here → cancels the forwarder, drains the channel,
+            // and flushes the exporter.
+        })
+        .map_err(|e| InstallError::Spawn(e.to_string()))?;
```
johanpel: I'm not quite sure why it is necessary to run a background thread with all the other machinery associated with it. Couldn't we provide the means to install callbacks executed from NVTX instrumentation so we can directly emit through a Quent Observer/Handle ?

E.g.:
```
  install_hook(move |event| {
      handle.emit(NvtxEventEntity(event));
  })
```

johanpel (commented):
Thanks for the work so far @9prady9. I have some comments which might require a significant change to the approach here, so before I proceed to review in detail:

integrations/nvtx/instrumentation/c/sample_app.cpp
9prady9: No strong reason other than fact that it is trying to mimic our most obvious users - c++ land libs like Sirius, libcudf, cucascade etc.

9prady9:
Had a quick call with @johanpel about the comment threads https://github.com/rapidsai/quent/pull/402#discussion_r3613887111 and https://github.com/rapidsai/quent/pull/402#discussion_r3613963283

I had different use case that I was trying to cover as well - app and/or library without any code changes which complicated the bridge design.

I am addressing all of the above that simplifies the use case - quent instrumentation with additional nvtx range annotations.

integrations/nvtx/example/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+            Uuid::parse_str(text)?
+        }
+        None => Uuid::now_v7(),
+    };
+
+    // The app owns the context and picks the exporter (ndjson here; a real app
+    // would use its configured pipeline).
+    let entity_dir = out_dir.join(session.to_string()).join("NvtxEvent");
+    let ctx = Context::try_new(session)?;
+    let options = FileSystemExporterOptions::new(FileSystemFormat::Ndjson, out_dir);
```
johanpel: Nit: for demo purposes it would be nice to use the callback exporter to directly debug print the events. Saves a few steps.

integrations/nvtx/example/tests/capture.rs
```diff
@@ -0,0 +1,93 @@
...
+        "RangePop",
+        "RangeStart",
+        "RangeEnd",
+    ] {
+        assert!(
+            present.contains(expected),
+            "missing captured kind {expected}; present: {present:?}"
+        );
+    }
+}
```
johanpel, line 68: Instead of invoking cargo from a test and round-tripping through files, could we please simplify this by putting the example in a lib that this test reuses, and then use the callback exporter to inspect the events? This keeps everything nicely isolated.

integrations/nvtx/example/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+publish = false
+
+[[bin]]
+name = "nvtx_capture"
+path = "src/main.rs"
+
+[dependencies]
+# `static-injection` links a strong `InitializeInjectionNvtx2` so NVTX
+# initializes injection in-process at the first NVTX call — no cdylib, no
+# NVTX_INJECTION64_PATH.
```
johanpel: This exposes some rather obscure details about the internals of how NVTX and the Rust bindings work. This may cause confusion about details users need not care about. Let's instead document "static-injection" where this matters: https://doc.rust-lang.org/cargo/reference/features.html#feature-documentation-and-discovery

crates/instrumentation/src/observer.rs
```diff
@@ -117,6 +117,15 @@ impl<T> Observer<T> {
     pub fn emit(&self, id: Uuid, event: impl Into<T>) {
         self.events_sender.emit(id, event);
     }
+
+    /// A cloned [`EventSender`] feeding this observer's pipeline.
+    ///
+    /// Lets a `'static` producer emit into the observer while the caller keeps
+    /// ownership (and still flushes on drop). The sender does not keep the
+    /// observer alive; sends after it is dropped are silently discarded.
```
johanpel: This logs through the `tracing` crate, so it's not entirely silent.

johanpel (commented):
Looks good overall, great work @9prady9. Some minor comments:

integrations/nvtx/example/Cargo.toml
9prady9: Good call — moved the `static-injection` documentation onto `nvtx-injection` itself (a `[features]` comment plus an "Attach modes" section in the crate docs) and dropped the internals from the example manifest.

integrations/nvtx/example/tests/capture.rs
9prady9: 9fde4866af290ef421deb58d69ce5d6a48cadfa7

johanpel (approved):
Nice, happy to see this integration come alive!

9prady9:
> Nice, happy to see this integration come alive!

a few more things here and there I think.

PR #404, opened 2026-07-17
ci: run zizmor directly

johanpel:
# Description

Replace the CI pre-commit wrapper with a direct pinned zizmor invocation.

🤖 Generated with Codex

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -7,12 +7,13 @@ permissions:
   contents: read
 
 jobs:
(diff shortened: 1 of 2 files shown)


mbrobbel (approved):
What is the motivation for this change?

johanpel:
> What is the motivation for this change?

I find pre-commit annoying, our CI runs quickly, and this was the only check in pre-commit, so it felt a bit weird to configure it like this.

PR #405, opened 2026-07-17
ci: bump Rust to 1.97.0 and use Pixi for dependency policy

johanpel:
- Pin Rust 1.97.0 across Docker and Pixi.
- Pin cargo-deny 0.20.2 in Pixi and run dependency-policy checks through the shared Pixi cache.
- Resolve the reported cxx advisory and yanked spin dependencies.
- Fix the Rust 1.97 Clippy failure.

🤖 Generated with Codex

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-FROM rust:1.91-trixie AS builder
+FROM rust:1.97.0-trixie AS builder
 
 WORKDIR /quent
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,7 +31,6 @@ env:
   CARGO_INCREMENTAL: 0
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-FROM rust:1.91-trixie AS builder
+FROM rust:1.97.0-trixie AS builder
```
mbrobbel, line 3: Follow-up: we could also use https://github.com/prefix-dev/pixi-docker here.

Or maybe we can remove this? Is this still used?

mbrobbel (approved):

Dockerfile
johanpel: I suspect it may be unused because it doesn't build with the `cors` feature which I think would be required for UI dev. IIRC this was once not feature-gated.
 
cc. @cmatzenbach @johallar please let us know whether we can remove this Dockerfile

johanpel: https://github.com/rapidsai/quent/issues/417

PR #406, opened 2026-07-17
feat(yaml): finite-state-machine entities via an fsms: block

johanpel:
# Description

Adds finite-state-machine entities to the YAML DSL. 

A top-level `fsms:` block declares an entity whose events represent transitions into its states, with cardinality derived from the topology and the FSM constraint validated at build time. 

Each state lists its outgoing transitions under `to:`, terminating via the reserved `exit` target.

The instrumentation-build example models a `Query` FSM and decodes its events through the `AnyEvent` callback exporter.

## Related Issues

Part of #352.


mbrobbel (approved):
Very nice. We should also write some user docs for this stuff.

johanpel:
> Very nice. We should also write some user docs for this stuff.

Absolutely: https://github.com/rapidsai/quent/issues/419

PR #425, opened 2026-07-21
fix(build-info): avoid rebuilds from Git metadata changes

johanpel:
# Description

Stop `quent-build-info` from rebuilding when no repository content changed.

Git commands and IDE integrations may refresh cached metadata in the Git index, changing its timestamp without changing staged files. Cargo interpreted that timestamp change as a reason to rerun the build script.

This change stops watching the Git index, resolves other Git metadata paths through `git rev-parse --git-path`, and ignores paths that do not exist.

## Reproduction

Before this change, run the same build twice:

```bash
pixi run cargo build -p quent-simulator-server
pixi run cargo build -p quent-simulator-server -vv 2>&1 | grep 'Dirty quent-build-info'
```

The second build reports quent-build-info as dirty and recompiles its dependent crates despite no source changes.

Written by Codex.

### DIFF
--- a/crates/build-info/src/git.rs
+++ b/crates/build-info/src/git.rs
@@ -35,6 +35,23 @@ fn run(dir: &Path, args: &[&str]) -> Option<String> {
     if value.is_empty() { None } else { Some(value) }
 }
 
+fn git_path(dir: &Path, path: &str) -> Option<std::path::PathBuf> {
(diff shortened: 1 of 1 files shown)


mbrobbel (commented):
I think we should at least follow the suggestion from 🐰: `Watch the unborn branch ref too.`

mbrobbel (approved):

PR #437, opened 2026-07-22
refactor(yaml): simplify event declarations

johanpel:
# Description

Simplifies entity event declarations. Events default to once-cardinality and use `multi: true` when repeatable. Entity events and FSM states use `attributes:`, while records retain `fields:`.

_Written by Codex._

## Related Issues

This will also make additional syntax introduced by #427 across FSM and entity blocks consistent.


mbrobbel (approved):
Nice

PR #438, opened 2026-07-22
feat(resource): restrict resource records in lists

johanpel:
# Description

Reject resource usage and bounds records inside list types. This is ambiguous, because which instance of the bounds is the truth?

Options are allowed, because I don't want to restrict users from emitting this conditionally to prevent an explosion of event declarations.

If the user desires to do something with bounds (e.g. visualize, check for leaks, etc.) then in the case of optional bounds events it is their responsibility to make sure they are conveyed. This is already the case because we can't force client code to emit certain events, so if a resource has multiple events of which only one sets bounds, but doesn't emit that event, you're basically in the same boat.

## Related Issues

Surfaced while working on #427

### DIFF
--- a/crates/resource/src/lib.rs
+++ b/crates/resource/src/lib.rs
@@ -32,15 +32,16 @@ pub use builder::{BuildError, ResourceBuilder, ResourceParts};
 /// ## Requirements
 ///
 /// 1. [`Capacity`] identifiers are unique within a resource.
-/// 2. If and only if any of the resource's capacities have a bound, the
-///    resource entity has at least one event (the "bounds event") which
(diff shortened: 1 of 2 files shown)


mbrobbel (commented):
I understand this for bounds, but why do you also want to restrict usage in a list?

johanpel:
> I understand this for bounds, but why do you also want to restrict usage in a list?

Good point, we should certainly allow a list of references with usages to multiple instances of the same type of resource entity: https://github.com/rapidsai/quent/pull/438/commits/2bdcde0fb465d30b60c2f55d6b72ae4fa9771916

mbrobbel (approved):

Issue #292, opened 2026-07-01
refactor(ui): refactor/redesign engine/query group/query selection


johallar:
The landing page/engine-query group-query selection has been around since MVP days. This flow needs a re-design, especially with features like: https://github.com/rapidsai/quent/issues/285 incoming. 

Proposals:
* Searchable/Filterable on [engine attributes](https://github.com/rapidsai/quent/issues/285), whatever those end up being. Will there also be attributes added for query groups/queries?
* If query group is made oiptional or renamed this should include those changes: 
  * https://github.com/rapidsai/quent/issues/186
  * https://github.com/rapidsai/quent/issues/78
* Table view proposed - containing custom attributes, or wherever additional engine/query group/query metadata ends up residing.

Issue #288, opened 2026-07-01
Turn required Query Engine model elements into schema constraint / metadata


johanpel:
Once #191 is finished, our in-tree Query Engine use case can be turned into something more free-form. The idea is that certain pieces of data required to be present on certain entities can be expressed through schema constraints, and we can add the necessary model capture support for it to make it really easy to express those constraints on top of the schema.

From here, application event models can be completely free form, so they dont have to adhere to the hierarchy currently imposed by this domain-specific model. A concrete example would be expressing directed acyclic graphs. It's not necessary for an "Operator" entity to exist, but we merely want to place a constraint on SOME entity that it represents a DAG node, for which it needs some attributes that declare its connectivity.

This is similar in spirit to [OTel semantic conventions](https://github.com/open-telemetry/semantic-conventions) or [Rerun's Archetypes](https://rerun.io/docs/reference/types/archetypes). The main difference is that everything is validated and determined at compile-time.

If analysis/visualization sees a schema with these constraints as metadata, then it is possible for it to analyze the event data in such a way that it can produce input for the visualization. The very code or queries to perform this analysis may also be generated from the schema (potentially at compile-time).

johanpel:
WIP on this issue has started here: #683

PR #440, opened 2026-07-23
refactor(ui, experiment): landing page redesign

johallar:
# Description
Possible landing page redesign

@johanpel This may require some BE work to fetch paginated queries, it fetches all profiles in a funky way to do this prototype.

## Related Issues
#292 

## Screenshots

https://github.com/user-attachments/assets/aecfb204-a7b1-4c68-be41-e25f2196ef21


johanpel:
Another thing to keep in mind here is that in https://github.com/rapidsai/quent/issues/288 I am going to remove query group as some kind of special scoping layer. We may already want to pretend that is just another arbitrary filter key, at least visually speaking.

docs/api/query-search.contract.ts
```diff
@@ -0,0 +1,86 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export type UUID = string;
+
+export type QueryGroupFilter =
```
johallar, line 6: Vibed up API contract for a query search landing page for discussion. Probably not worth moving this search+table style landing page forward without having something like this in place.

This optional query group mechanism seems a little overwrought... will give it a human pass

PR #446, opened 2026-07-23
refactor: hookify min zoom, use in all charts

### LINKED ISSUE #445: bug: operator timeline allows unlimited zoom
### Summary

Shift + Scroll zooming from the operator timeline chart allows users to zoom past the min bin size set my the resource timeline viz. This should respect the same limit, as should any future chart types added. 

### Version / commit

main

1. open any profile
2. Hover over operator timeline
3. Shift + scroll to zoom 
4. Keep zooming and zooming, see the relative time scale above timelines break, bins start dropping out, etc.

### Expected behavior

Zoom should stop at the predefined minimum bin size (FE limitation)

### Environment

- OS and architecture:
- Rust: `rustc --version`
- Cargo: `cargo --version`
- Node.js: `node --version`
- pnpm: `pnpm --version`
- Browser and version:
- Docker / Docker Compose version:
- Using `pixi shell`? yes/no

### Rust details

_No response_

### UI details

_No response_

### Screenshot or video

_No response_

### Additional logs or context

_No response_

johallar:
# Description
* Fixes a bug where users could zoom indefinitely from the operator timeline
* Captures horizontal scrolling as well, so operator timelines can horizontal scroll

## Related Issues
Fixes #445 

## Testing
1. Hover over timeline and shift + scroll to scroll in as far as possible
2. Expect it should stop at the minimum bin size
3. Zoom back out
4. Hover an operator timeline and shift + scroll to zoom as far as possible
5. Expect it to stop at min bin size
6. Horizontal scroll on timelines and operator timelines, should not require shift keypress.

## Screenshots

https://github.com/user-attachments/assets/a2f060de-5705-4057-b440-c3281400173e

### DIFF
--- a/ui/packages/@quent/components/src/lib/useTimelineWheelNavigation.test.ts
+++ b/ui/packages/@quent/components/src/lib/useTimelineWheelNavigation.test.ts
@@ -1,0 +1,129 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { act, renderHook } from '@testing-library/react';
+import type { EChartsInstance } from 'echarts-for-react';
+import { describe, expect, it, vi } from 'vitest';
+import { useTimelineWheelNavigation } from './useTimelineWheelNavigation';
+
+function createChart(dataZoom: { start: number; end: number }) {
+  const dom = document.createElement('div');
+  vi.spyOn(dom, 'getBoundingClientRect').mockReturnValue({
+    width: 1010,
+    height: 100,
+    top: 0,
+    right: 1010,
+    bottom: 100,
+    left: 0,
+    x: 0,
+    y: 0,
+    toJSON: () => undefined,
+  });
+
+  let dataZoomListener: (() => void) | undefined;
+  const dispatchAction = vi.fn();
+  const off = vi.fn();
+  const instance = {
+    getDom: () => dom,
+    getOption: () => ({ dataZoom: [dataZoom] }),
+    dispatchAction,
+    isDisposed: () => false,
+    on: vi.fn((event: string, listener: () => void) => {
+      if (event === 'datazoom') dataZoomListener = listener;
+    }),
+    off,
+  } as unknown as EChartsInstance;
+
+  return { instance, dom, dispatchAction, off, emitDataZoom: () => dataZoomListener?.() };
+}
+
+describe('useTimelineWheelNavigation', () => {
+  it('pans the visible range on horizontal wheel input', () => {
+    const chart = createChart({ start: 25, end: 75 });
+    const { result } = renderHook(() => useTimelineWheelNavigation(10));
+    act(() => result.current(chart.instance));
+
+    const event = new WheelEvent('wheel', {
+      deltaX: 100,
+      bubbles: true,
+      cancelable: true,
+    });
+    act(() => chart.dom.dispatchEvent(event));
+
+    expect(event.defaultPrevented).toBe(true);
+    expect(chart.dispatchAction).toHaveBeenCalledWith({
+      type: 'dataZoom',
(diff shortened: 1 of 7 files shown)


ui/packages/@quent/components/src/timeline/Timeline.tsx
```diff
@@ -359,81 +355,8 @@ export function Timeline({
...
-    // Prevent the browser from handling shift+wheel-in when ECharts can't zoom further
-    dom.addEventListener(
-      'wheel',
-      e => {
-        if (e.shiftKey && e.deltaY < 0) e.preventDefault();
-      },
-      { passive: false }
-    );
-  }, []);
+    attachWheelNavigation(instance);
```
cmatzenbach, line 358: While it seems to be fine here, we could match the other call site (from `OperatorGantChart.tsx`) and wrap this call in a `useCallback(() => {}, [attachWheelNavigation])` for consistency. Not necessary though, would be more stylistic/less confusing to other devs in the future wondering why the same wrapping is not necessary here and worrying about repeated listener registration.

ui/packages/@quent/components/src/lib/useTimelineWheelNavigation.test.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    act(() => {
+      chart.dom.dispatchEvent(
+        new WheelEvent('wheel', { deltaX: 100, bubbles: true, cancelable: true })
+      );
+    });
+
+    expect(chart.dispatchAction).not.toHaveBeenCalled();
+  });
+});
```
cmatzenbach, line 232: One test that would be nice to have would be one that covers calling `attachWheenNavigation` a second time, to ensure `cleanupRef` is working correctly and we can't accidentally register repeated listeners.

cmatzenbach (approved):
Two comments, nothing critical though. Works really well, great job solving the infinite scrolling issue.

ui/packages/@quent/components/src/timeline/Timeline.tsx
johallar: Ah ha, I think we actually don't need the callback in OperatorGanttChart and it was leftover, i'm going to go for that since it's simpler.

ui/packages/@quent/components/src/lib/useTimelineWheelNavigation.ts
```diff
@@ -0,0 +1,94 @@
...
+        const isHorizontalScroll =
+          event.deltaX !== 0 && Math.abs(event.deltaX) > Math.abs(event.deltaY);
+
+        if (event.shiftKey && !isHorizontalScroll) {
+          if (event.deltaY < 0 && isAtZoomLimit()) {
+            event.preventDefault();
+            event.stopPropagation();
+          }
+          return;
+        }
```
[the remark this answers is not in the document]
johallar: It should NOT pan instead of zoom once it hits the limit. It should pan when scrolling horizontally however

ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
```diff
@@ -325,9 +340,10 @@ export function OperatorGanttChart({
   // Join timeline-sync-group for frame-rate-level x-axis zoom sync via ECharts connect().
   // The y-axis dataZoom (index 3, when present) has a unique component ID and does not
   // propagate to resource timelines that have no matching component.
-  const onChartReady = useCallback((instance: EChartsInstance) => {
+  const onChartReady = (instance: EChartsInstance) => {
     registerAxisPointerSync(instance, 0, { receiveShowTip: false });
-  }, []);
+    attachWheelNavigation(instance, wrapperRef.current ?? undefined);
+  };
```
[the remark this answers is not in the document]
johallar: OK that's legit, it will cleanup if the whole component is unmounted from the effect callback in attachWheelNavigation, but if the echarts _instance_ is removed or replaced then it does not call the cleanup function, correct?

PR #469, opened 2026-07-27
feat(fsm): support attributed final-state transitions

johanpel:
# Description

Model FSM termination as a transition into an ordinary final state instead of through the reserved `exit` target.

A state without outgoing transitions is final, final-state transitions may carry attributes, and the initial and final states must be distinct. The resource constraint rejects final-state attributes that introduce a resource usage, ensuring usages cannot outlive the FSM.

## Related Issues

Part of #191.

_Written by Codex._


crates/fsm/src/lib.rs
```diff
@@ -94,16 +72,19 @@ impl Fsm {
...
-        std::iter::once(&self.exit_from_states.state).chain(self.exit_from_states.others.iter())
+    /// Whether `state` is a final state of this FSM.
+    ///
+    /// A final state is named by the topology and has no outgoing transition.
+    pub fn is_final_state(&self, state: &Identifier) -> bool {
+        self.states().any(|candidate| candidate == state)
+            && !self
+                .transitions
+                .iter()
+                .any(|transition| &transition.source == state)
```
dhruv9vats, line 87: Does this mean we can possibly have multiple final states?

dhruv9vats (approved):

crates/fsm/src/lib.rs
johanpel: Yes

PR #471, opened 2026-07-28
feat(ui): add quantity metadata to operator statistics

johanpel:
# Description

Add optional quantity metadata to operator statistics, referencing the quantity specifications in the query bundle.

- Preserve quantity keys when parsing statistics in the UI.
- Leave quantities unset in the generic query-engine analyzer.
- Annotate simulator byte statistics with the registered `bytes` quantity specification.
- Test registered and untyped statistic behavior.

_Written by Codex._


cmatzenbach:
@johanpel This is great, exactly what we needed. The one thing I'm seeing is all the time-related fields are getting a `quantity` value of `null`, for example:
```
probe_time_ns: {
    "value": {
        "U64": 1409839366
    },
    "quantity": null
}
```
Perhaps this is a known limitation, but just wanted to point it out. Otherwise looks good!

johanpel:
> @johanpel This is great, exactly what we needed. The one thing I'm seeing is all the time-related fields are getting a `quantity` value of `null`, for example:
> 
> ```
> probe_time_ns: {
>     "value": {
>         "U64": 1409839366
>     },
>     "quantity": null
> }
> ```
> 
> Perhaps this is a known limitation, but just wanted to point it out. Otherwise looks good!

Should be fixed in:
https://github.com/rapidsai/quent/pull/471/commits/dc68e891f78569594cfa6343f8a07262d16b23b6

johanpel:
I added this commit: https://github.com/rapidsai/quent/pull/471/commits/3d748c5af5586096208b6702b4a7183370c51abd

If the analyzer provides a value with a "seconds" quantityspec derived, it is confusing to keep the "_ns" suffix in the field name, so this commit fixes that.

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -129,6 +129,20 @@ export function OperatorTable({ queryBundle }: OperatorTableProps) {
     [entities, includedPlanIds]
   );
 
+  const statQuantitySpecs = useMemo(() => {
+    const quantitySpecs = queryBundle.quantity_specs;
+    const result: Record<string, import('@quent/utils').QuantitySpec> = {};
```
johallar: Prefer imports at the top

ui/packages/@quent/components/src/dag/DAGChart.tsx
```diff
@@ -317,6 +322,21 @@ const FlowLayout = ({
     [data.nodes]
   );
 
+  // Build a stat-key → QuantitySpec map for quantity-aware legend formatting.
+  const statQuantitySpecs = useMemo((): Record<string, QuantitySpec> => {
```
johallar: I think we could avoid assembling this object +  deep prop passing of `quantitySpecs` and `statQuantitySpecs` by using the `useQueryBundle` hook to have access to the quantitySpecs wherever in the tree. Or implementing a `useQuantitySpecs` hook that would pull it out for you.

ui/packages/@quent/components/src/dag/DAGLegend.tsx
```diff
@@ -33,10 +34,20 @@ interface ContinuousLegendProps {
...
 
-const ContinuousLegend = ({ field, min, max, palette, isDark }: ContinuousLegendProps) => {
-  const fmt = inferFieldFormatter(field);
+const ContinuousLegend = ({
+  field,
+  min,
+  max,
+  palette,
+  isDark,
+  quantitySpec,
```
johallar: This seems like a generic enough component, we could/should keep "quantity spec" idea out of it and pass a `fmt` function, and the DAG* components could deal with it.

ui/src/components/operator-table/OperatorTable.tsx
johallar: There's more of this throughout just FYI, prefer it not inline anywhere

examples/simulator/analyzer/src/lib.rs
```diff
@@ -190,7 +262,25 @@ impl UiAnalyzer for SimulatorUiAnalyzer {
...
+            .map(|operator| {
+                let mut ui_operator = operator.to_ui(epoch);
+                if let Some(statistics) = &mut ui_operator.statistics {
+                    statistics.custom_statistics =
+                        std::mem::take(&mut statistics.custom_statistics)
+                            .into_iter()
+                            .map(|(name, mut statistic)| {
+                                scale_operator_statistic(&name, &mut statistic.value);
+                                statistic.quantity =
+                                    operator_statistic_quantity(&name).map(str::to_owned);
```
9prady9: shouldn't this be gated ? on `scale_operator_statistic` returning non-`None` value ?

What would happen for stat that is `_ns` that has value type that is not `U64` ?

dhruv9vats (approved):
Rust side

9prady9 (approved):
code under `domains` folder looks good to me. Just one clarification needed for me.

examples/simulator/analyzer/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/471/commits/577648a3db65a2afbd0917b7d21ad30f5f934f4a

johanpel:
@cmatzenbach will leave dealing with the TS comments from @johallar to your expert hands, feel free to merge after fixing those

ui/src/components/operator-table/OperatorTable.tsx
cmatzenbach: Yeah that was stupid, no idea why claude did that. Went and revised

ui/packages/@quent/components/src/dag/DAGChart.tsx
cmatzenbach: Went ahead and replaced this with two atoms -  `quantitySpecsAtom` (raw quantity_specs from the bundle) and `statQuantitySpecsAtom` (stat-key `QuantitySpec` derived from DAG node data). Now `QueryPlan` sets `quantitySpecsAtom` via a `useEffect` when the bundle loads, and `FlowLayout` sets `statQuantitySpecsAtom` after processing nodes. All components that need the data now just read from the atoms.

ui/packages/@quent/components/src/dag/DAGLegend.tsx
cmatzenbach: Went ahead and removed that to keep it generic - now just takes an optional `formatValue` function and falls back to `inferFieldFormatter`

ui/packages/@quent/components/src/dag/DAGChart.tsx
```diff
@@ -317,6 +318,24 @@ const FlowLayout = ({
     [data.nodes]
   );
 
+  const quantitySpecs = useQuantitySpecs();
```
johallar: I don't think we need a new atom for `quantitySpecs` since that would duplicating state that tanstack is already caching for us from the query bundle. 

Instead of storing this in jotai as well, and keeping up to date w react effects, `useQuantitySpecs` could just pull this from the query bundle directly (`useQueryBundle`)

ui/packages/@quent/components/src/dag/DAGChart.tsx
```diff
@@ -317,6 +318,24 @@ const FlowLayout = ({
...
+  const setStatQuantitySpecs = useSetStatQuantitySpecs();
+
+  // Derive and publish a stat-key → QuantitySpec map for quantity-aware legend formatting.
+  useEffect(() => {
+    if (!quantitySpecs) return;
```
johallar: If `stat` already has a `quantity` field, can we avoid assembling this object + state, and just look that up from `quantitySpecs` in the query bundle?

eg. 

1.  `quantitySpecs` object goes away
2. Update `DagNodeInfoPanel`

```
const {quantity_specs} = useQueryBundle(engineId, queryId);  // Extract from bundle however, this probably isn't right
....
{selectedNodeData.statistics?.map(({ key, value, quantity }) => {
              const quantitySpec = quantitySpecs[quantity]; // Not sure this is exactly right
              return <div key={key} className="text-xs">
                {Array.isArray(value) ? (
                  <div className="flex items-center justify-between gap-0.5">
                    <DataText className="capitalize">{key.replace(/_/g, ' ')}:</DataText>
                    <div className="ml-2 flex flex-col gap-0.5">
                      {value.map((item, i) => (
                        <DataText key={i} className="text-muted-foreground whitespace-pre-line">
                          {item}
                        </DataText>
                      ))}
                    </div>
                  </div>
                ) : (
                  <div className="flex items-center justify-between">
                    <DataText className="capitalize">{key.replace(/_/g, ' ')}:</DataText>
                    <DataText className="text-muted-foreground ml-1">
                      {typeof value === 'number'
                        ? formatStatWithQuantity(value, key, quantitySpec)
                        : String(value)}
                    </DataText>
                  </div>
                )}
...
```

And `formatStatWithQuantity` becomes simpler

```
export function formatStatWithQuantity(
  value: number,
  key: string,
  quantitySpec: QuantitySpec | undefined
): string {
  if (quantitySpec) {
    return formatQuantity(value, quantitySpec, 'Occupancy');
  }
  return inferFieldFormatter(key)(value);
}
```

ui/packages/@quent/utils/src/formatters.ts
```diff
@@ -359,3 +359,24 @@ export function formatQuantity(
...
+ */
+export function formatStatWithQuantity(
+  value: number,
+  key: string,
+  quantity: string | undefined,
+  specs: { [key: string]: QuantitySpec | undefined } | undefined
+): string {
+  if (quantity && specs) {
+    const spec = specs[quantity];
+    if (spec) return formatQuantity(value, spec, 'Occupancy');
```
johallar: Are we sure that all of the custom_statistics will all be 'Occupancy' kinds and not 'Rate'? Is this actually something we may need from the BE in addition to the capacity key @johanpel ? Or does a Rate kind not make sense in this context.

ui/packages/@quent/components/src/pivot-table/PivotedStatTable.tsx
```diff
@@ -180,6 +180,7 @@ function DataCell({ row, stat }: DataCellProps<PivotedRow>) {
     onMouseEnter: () => interaction.setHoveredStat(derived.buildHoveredStatInfo(stat)),
     onMouseLeave: () => interaction.setHoveredStat(null),
   };
+  const quantitySpec = display.statQuantitySpecs?.[stat];
```
johallar: `PivotedStatTable` should not know about quantity specs, no BE generated TS bindings allowed, would prefer pattern similar to above with a `formatValue` prop to lift this responsibility to the calling component. 

And then if we can just use the stat.quantity + queryBundle quantity specs instead of assembling the `statQuantitySpecs` object that would be preferable as well.

ui/packages/@quent/utils/src/formatters.ts
cmatzenbach: This was something I had looked into - currently all the ones we are receiving are of the `Occupancy` type. However, if we were to receive 'Rate' types, then yes, we would need this from the BE as well, as we have no way of determining in the FE. But I wasn't sure myself if 'Rate' types made any sense here.

ui/packages/@quent/components/src/dag/DAGLegend.tsx
```diff
@@ -144,13 +152,7 @@ function NodeLegendContent({
...
   if (coloring.type === 'continuous') {
     return (
-      <ContinuousLegend
-        field={field}
-        min={coloring.min}
-        max={coloring.max}
-        palette={palette}
-        isDark={isDark}
-      />
+      <ContinuousLegend field={field} min={coloring.min} max={coloring.max} palette={palette} isDark={isDark} />
```
johallar: Do we need to pass `(v: number) => formatQuantity(v, quantitySpec, 'Occupancy')` for the format prop now after moving to that pattern? I think it will awlays fall back to inferring the formatter RN.

ui/src/components/QueryPlan.tsx
```diff
@@ -214,7 +217,7 @@ export function QueryPlan({ queryId, engineId }: { queryId: string; engineId: st
               </Suspense>
             </div>
             <DagPlayhead startTimeUnixNs={queryBundle.start_time_unix_ns} />
-            <DAGNodeInfoPanel isDark={isDark} />
+            <DAGNodeInfoPanel isDark={isDark} quantitySpecs={queryBundle.quantity_specs} />
```
johallar, line 244: nit: `DAGNodeInfo` could get this itself from `useQueryBundle`

ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
```diff
@@ -95,7 +101,13 @@ export const DAGNodeInfoPanel = ({ isDark = false }: { isDark?: boolean }) => {
                   <div className="flex items-center justify-between">
                     <DataText className="capitalize">{key.replace(/_/g, ' ')}:</DataText>
                     <DataText className="text-muted-foreground ml-1">
-                      {typeof value === 'number' ? inferFieldFormatter(key)(value) : String(value)}
+                      {typeof value === 'number'
+                        ? formatStatWithQuantity(
+                            value,
+                            key,
+                            quantity && quantitySpecs ? quantitySpecs[quantity] : undefined
```
johallar, line 108: golf: `quantity ? quantitySpecs?.[quantity] : undefined`

johallar (approved):
One substantive comment: https://github.com/rapidsai/quent/pull/471/changes#r3692590810

Couple other style things, feel free to merge after the format prop comment is addressed

ui/packages/@quent/components/src/dag/DAGLegend.tsx
cmatzenbach: Nope you're right, that's pointless after the big refactor. Removed

ui/src/components/QueryPlan.tsx
cmatzenbach: I don't think it's worth it - we would need to pass down two additional props (`queryId` and `engineId`) for it to be able to do so - I think it's better to just pass down the one prop it needs for rendering

johallar: You can grab those from the router, since they're in the FE route 😄 `const { engineId } = Route.useParams();` I believe

ui/packages/@quent/components/src/dag/DAGLegend.tsx
johallar: I think we need to _add_ `(v: number) => formatQuantity(v, quantitySpec, 'Occupancy')` as the `formatValue` prop don't we? Right now I think it will always fall back to the [infer formatter path](https://github.com/rapidsai/quent/pull/471/changes/BASE..385de43b9f413410e5260348b29477e882357824#diff-30aac349e10ae590b101a1f9273a0b828ae653d0c5263934f77616fed5bfcc1cR47)

ui/src/components/QueryPlan.tsx
johallar: The joys of a type safe router

ui/packages/@quent/components/src/dag/DAGLegend.tsx
johallar: Is this one resolved? Don't we need to pass the `formatValue` prop through to take advantage of the capacity or am I misinterpreting

PR #473, opened 2026-07-28
feat(nvtx): tolerant NVTX model and analyzer

### LINKED ISSUE #372: Phase 2: Reconstruct NVTX Ranges (Model & Tolerant Analyzer)
**What this delivers:** turning the raw captured events into labeled, structured NVTX ranges — reliably, even when the incoming data is messy.

**In scope**
- Rebuild nested ranges per thread, and match start/end pairs that cross threads
- Resolve the human-readable names (registered strings, domains, categories) with correct grouping
- Tolerate malformed input — unclosed ranges, out-of-order and duplicate-timestamp events — without ever crashing
- Compute per-range statistics (count, total/avg/min/max duration)

**Out of scope (later phases)**
- Any HTTP endpoint or UI
- Decoding payload/extension data
- Running alongside Nsight/other tools

**Deliverables**
- An NVTX domain model (ranges, marks, domains, threads, categories)
- A tolerant analyzer that reconstructs ranges from an event stream
- Malformed-stream test fixtures

**Done when**
- [ ] Nested ranges reconstruct correctly per thread; cross-thread start/end pairs match
- [ ] String/domain/category names resolve, with stable placeholders when a name is missing
- [ ] A stream with unclosed, out-of-order, and duplicate events analyzes to completion with no crash; still-open ranges are closed at the end
- [ ] Per-range statistics are computed
- [ ] Marks, domains, threads, and categories are represented in the model

> Note: the shared analyzer today **panics** on malformed/out-of-order events. Making reconstruction tolerant is a prerequisite here, not optional polish.

---
Part of #76 · Phase 2 of 5

9prady9:
## Summary

Closes #372. Adds `integrations/nvtx/analyzer` — a framework-free reconstruction core turning a captured `Event<NvtxEventEntity>` stream into a labeled `NvtxModel`. No dependency on any legacy macros.

**Two passes.** The first orders by timestamp and learns every handle registration; the second replays against those tables. By then neither arrival order nor forward references matter.

**Four match keys**, each chosen because the alternatives fail silently rather than loudly:

| Construct | Key | Why not coarser |
|---|---|---|
| `RangeStart`/`RangeEnd` | `range_id` alone | Ids come from one process-global counter, so they match across threads and domains; the domain on an end is redundant |
| `RangePush`/`RangePop` | stack per `(thread_id, domain)` | A global stack closes another thread's push and reconstructs believable-but-wrong nesting |
| Resource create/destroy | bare handle | The destroy carries no domain; keying on the pair makes every resource reconstruct as a leak |
| Registered strings, category names | per domain | Handle values and small category ids collide across domains |

Kind-conditional fields live in `SpanKind` variant data — thread and parent on `PushPop`, identifier type on `Resource` — so a start/end range with a thread is unrepresentable rather than merely undocumented.

**Tolerance.** Nothing is substituted for what the stream never said. An open with no close keeps everything the open stated and ends `None`, as does one whose key was reused while still open. A close with no open yields no span — the closing events carry only a correlation key, with no name or attributes to build one from. Both that and the reuse are counted in `ReconstructionAnomalies`. Drains sort for determinism, arithmetic is saturating or checked throughout, and unresolved handles get placeholders that are pure functions of the raw id, so they cannot pass as real names.

**Statistics.** `RangeStats` per `(name, domain, category)` — count and total/avg/min/max duration. Marks have no duration and resource lifespans measure existence rather than work, so neither participates; a span whose close was never captured raises `count` without raising `observed_count`.

## Test plan

- [x] `cargo test -p nvtx-analyzer` — 33 tests, hermetic
- [x] `pixi run cargo test -p nvtx-analyzer --features real-capture-tests` — adds the roundtrip over a real in-process capture (needs pixi for the nvtx-c headers)
- [x] `pixi run cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- [x] `pixi run cargo fmt --all -- --check`

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/integrations/nvtx/analyzer/src/lib.rs
+++ b/integrations/nvtx/analyzer/src/lib.rs
@@ -1,0 +1,43 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Hand-written, framework-free reconstruction core for captured NVTX events.
+//!
+//! Turns a stream of verbatim [`NvtxEvent`](nvtx_events::NvtxEvent)s — carried in
+//! Quent's [`Event`](quent_events::Event) envelope — into an in-memory
+//! [`NvtxModel`] of plain [`NvtxSpan`]s.
+//!
+//! A capture is a partial observation: it watches a process that was already
+//! running and keeps running afterwards. Incomplete pairs are therefore ordinary
+//! input, not errors — and this layer resolves what the stream referenced
+//! without substituting anything for what it never said. An open with no close
+//! keeps everything the open stated and has [`end`](NvtxSpan::end) `None`, so
+//! what to bound it at is the consuming analysis's decision rather than one made
+//! here. A close with no open produces no span at all, because the closing
+//! events carry only a correlation key — no name, no attributes — and is counted
+//! in [`ReconstructionAnomalies`] instead.
+//!
+//! Events may arrive out of timestamp order, and a label's registration may
+//! arrive after the range using it or never; replay sorts first and resolves
+//! names in a prior pass. See [`NvtxModelBuilder::build`].
+//!
+//! The crate defines its own span type and depends on neither the shared
+//! analyzer nor the shared model crate, which is what lets the above states be
+//! representable at all.
+
+mod anomalies;
+mod model;
+mod ranges;
+mod resource;
+mod span;
+mod stats;
+mod tables;
+
+pub use anomalies::ReconstructionAnomalies;
+pub use model::{NvtxModel, NvtxModelBuilder};
+pub use span::{NvtxCategory, NvtxDomain, NvtxMark, NvtxSpan, NvtxThread, SpanId, SpanKind};
+pub use stats::{RangeStats, StatsKey};
+
+// Re-exported so consumers can read span attributes without depending on the
+// vocabulary crate directly. Carried verbatim, exactly as captured.
+pub use nvtx_events::{NvtxColor, NvtxPayload};
--- a/integrations/nvtx/analyzer/src/model.rs
+++ b/integrations/nvtx/analyzer/src/model.rs
@@ -1,0 +1,332 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! The two-pass reconstruction entry point.
+//!
+//! Pass 1 materializes the stream in timestamp order and builds the
+//! handle-resolution tables; pass 2 replays it against them. Splitting the two
+//! is what makes replay tolerant of arrival order in both dimensions — by then
+//! neither an end before its start nor a `RegisterString` after the range using
+//! it is still possible.
+
+use std::collections::BTreeMap;
+
+use nvtx_bridge::NvtxEventEntity;
+use nvtx_events::NvtxEvent;
+use quent_events::Event;
+use quent_time::{TimeOrderedCollector, TimeUnixNanoSec};
+
+use crate::anomalies::ReconstructionAnomalies;
+use crate::ranges::{PushPopRanges, StartEndRanges};
+use crate::resource::Resources;
+use crate::span::{
+    NvtxCategory, NvtxDomain, NvtxMark, NvtxSpan, NvtxThread, SpanId, SpanKind, category_id,
+};
+use crate::stats::{self, RangeStats, StatsKey};
+use crate::tables::ResolutionTables;
+
+/// An in-memory model reconstructed from a captured NVTX event stream.
+#[derive(Debug, Default)]
+pub struct NvtxModel {
+    spans: Vec<NvtxSpan>,
+    marks: Vec<NvtxMark>,
+    domains: Vec<NvtxDomain>,
+    threads: Vec<NvtxThread>,
+    categories: Vec<NvtxCategory>,
+    /// Retained so names not attached to a reconstructed entity — an unnamed
+    /// thread, a category referenced by nothing yet — still resolve on demand.
+    tables: ResolutionTables,
+    anomalies: ReconstructionAnomalies,
+    /// Event timestamp bounds. Not recoverable from the spans — see the
+    /// accessors.
+    trace_start: TimeUnixNanoSec,
+    trace_end: TimeUnixNanoSec,
+}
+
+impl NvtxModel {
+    /// Every reconstructed span.
+    ///
+    /// A span's index here *is* its [`SpanId`], which is what makes
+    /// [`SpanKind::PushPop::parent`](SpanKind::PushPop) resolvable. Ordering is
+    /// by opening: push/pop ranges take their slot when pushed, start/end ranges
+    /// when they close.
+    pub fn spans(&self) -> &[NvtxSpan] {
+        &self.spans
+    }
+
+    /// The span a [`SpanId`] refers to, or `None` if it came from another model.
+    pub fn span(&self, id: SpanId) -> Option<&NvtxSpan> {
+        self.spans.get(id.0)
+    }
+
+    /// What the stream did that the span list does not show — dropped closes,
+    /// and keys reused while still open.
+    ///
+    /// Check [`is_faithful`](ReconstructionAnomalies::is_faithful) before
+    /// treating [`spans`](Self::spans) as the whole population: attaching to a
+    /// process with work already running truncates the head of the trace, and a
+    /// dropped close leaves nothing behind to notice it by.
+    pub fn anomalies(&self) -> ReconstructionAnomalies {
+        self.anomalies
+    }
+
+    /// The earliest event timestamp, or `0` for an empty stream.
+    ///
+    /// Not derivable from [`Self::spans`]: the first event may be a mark, or a
+    /// close whose open was never captured.
+    pub fn trace_start(&self) -> TimeUnixNanoSec {
+        self.trace_start
+    }
+
+    /// The latest event timestamp, or `0` for an empty stream.
+    ///
+    /// Where observation stopped — the bound to reach for when an analysis needs
+    /// to give a span with no [`end`](NvtxSpan::end) a right edge. Reconstruction
+    /// does not apply it for you, because whether that bound belongs in a figure
+    /// depends on the figure. The largest span `end` is not a substitute: the
+    /// last event may carry no span at all.
+    pub fn trace_end(&self) -> TimeUnixNanoSec {
+        self.trace_end
+    }
+
+    /// Every reconstructed mark, in timestamp order.
+    pub fn marks(&self) -> &[NvtxMark] {
+        &self.marks
+    }
+
+    /// Every reconstructed resource lifespan: the interval between a create and
+    /// its destroy.
+    ///
+    /// NVTX says only that the handle existed and what it was called, so nothing
+    /// about its size or occupancy is inferred.
+    pub fn resources(&self) -> impl Iterator<Item = &NvtxSpan> {
+        self.spans
+            .iter()
+            .filter(|span| matches!(span.kind, SpanKind::Resource { .. }))
+    }
+
+    /// Every domain the stream mentioned, ordered by raw handle.
+    pub fn domains(&self) -> &[NvtxDomain] {
+        &self.domains
+    }
+
+    /// Every OS thread the stream mentioned, ordered by id.
+    pub fn threads(&self) -> &[NvtxThread] {
+        &self.threads
+    }
+
+    /// Every non-zero category the stream mentioned, ordered by `(domain, id)`.
+    pub fn categories(&self) -> &[NvtxCategory] {
+        &self.categories
+    }
+
+    /// Aggregated durations per `(name, domain, category)`.
+    ///
+    /// Range spans only. Marks have no duration and resource lifespans measure
+    /// existence rather than work, so neither participates.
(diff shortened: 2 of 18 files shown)


integrations/nvtx/analyzer/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+
+//! Hand-written, framework-free reconstruction core for captured NVTX events.
+//!
+//! Turns a stream of verbatim [`NvtxEvent`](nvtx_events::NvtxEvent)s — carried in
+//! Quent's [`Event`](quent_events::Event) envelope — into an in-memory
+//! [`NvtxModel`] of plain [`NvtxSpan`]s.
+//!
+//! The core is deliberately **off** the shared analysis framework: it depends on
+//! neither the shared analyzer nor the shared model crate and uses none of the
+//! `model!` / `fsm!` / `entity!` macro DSL, defining its own span type instead.
```
johanpel: Nit: let's document what it is, not what it is not as it is not many things

integrations/nvtx/analyzer/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+//! That independence is what makes it **tolerant by construction** — foreign
+//! telemetry is untrusted and frequently malformed, so:
```
johanpel: Not entirely sure what "untrusted and frequently malformed" implies here, could you elaborate?

integrations/nvtx/analyzer/src/model.rs
```diff
@@ -0,0 +1,282 @@
...
+    /// Tolerant by construction: out-of-order events are reordered, forward
+    /// references to not-yet-registered handles resolve anyway, duplicate
+    /// timestamps reconstruct deterministically, ranges left open at the end of
+    /// the stream are closed and flagged, and closes with no matching open are
+    /// logged and skipped. Handles that never resolve get a stable placeholder.
+    /// No malformed stream aborts the build or panics.
+    ///
+    /// Infallible by signature: there is no anomaly a caller could be asked to
+    /// handle. `build` takes events already decoded, so a decode failure
+    /// belongs to whichever crate does the decoding.
```
johanpel: Nit: since the signature is clearly visible, this comment is a bit noisy

integrations/nvtx/analyzer/src/model.rs
```diff
@@ -0,0 +1,282 @@
...
+    pub fn build(events: impl IntoIterator<Item = Event<NvtxEventEntity>>) -> NvtxModel {
+        // Pass 1a — materialize in timestamp order. Equal timestamps keep
+        // arrival order, so replay is deterministic.
+        let mut collector = TimeOrderedCollector::default();
```
johanpel, line 182: Just a thought for follow-up: we could perhaps turn TimeOrderedCollector in an iterator that just iterates over its input without doing anything but checking whether the input is ordered. It would only lazily reorder if it encounters an unordered input, so best-case it never collects if the input iterator is already time ordered.

integrations/nvtx/analyzer/src/model.rs
```diff
@@ -0,0 +1,282 @@
...
+                } => {
+                    let name = tables.resolve_message(domain, &attributes.message);
+                    if let Some(span) =
+                        ranges.start(range_id, domain, name, attributes, event.timestamp)
+                    {
+                        slots.push(Some(span));
+                    }
+                }
+                // The domain on a `RangeEnd` is redundant: `range_id` is
+                // process-globally unique, so it alone identifies the range.
```
johanpel, line 225: One thought about process-global uniqueness. Quent is designed to work for distributed systems too. If multiple processes are using NVTX, we need to ensure all NVTX data is properly attributed to that scope. This could be solved at a higher layer than what `NvtxModel` will provide so this PR doesn't necessarily need changes right now, but it is food for thought.

For our immediate purpose of instrumenting query engines there's already a "worker" entity that is supposed to represent a "process". One thought here is to represent the entire injection library as an "entity" so it gets its own unique id, and it would have two events:
- one once-event with an entity reference to the entity representing the process (e.g. in the query engine case the worker entity).
- one multi-event to push nvtx events

This way in QE analysis we can construct one `NvtxModel` per "worker".

Opened this issue: https://github.com/rapidsai/quent/issues/522 to address external sources referencing OS id-ed things in general

integrations/nvtx/analyzer/src/ranges.rs
```diff
@@ -0,0 +1,289 @@
...
+    fn close(self, end: TimeUnixNanoSec, parent: Option<SpanId>, synthetic_end: bool) -> NvtxSpan {
+        NvtxSpan {
+            domain: self.domain,
+            thread_id: self.thread_id,
+            name: self.name,
+            category: self.category,
+            color: self.color,
+            payload: self.payload,
+            start: self.start,
+            end: end.max(self.start),
```
johanpel: Seems like an error worth raising instead of silently patching event data. What would e.g. happen if we calculated averages based on this span?

integrations/nvtx/analyzer/src/ranges.rs
```diff
@@ -0,0 +1,289 @@
...
+pub(crate) struct StartEndRanges {
+    open: HashMap<u64, OpenRange>,
+}
+
+impl StartEndRanges {
+    /// Record a `RangeStart` under its already-resolved `name`.
+    ///
+    /// Returns a span when this start *displaces* one still open under the same
+    /// id. That reuse is malformed, but the displaced start was observed, so it
+    /// is closed at the restart and flagged synthetic rather than erased.
```
johanpel: Can we sensibly perform analysis if the NVTX event stream is malformed? I think we should rather bubble the error up loudly, since we might derive wildly erroneous conclusions from the analysis.

integrations/nvtx/analyzer/src/ranges.rs
```diff
@@ -0,0 +1,289 @@
...
+            debug!("orphan nvtx range end for id 0x{range_id:X} with no open start; skipping");
+            return None;
+        };
+        Some(open.close(end, None, false))
+    }
+
+    /// Close every range still open at the end of the trace.
+    ///
+    /// Ordered by `(start, range_id)` so several leaked ranges still
+    /// reconstruct deterministically out of an unordered `HashMap`.
```
johanpel, line 130: https://github.com/rapidsai/quent/pull/473/changes#diff-dc4ca8856c582aaaf496ae4f63faea203bcc48e6f669beab5903dac2805195c5R121-R122 

implies we might attach/detach mid-flight of NVTX, but there we drop events that have no start, which feels a bit inconsistent.

Perhaps it would be best to make this layer purely focused on resolving internal references where possible. We could simply let timestamps be `Option`s and let downstream analysis deal with missing data however it pleases. It might be sensitive to the algorithm too (e.g. calculating averages vs. counting certain events for a histogram).

integrations/nvtx/analyzer/src/ranges.rs
```diff
@@ -0,0 +1,289 @@
...
+        trace_end: TimeUnixNanoSec,
+    ) -> Vec<(SpanId, NvtxSpan)> {
+        let mut leaked: Vec<(StackKey, Vec<OpenPushSpan>)> = self.stacks.drain().collect();
+        leaked.sort_by_key(|(key, _)| *key);
+
+        let unpopped: usize = leaked.iter().map(|(_, stack)| stack.len()).sum();
+        if unpopped > 0 {
+            warn!(
+                "{unpopped} nvtx range(s) were never popped; closing them at trace end ({trace_end})"
+            );
```
johanpel: Same as above, instead of patching the timestamp data in ways that could make no sense for certain analyses, we could just focus on resolving internal references and let downstream analysis layers deal with missing elements.

integrations/nvtx/analyzer/src/resource.rs
```diff
@@ -0,0 +1,165 @@
...
+        leaked.sort_by_key(|(handle, open)| (open.start, *handle));
+
+        // One `warn!` for the count, per-handle detail at `debug!`, so a stream
+        // leaking many resources does not emit one warning each.
+        if !leaked.is_empty() {
+            warn!(
+                "{} nvtx resource(s) were never destroyed; closing them at trace end ({trace_end})",
+                leaked.len()
+            );
+        }
```
johanpel, line 129: Same here

integrations/nvtx/analyzer/src/span.rs
```diff
@@ -0,0 +1,149 @@
...
+    ///
+    /// `None` for every other kind. A core/generic NVTX resource type gets a
+    /// static label; an unrecognized or CUDA-extension type passes through as
+    /// `"<identifier_type {n}>"` rather than being guessed at.
+    pub identifier_type_label: Option<String>,
+    /// The enclosing span, for nested push/pop ranges.
+    pub parent: Option<SpanId>,
+    /// `true` when the close was never observed and was synthesized at trace end.
+    pub synthetic_end: bool,
+}
```
johanpel, line 157: It would be more idiomatic to fold these fields into the `SpanKind` enum as variant data and then make `SpanKind` the `NvtxSpan`.

integrations/nvtx/analyzer/src/span.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// Saturating, so a malformed stream can never underflow it.
+    pub fn duration(&self) -> u64 {
+        self.end.saturating_sub(self.start)
+    }
+}
+
+/// A reconstructed `nvtxDomainMarkEx` instant.
+///
+/// Deliberately *not* a zero-length [`NvtxSpan`]: a consumer rendering a
+/// timeline needs to tell "happened at" from "lasted zero".
```
johanpel: Nit: superfluous comment, every distinct type is (I hope) deliberate and not some other type

integrations/nvtx/analyzer/src/stats.rs
```diff
@@ -0,0 +1,122 @@
...
+            self.min_duration = duration;
+            self.max_duration = duration;
+        } else {
+            self.min_duration = self.min_duration.min(duration);
+            self.max_duration = self.max_duration.max(duration);
+        }
+
+        self.count += 1;
+        // Saturating: a saturated total is a visibly wrong number rather than a
+        // silently wrapped plausible one.
```
johanpel: Perhaps we can keep a bool in the output to flag values as saturated

integrations/nvtx/analyzer/src/tables.rs
```diff
@@ -0,0 +1,310 @@
...
+//!   registers strings per domain, so one handle value names different strings
+//!   in different domains;
+//! - category names key on `(domain, category)`, never globally — small ids like
+//!   `1` collide across domains constantly.
+//!
+//! Anything unresolved gets a placeholder that is a pure function of the raw id,
+//! so the same stream always renders the same labels and an unresolved name is
+//! visibly bracketed rather than passing as a real one.
+
+use std::collections::{BTreeSet, HashMap};
```
johanpel: Nit: for perf we could leverage the `rustc-hash` maps already in the workspace

johanpel (commented):
Thanks @9prady9, this is shaping up nicely. Reviewed the non-test parts to begin with:

integrations/nvtx/analyzer/src/ranges.rs
9prady9: Done — `NvtxSpan::end` is now `Option`, so an open whose close never arrived reconstructs with `end: None` and nothing is synthesized in either direction. Orphan closes stay counters rather than spans because `RangeEnd`/`RangePop`/`ResourceDestroy` carry no message or attributes — a span built from one would have no name, category, colour or payload. `trace_end()` is on the model for a consumer that wants a right edge.

integrations/nvtx/analyzer/src/ranges.rs
9prady9: Done — now `drain_unclosed`, no timestamp argument, `end: None`.

integrations/nvtx/analyzer/src/resource.rs
9prady9: Same — `drain_unclosed`, `end: None`. A handle recreated while still open now also ends `None`, which reads like any other missing close, so the reuse is counted in `ReconstructionAnomalies::reused_resource_handles` (and `reused_range_ids` for the range equivalent).

integrations/nvtx/analyzer/src/span.rs
9prady9: Done — `PushPop { thread_id, parent }`, `StartEnd`, `Resource { identifier_type }`, with `identifier_type_label()` computed on demand so a resource span no longer allocates a `String`; `SpanKind` stays `Copy`. I kept `domain`/`name`/`start`/`end` on `NvtxSpan` rather than making `SpanKind` *be* the span — they are common to all three, and per-variant copies would push a match into `range_statistics`' grouping key and every filter. `category`/`color`/`payload` are always `None` for `Resource` too, if you want those folded as well.

integrations/nvtx/analyzer/src/model.rs
9prady9: Filed as #541. Worth noting it would not help this crate: the collector already has an O(1) in-order fast path and never sorts, and `build` must materialize regardless because pass 1b hands the ordered slice to `ResolutionTables::build` before replay.

integrations/nvtx/analyzer/src/model.rs
9prady9: Agreed, leaving as-is — one `NvtxModel` covers one process's id spaces, so a distributed capture builds one per worker. Thanks for filing #522.

johanpel (approved):
Changes look great, nice work @9prady9! Can't wait to see this in action.

PR #474, opened 2026-07-28
feat(nvtx): add thread_id to RangePush/RangePop and stamp OS thread id in callbacks

9prady9:
Prerequisite for the NVTX tolerant analyzer (#473 ).

Adds `thread_id: u32` to the `RangePush` and `RangePop` variants in `nvtx-events` so the analyzer can reconstruct per-thread nested stacks. Without a stamped thread id, Push/Pop reconstruction would require a global (incorrect) stack or best-effort heuristics.

Changes:
- `nvtx-events`: `RangePush` and `RangePop` gain a `thread_id: u32` field
- `nvtx-injection`: `on_range_push` / `on_range_pop` callbacks stamp the OS thread id via `gettid` (Linux); non-Linux falls back to a hash of `std::thread::current().id()` truncated to `u32`
- `nvtx-injection`: `on_domain_range_push_ex` / `on_domain_range_pop` (CORE2 path) updated identically
- `nvtx-example`: new `tests/thread_id.rs` — captures a two-thread interleaved Push/Pop session and asserts each event carries the expected thread id

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -130,6 +130,7 @@ cargo_metadata = "0.23"
 cc = "1.2"
 http = "1"
 indexmap = "2"
+libc = "0.2"
 log = {version = "0.4.28", features = ["kv"]}
 petgraph = "0.8.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -130,7 +130,6 @@ cargo_metadata = "0.23"
 cc = "1.2"
 http = "1"
 indexmap = "2"
-libc = "0.2"
 log = {version = "0.4.28", features = ["kv"]}
 petgraph = "0.8.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
--- a/integrations/nvtx/example/tests/thread_id.rs
+++ b/integrations/nvtx/example/tests/thread_id.rs
@@ -1,0 +1,74 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! End-to-end proof that captured Push/Pop ranges carry a real OS thread id.
+//!
+//! Reuses the crate's [`run_capture`](nvtx_example::run_capture) with a
+//! collecting callback exporter and asserts every captured `RangePush`/`RangePop`
+//! carries a nonzero `thread_id`, and that Push/Pop emitted from the single
+//! example thread share one id. No GPU, no subprocess, no files.
+//!
+//! This lives in its own test file (hence its own test binary/process) rather
+//! than in `capture.rs`: `nvtx_injection::install_hook` is one-shot per process,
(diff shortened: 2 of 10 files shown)


integrations/nvtx/injection/src/init.rs
```diff
@@ -38,6 +38,33 @@ pub(crate) fn next_handle() -> u64 {
...
+/// (captured as [`NvtxEvent::NameThread`]) uses, so per-thread Push/Pop ranges
+/// resolve against a named thread. Read on the app thread from inside a callback.
+#[cfg(target_os = "linux")]
+pub(crate) fn current_thread_id() -> u32 {
+    // Use the raw `SYS_gettid` syscall rather than the glibc `gettid()` wrapper:
+    // the wrapper symbol is only exported by glibc >= 2.30, whereas the syscall
+    // works against every Linux libc (including the older conda sysroot in CI).
+    // SAFETY: `gettid` takes no arguments and cannot fail; it returns the calling
+    // thread's kernel task id (the Linux `gettid` id space).
+    unsafe { libc::syscall(libc::SYS_gettid) as u32 }
```
johanpel, line 63: Rather than performing this syscall for every push / pop, perhaps we could cache it in a thread-local `OnceCell`?

Cargo.toml
```diff
@@ -130,6 +130,7 @@ cargo_metadata = "0.23"
 cc = "1.2"
 http = "1"
 indexmap = "2"
+libc = "0.2"
```
johanpel: Nit: Since no other crate will use this (hopefully), I don't think this needs to be a workspace dep.

johanpel (commented):
Looks good overall, just two minor things we can improve:

integrations/nvtx/example/tests/thread_id.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    assert!(saw_push, "expected at least one captured RangePush");
+    assert!(saw_pop, "expected at least one captured RangePop");
+
+    // The example drives every Push/Pop from a single thread, so all stamped ids
+    // must match — proving a push and its matching pop carry the same thread_id.
+    let first = thread_ids[0];
+    assert!(
+        thread_ids.iter().all(|&t| t == first),
+        "all Push/Pop from the single example thread must share one thread_id; got {thread_ids:?}"
```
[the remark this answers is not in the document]
9prady9: Addressed in commits 52dc26d and 05077b6.

Replaced the single-thread test with `pushpop_four_threads_get_distinct_ids`: four threads synchronise at a `Barrier` before their first NVTX call so their push/pop events genuinely interleave in time. The test asserts:
1. Every stamped `thread_id` is nonzero.
2. Exactly 4 distinct `thread_id`s appear on Push events and 4 on Pop events — proving `gettid` returns per-thread values, not a single constant.
3. The Push set equals the Pop set — each thread's push and pop carry the same id.

The single-thread test was removed rather than kept alongside, because both tests share a process binary and `install_hook` is one-shot per process; the 4-thread test is a strict superset of what the original checked.

johanpel (approved):

Issue #433, opened 2026-07-21
DAG data-flow timeline cleanup


cmatzenbach:
- Move tasks/memory usage out of stats panel (or move into separate tab on that panel)
  * The problem here is the data panel gets real jumpy/jittery when scrubbing through the timeline. I noticed in the demo that there are no additional custom statistics to display in the panel, so nothing jumps around. 
- Don't render bars when there is no data being shown in the nodes
  * Showing a placeholder text in the bar or something might be a good option as well, I think the problem is its not clear what the bars are if there's no data displayed.
- Hide time slider + data flow bars when not relevant (i.e. when viewing logical plan)
  * Maybe default 'disabled' if there is no data-flow data for a plan, this may not depend on if its higher level plan or not. 
- Differentiate colors so it's clear which colors correspond between legend and bars
  * There can be n dimensions (data locations) so this is actually just the first two colors of the same palette, I think the decision is: do we want to use the same palette for both of these.
- Hovering over timeline while playing causes line to jump - add new line/indicator that only belongs to timeline playing (https://github.com/rapidsai/quent/issues/487)
- Highlight nodes that currently have usage
- Potentially allow controlling scrubber from timelines themselves
- Tasks can be hard to identify/differentiate during periods of low memory usage
  * Node detail panel should take care of this mostly, but this assumes you have a node selected.
- Same time is taken to cycle through, regardless of how long of a time is actually selected on the main scrubber
  * What is the expectation here? Real time?
- Clarify what numbers in bottom right on nodes represent

johallar:
Another thing i'm thinking about is the operator table integration with the time scrubber

PR #477, opened 2026-07-28
refactor(ui): organize settings

### LINKED ISSUE #478: refactor(ui): organize settings into sections, align labels/inputs
Enough settings have been added that it's getting hard to figure out whats what in the lefthand panel settings. 

Image

Time to refactor this into sections, and get things lined up so it's easier to visually sort through.

johallar:
# Description
* Organizes the DAG settings into sections
* Creates a re-usable control grid + section + field components so that this style of settings can be re-used. 
  * Aligns labels/inputs vertically so it's easier to visually grok. 
  * Configurable number of columns, with minimum field widths that will wrap when fields get too small. 
* More explicit controls/labels for the "data location" dimension, so it's clear that one or more MUST be selected. Still gives this room to grow to n dimensions

## Related Issues
Fixes #478 

## Testing
* All settings work as before

## Screenshots

Screenshot 2026-07-28 at 2 40 01 PM

Wrapping behavior:

https://github.com/user-attachments/assets/66c67d09-45d4-4a03-9376-4902b854d75c


cmatzenbach:
I wonder if we should disable the settings in the "Data Flow" category when it is disabled - I feel like that is fairly standard UX, to not allow modifications of disabled features.

cmatzenbach:
Maybe not something that needs to be addressed here, but I do wonder if this area is getting a bit too crowded - the label is running behind the "x", which makes it a bit hard to see. Lmk what you think.
image

I wonder if we could slightly decrease the spacing between the "x" and caret icon, giving it a bit more room to stand out. Or maybe even bump the caret icon to the very end of the dropdown, since it's just a visual indicator and not something that needs to be clicked (unlike the "x").

cmatzenbach:
Not a big deal, but looking at it, wondering if we should decrease the size and padding around the checkmark - it seems to be bigger than the text and taking up more room than is probably necessary:

image

Also, could we remove that empty padding to the left of the selection in the dropdown? Seems unnecessary and is cutting out room for the content to display.

image

cmatzenbach:
Another thing I noticed - the "Bar Labels" dropdown has an inherent default selection of "Tasks"  but has placeholder text - we should remove the placeholder text and show the default selection instead (like how "Flow measure" does it).

image

johallar:
Agree on all. With some of the finer details of the `select` ui component I think we can tackle in a separate issue since that'll have effects across the FE. I fixed the left padding, and the x covering the text (now truncates w ellipses), and moved the caret over a bit to give us max space.

johallar:
> Another thing I noticed - the "Bar Labels" dropdown has an inherent default selection of "Tasks" but has placeholder text - we should remove the placeholder text and show the default selection instead (like how "Flow measure" does it).
> 
> image

I do agree, but trying to keep this PR focused on the settings layout. Logging an issue in https://github.com/rapidsai/quent/issues/433 issue though

ui/packages/@quent/components/src/ui/control-grid.test.tsx
```diff
@@ -0,0 +1,63 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cmatzenbach: Will need to add the comma in these to make `pre-commit-hooks` happy :)

johallar:
https://github.com/rapidsai/quent/issues/486 For the select placeholder -> option change

cmatzenbach:
Looks _so much_ better now, thanks for taking care of those! The dropdowns are much more readable now. Might want to add the commas on the new file headers, but other than that good to go.

cmatzenbach (approved):

PR #479, opened 2026-07-28
chore: modify license headers to match canonical rapidsai format

cmatzenbach:
# Description

Previously-added license headers were missing a comma that is auto-inserted by `pre-commit-hooks`. This updates all license headers to match this format.


johallar (approved):

johallar (changes requested):
Actually, there are two places where we should update this as well while we're here, `CONTRIBUTING.md` and `ui/examples/AGENTS.md`

cmatzenbach:
> Actually, there are two places where we should update this as well while we're here, `CONTRIBUTING.md` and `ui/examples/AGENTS.md`

It was updated in both places - I did a grep find and replace so nothing was missed

johallar (approved):

johallar:
my bad

PR #480, opened 2026-07-28
feat(ui): pivot table bigint support

### LINKED ISSUE #475: Refactor pivot table components to accept bigints
To finish what was started in PR #394, there are three areas where we are currently accepting precision loss, which require this refactor:

1. getSortValue line 176 - sort order for non-aggregated rows with large bigints
2. getSortValue agg path - agg.sum, agg.min, agg.max come from PivotedRowAgg which stores number | null, so even though we now compute them more accurately as bigints, they're converted to number before being stored and returned here
3. gradientBg color scaling - uses the number sort value to position a value on a color gradient, so large bigints that lost precision in getSortValue get slightly wrong colors

Cascading changes are needed for gradientBg and the column range comparisons in PivotedStatsTable.tsx.

cmatzenbach:
# Description

Modifies pivot table sorting function, gradient background generation, stats computation, and typescript definitions to support `BigInt` without converting to `Number`. The only places we are still coercing `BigInt` are where it is needed - for mean computation and gradient coloring, where precision loss is acceptable. In addition, updates unit tests to test for `BigInt` without coercing.

## Related Issues

Closes #475

## Testing

1. Ensure unit tests pass
2. If you have a `ndjson` file with `BigInt` statistics, ensure the pivot table functions as normal

### DIFF
--- a/ui/packages/@quent/components/src/pivot-table/PivotedStatTable.tsx
+++ b/ui/packages/@quent/components/src/pivot-table/PivotedStatTable.tsx
@@ -43,7 +43,7 @@ const numericSortingFn: SortingFn<PivotedRow> = (rowA, rowB, columnId) => {
   if (a == null) return 1;
   if (b == null) return -1;
   if (typeof a === typeof b) return a < b ? -1 : 1;
-  return Number(a) < Number(b) ? -1 : 1;
+  return (a as number) < (b as number) ? -1 : 1;
 };
 
 function DataHeader({ stat, sortInfo, onSort, className, style }: DataHeaderProps) {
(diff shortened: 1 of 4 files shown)


ui/packages/@quent/components/src/pivot-table/PivotedStatTable.tsx
```diff
@@ -36,6 +36,16 @@ import {
 
 const HIGHLIGHT_WASH = 'inset 0 0 0 999px hsl(var(--primary) / 0.07)';
 
+const numericSortingFn: SortingFn<PivotedRow> = (rowA, rowB, columnId) => {
```
johallar: Better to live in operator table utils

johallar (approved):
One organization thing, otherwise 🙇 tyty

ui/packages/@quent/components/src/pivot-table/PivotedStatTable.tsx
cmatzenbach: Good point, moved it over

PR #482, opened 2026-07-29
fix(time): keep arrival order for equal timestamps in TimeOrderedCollector

9prady9:
`TimeOrderedCollector::push` has two insertion paths that disagree about where a tie goes:

- **Fast path** (`last.timestamp() <= new`) — appends *after* existing equals.
- **Slow path** — used `partition_point(|s| s.timestamp() < new)`, a *lower* bound, which inserts *before* every item it ties with, reversing their arrival order.

The slow path only runs when the newcomer is older than the last element, so a stream that never goes backwards in time never reaches it. That is why existing coverage missed this — a monotonically non-decreasing stream takes the fast path every time.

Concretely, pushing `[(t=10,a), (t=20,b), (t=10,c)]` yielded `[c, a, b]` — `c` jumped ahead of `a` despite arriving after it.

**Why it matters:** equal timestamps mean the clock lost the ordering, leaving arrival order as the only remaining signal for which item came first. Reversing it silently corrupts reconstruction downstream. For an NVTX push/pop pair sharing a clock tick, it turns a valid zero-duration range into a fabricated duration plus a spurious synthetic close — plausible-looking output, no error raised.

Changes the predicate to `<=` (upper bound) so both paths agree, and adds regression tests that drive the slow path with real ties. The new tests live in their own module because the existing one is gated behind `__test-clock-override`, which is excluded from `default-members`.

Shared by `crates/analyzer` (`fsm/runtime.rs`, `fsm/events.rs`, `resource/runtime.rs`); its suite and the full workspace suite pass unchanged.

### DIFF
--- a/crates/time/src/lib.rs
+++ b/crates/time/src/lib.rs
@@ -124,10 +124,7 @@ pub trait Timestamp {
 /// in which case [`Self::push`] is O(1). Out-of-order items are inserted via
 /// binary search (O(log n) search + O(n) insertion).
 ///
-/// The ordering is **stable**: items sharing a timestamp keep their arrival
-/// order regardless of which insertion path they took. This matters when equal
-/// timestamps mean the clock lost the ordering, leaving arrival order as the
-/// only remaining signal for which item came first.
+/// Stable: items with equal timestamps keep their arrival order.
 pub struct TimeOrderedCollector<T>(Vec<T>);
 
 impl<T> Default for TimeOrderedCollector<T> {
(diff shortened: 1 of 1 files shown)


crates/time/src/lib.rs
```diff
@@ -141,9 +143,11 @@ where
         {
             self.0.push(state);
         } else {
+            // `<=` (upper bound): late arrivals land after ties, matching the fast path.
+            // `<` (lower bound) would insert before ties, reversing arrival order.
```
johanpel, line 147: Nit: this bug seems trivial enough to not require a historical comment:
```suggestion
```

crates/time/src/lib.rs
```diff
@@ -123,6 +123,8 @@ pub trait Timestamp {
 /// Optimized for when the common case is that items arrive in timestamp order,
 /// in which case [`Self::push`] is O(1). Out-of-order items are inserted via
 /// binary search (O(log n) search + O(n) insertion).
+///
+/// Stable: items with equal timestamps keep their arrival order.
```
johanpel, line 127: Nit: since "arrival" is unnecessarily context-sensitive
```suggestion
/// Stable: items with equal timestamps retain the order in which they are pushed
```

johanpel (approved):
Good improvement / fix, thanks @9prady9

9prady9:
will take care of the docs corrections in phase 2 PR. dont need to do another round of push, approve for these I think.

Issue #491, opened 2026-07-30
Migrate query engine domain and examples to schema-based events


johanpel:
YAML can capture everything the legacy modeling crates can. Migrate the event pipeline into the analyzer to use the types generated in this way

PR #493, opened 2026-07-30
refactor(analyzer): decouple QE entities from in-memory storage

johanpel:
# Description

- Define entity-specific analyzer traits and expose entity implementations through `QueryEngineModel` associated types.
- Move the existing event-backed model, view, and entities under `plain::legacy`.
- Make plan-tree construction generic over `PlanEntity`.
- Route derived operator-span updates through mutable model and entity traits instead of directly accessing legacy storage.

## Related Issues

Part of #491 

_Written by Codex._


dhruv9vats (approved):

9prady9 (approved):
lgit

Issue #215, opened 2026-06-10
Entities stacked swimlane/gantt view


johallar:
Instead of overlaying on timelines, create a new view similar to operator timelines using vertically stacked swimlanes. This resolves many overlapping entities problems

PR #496, opened 2026-07-30
feat(ui): gantt chart implementation, entities swimlanes 2

johallar:
# Description
Captures common quent functionality (data zoom sync, xaxis pointer, etc) in a GanttChart component, much like Timeline. This implementation will be used to implement the operators gantt chart and the entities gantt chart. This PR only updates the OperatorGantt to the new base implementation. 

## Related Issues
relates to #215 

## Testing
Operators gantt chart functions as previously

### DIFF
--- a/ui/packages/@quent/components/src/gantt-chart/GanttChart.tsx
+++ b/ui/packages/@quent/components/src/gantt-chart/GanttChart.tsx
@@ -1,0 +1,166 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+  type ComponentProps,
+  type ReactNode,
+} from 'react';
+import EChartsReactCore from 'echarts-for-react/lib/core';
+
+import type { EChartsInstance } from 'echarts-for-react';
+import { echarts } from '../lib/echarts';
+import { registerAxisPointerSync, unregisterAxisPointerSync } from '../lib/timeline.utils';
+import { useChartConnect } from '../lib/useChartConnect';
+import { useMinZoomSpanPct } from '../lib/useMinZoomSpanPct';
+import { useTimelineWheelNavigation } from '../lib/useTimelineWheelNavigation';
+import { CHART_GROUP } from '../timeline/Timeline';
+import { useTimelineEchartsTheme } from '../timeline/timelineEchartsTheme';
+import { HiddenScroll } from '../ui/thin-scroll';
+import { observeGanttHover, type GanttHover } from './hover';
+import {
+  buildGanttOption,
+  type GanttDatum,
+  type GanttGridSpacing,
+  type GanttRenderItem,
+  type GanttSeriesCursor,
+} from './options';
+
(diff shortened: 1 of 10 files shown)


ui/packages/@quent/components/src/gantt-chart/GanttChart.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  const { handleChartReady, instanceRef } = useChartConnect({
+    durationSeconds,
+    chartGroup: CHART_GROUP,
+    onReady: onChartReady,
+  });
+
+  useEffect(() => {
+    return () => {
+      chartCleanupRef.current?.();
```
cmatzenbach, line 137: In OperatorGantChart.tsx there was a comment ("empty data replaces the chart without unmounting"). Do we need to do something similar here or are we fine unmounting each time?  i.e.
```
  return () => {
    if (data.length > 0) return;
    chartCleanupRef.current?.():
    ....
```

ui/packages/@quent/components/src/gantt-chart/GanttChart.tsx
```diff
@@ -0,0 +1,166 @@
...
+  const { themeName } = useTimelineEchartsTheme(isDark);
+  const [hover, setHover] = useState<GanttHover | null>(null);
+  const minZoomSpanPct = useMinZoomSpanPct(durationSeconds);
+  const attachWheelNavigation = useTimelineWheelNavigation(minZoomSpanPct);
+  const wrapperRef = useRef<HTMLDivElement | null>(null);
+  const chartCleanupRef = useRef<(() => void) | null>(null);
+
+  const { yAxisCategories, rowCount } = useMemo(() => {
+    if (data.length === 0) return { yAxisCategories: [] as number[], rowCount: 0 };
+    const maxRow = Math.max(...data.map(datum => datum.value[2]));
```
cmatzenbach: In the case where we have a really large dataset, we could hit the limit for `Array.map` - maybe considering using `reduce` instead:
```
 const maxRow = data.reduce((max, datum) => Math.max(max, datum.value[2]), 0);
```

ui/packages/@quent/components/src/gantt-chart/options.ts
```diff
@@ -0,0 +1,118 @@
...
+  seriesName,
+  renderItem,
+  minZoomSpanPct,
+  cursor,
+  gridSpacing = TIMELINE_SPACING,
+}: BuildGanttOptionParams<T>): EChartsOption {
+  return {
+    animation: false,
+    tooltip: {
+      show: true,
```
cmatzenbach: Previously in `OperatorGantChart` we used `tooltip: { show: false }` - is there any benefit to always keeping the tooltip when `renderTooltip` is not passed? Should this be conditional based on whether or not the caller needs a tooltip?

cmatzenbach (approved):
A few general questions for you to consider, but other than that looks good!

ui/packages/@quent/components/src/gantt-chart/GanttChart.tsx
johallar: That effect would run the cleanup function when there was 0 operators, but I think now we want to keep the xaxis pointer and mouse wheel effects even when there is no data so it's goneeee. This one will run the cleanup function when the component is unmounted so we don't leak mouse wheel listeners.

PR #517, opened 2026-08-03
chore(deps): bump syn to 3 and prettyplease to 0.3

9prady9:
# Description

Bumps `syn` 2 → 3 and `prettyplease` 0.2 → 0.3 in a single change, because
the two cannot be upgraded independently, and moves both to
`[workspace.dependencies]` so they stay in lockstep going forward.

`prettyplease` 0.3.0 is the first release built against `syn` 3, and it
exposes `syn::File` in its public API. Both `crates/codegen/src/common.rs`
and `crates/open/src/wrapper.rs` construct a `syn::File` and hand it to
`prettyplease::unparse`, so bumping either side alone puts two incompatible
`syn` versions on that boundary and fails to compile:

- syn-only (as in #505): `` the trait `syn::parse::Parse` is not implemented
  for `syn::file::File` `` — `syn::parse2` resolves to syn 3 while the
  inferred `File` comes from syn 2 via prettyplease.
- prettyplease-only: the mirrored `E0308`, `expected syn::file::File`
  (3.0.3), `found syn::File` (2.0.118).

Changes:

- Pin `syn = { version = "3.0.3", features = ["full", "parsing"] }` and
  `prettyplease = "0.3.0"` in `[workspace.dependencies]`
- `crates/codegen`, `crates/open`, `crates/instrumentation-build` and
  `crates/model-macros` now use `workspace = true`; model-macros keeps its
  extra `extra-traits` feature on top of the workspace base
- Regenerated `Cargo.lock` and `crates/instrumentation-build/example/Cargo.lock`

No source changes were needed — the syn 3 APIs used here (`parse2`,
`parse_str`, `File`, derive-input types) are unchanged. `syn 2.0.118` remains
in the lockfile as a transitive dependency of `bindgen`, `prost-build`,
`tonic-build` and `wit-bindgen`, which is expected and does not conflict.

## Related Issues

Supersedes #505 and #513, which split this upgrade across two PRs and
therefore both fail CI.

## Testing

- `pixi run cargo fmt --all --check` — clean
- `pixi run cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` — clean
- `pixi run cargo test --workspace --all-features --locked --all-targets` — all pass
- `pixi run cargo deny check` — advisories, bans, licenses, sources ok
- `cargo test --locked` in `crates/instrumentation-build/example` (its own
  workspace) — builds and passes
- The workspace-deps commit leaves `Cargo.lock` byte-identical, confirming
  the refactor is resolution-neutral

The `quent-open compatibility` job could not be run locally (it needs a
published prior commit), but its failure on #505 was this exact compile
error, which no longer reproduces.

## Screenshots

N/A — no UI changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,6 +133,9 @@ indexmap = "2"
 log = {version = "0.4.28", features = ["kv"]}
 petgraph = "0.8.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
+# `prettyplease` and `syn` are a matched pair: prettyplease exposes `syn::File`
+# in its public API, so both must move to a new `syn` major version together.
+prettyplease = "0.3.0"
 prost = "0.14.1"
 pyo3 = "0.29"
 quent-resource = { path = "crates/resource" }
@@ -141,6 +144,7 @@ rustc-hash = "2"
 serde = { version = "1.0.228", features = ["derive"] }
 serde_json = "1.0.145"
 smallvec = {version = "1.15.1", features = ["serde"] }
+syn = { version = "3.0.3", features = ["full", "parsing"] }
 thiserror = "2.0.17"
 tokio = "1.48.0"
 tokio-stream = "0.1.17"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,8 +133,6 @@ indexmap = "2"
 log = {version = "0.4.28", features = ["kv"]}
 petgraph = "0.8.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
-# `prettyplease` and `syn` are a matched pair: prettyplease exposes `syn::File`
-# in its public API, so both must move to a new `syn` major version together.
 prettyplease = "0.3.0"
 prost = "0.14.1"
 pyo3 = "0.29"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -315,7 +315,7 @@ dependencies = [
  "clang-sys",
  "itertools",
  "log",
- "prettyplease",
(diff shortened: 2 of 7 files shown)


johanpel (approved):
Looks good, thanks @9prady9. While we're modifying this, perhaps we can move these to workspace deps, pin it there, and use `workspace = true` in the crates.

9prady9:
Thanks @johanpel — done in 7e35103. Both are now pinned in `[workspace.dependencies]` (`syn = { version = "3.0.3", features = ["full", "parsing"] }`, `prettyplease = "0.3.0"`), with `codegen`, `open`, `instrumentation-build` and `model-macros` switched to `workspace = true`. `model-macros` adds `extra-traits` on top of the workspace base.

I left a comment above the entries noting that the two have to move together across a `syn` major, since that's the whole reason #505 and #513 couldn't land on their own.

That commit leaves `Cargo.lock` byte-identical, so the refactor is resolution-neutral; clippy/tests re-run clean.

Cargo.toml
```diff
@@ -133,6 +133,9 @@ indexmap = "2"
 log = {version = "0.4.28", features = ["kv"]}
 petgraph = "0.8.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
+# `prettyplease` and `syn` are a matched pair: prettyplease exposes `syn::File`
+# in its public API, so both must move to a new `syn` major version together.
```
johanpel: Nit: compiler will tell us so the comment seems noisy

johanpel (approved):

PR #520, opened 2026-08-03
feat(store): add schema-based filesystem event loading

johanpel:
# Description

Add `quent-store` with low-level typed APIs under `quent_store::event` and an initial `event::filesystem` implementation. 

Entity-specific retrieval is always available, enabling `umbrella_event` in `quent-store-build` adds model-wide retrieval through the generated umbrella event type, which is necessary during PoC -> schema based migration, but might be useful afterwards for other reasons too.

Filesystem retrieval validates context provenance, propagates import failures, and rejects recognized formats whose feature is disabled. Event-file traversal is deterministic by path, but raw event iteration has no timestamp or causal ordering guarantee.

Add `quent-store-build` to generate event types, entity membership, and optional model-wide filesystem descriptors from the same schema used for instrumentation. All currently supported filesystem importers deserialize through serde, so generated store event and record types always derive `Serialize` and `Deserialize` for now.

Extend the instrumentation example with reusable callback and NDJSON entry points, then add a store-build round trip demonstrating retrieval of one entity type and all model events. Both examples are workspace members and run in CI now.

## Scope and follow-ups

This PR establishes the raw event-retrieval layer. It deserializes each complete event into an owned Rust value, but does not interpret those events using higher-level schema semantics.

Separate follow-up changes will:

- Connect existing analyzers to this retrieval API, which will be the final step in the migration.
- Add APIs for inspecting files that were not consumed and define handling for malformed or symlinked filesystem layouts. This implementation targets directories produced directly by Quent filesystem exporters only.

Some more advanced future things intentionally deferred:
- Add schema-aware views that validate FSM transitions or derive resource state, group events by entity ID and reconstruct entity histories or current state.
- Asynchronous retrieval, lazy evaluation / dataframe-like APIs, and non-filesystem storage such as databases or live event streams

## Related Issues

Part of the import side of #191. Follows #519.

## Testing

- `pixi run cargo test -p quent-store --all-features --locked`
- `pixi run cargo test -p quent-store --no-default-features --features io-ndjson --locked`
- `pixi run cargo clippy -p quent-store --all-targets --all-features --locked -- -D warnings`
- `pixi run cargo test -p quent-store-build --locked`
- `pixi run cargo clippy -p quent-store-build --all-targets --locked -- -D warnings`
- `pixi run cargo run -p quent-store-build-example --locked`

_Written by Codex._ and a bit of human


crates/instrumentation-build/example/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+
+/// Exports the demo events as NDJSON and returns their context ID.
+pub fn run_with_ndjson(
+    root_export_path: impl Into<PathBuf>,
+) -> Result<Uuid, Box<dyn std::error::Error>> {
+    let context = Context::try_new(ExporterOptions::FileSystem(FileSystemExporterOptions::new(
+        FileSystemFormat::Ndjson,
+        root_export_path.into(),
+    )))?;
+    emit_events(context)
```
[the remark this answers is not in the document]
johanpel: Skipped. run_with_ndjson is already exercised end to end by the store-build example in CI, which validates the context sidecar through Store and decodes the exported NDJSON. A separate assertion-heavy test was intentionally avoided to keep the user-facing example simple.

crates/store/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+[features]
+default = ["io-ndjson", "io-msgpack", "io-postcard"]
+io-ndjson = ["quent-io/ndjson"]
+io-msgpack = ["quent-io/msgpack"]
+io-postcard = ["quent-io/postcard"]
+
+[dependencies]
+quent-build-info = { path = "../build-info" }
+quent-events = { path = "../events" }
+quent-io = { path = "../io", default-features = false }
```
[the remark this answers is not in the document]
johanpel: Skipped. The repository predominantly declares internal crates with local path dependencies and declares quote and tempfile directly in individual manifests. Converting only these new manifests would be inconsistent; a workspace-wide dependency-centralization change should be handled separately if that becomes repository policy.

crates/store-build/example/src/main.rs
```diff
@@ -0,0 +1,36 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Runs instrumentation and loads its filesystem-exported events.
+
+use demo::{Demo, Query};
+use quent_store::event::filesystem::Store;
+use quent_store::event::{EntityEventStore, ModelEventStore};
+
+#[allow(unused)]
```
mbrobbel: Nit: do we need this?
```suggestion
```

crates/store-build/src/lib.rs
```diff
@@ -0,0 +1,212 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Generates schema-based typed APIs for retrieving stored events.
+//!
+//! Add `quent-store-build` to `[build-dependencies]`, call [`generate`] from
+//! `build.rs`, and include the generated file from Cargo's `OUT_DIR`.
+//! The crate including that source needs normal dependencies on `quent-store`,
+//! serde-enabled `quent-events`, and derive-enabled `serde`.
```
mbrobbel, line 9: Nit: maybe add code snippets to doc:

`build.rs`:
```rust
...
```

`Cargo.toml`
```toml
[build-dependencies]
...

[dependencies]
...
```

crates/store/src/event/filesystem.rs
```diff
@@ -0,0 +1,411 @@
...
+                    source,
+                })
+            })) as EventIterator<T, Error>)
+        };
+
+        stream().unwrap_or_else(|error| Box::new(std::iter::once(Err(error))))
+    }))
+}
+
+fn event_files(context: &Path, entity: &str) -> Result<Vec<EventFile>> {
```
mbrobbel, line 236: Nit: these functions should have some doc comments

mbrobbel (approved):

crates/store-build/example/src/main.rs
johanpel: https://github.com/rapidsai/quent/pull/520/commits/08de2c50f732bc27d5502c0e1630d64648fa7989

But since the example isn't using all of the types imported by the generated code we still need an allow on those.

crates/store-build/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/520/commits/743756437fdb7dec9aa6352ad4d98f9a6a42371b

crates/store/src/event/filesystem.rs
johanpel: https://github.com/rapidsai/quent/pull/520/commits/e19d1fcacd98501fe1243bd168e66bece49ef9ea

PR #525, opened 2026-08-03
feat(os): add OS process and thread mod

### LINKED ISSUE #522: Provide constraint to mark entities as representing OS primitives
Provide some kind of `quent-os` "schema constraints" / semantic module that allow us marking certain entities as matching concepts from the OS, e.g. with "process" entities (i.e. "this entity represent an OS process, and this is its native ID"). Same for thread resource entities (i.e. "this unit resource represents an OS thread, it is a thread owned by this process entity id", etc.). This will allow us to correlate things that have a context-sensitive IDs (processes, threads, external event sources) to things that have universally unique ids (Quent entities) and figure out what goes where.

This is useful for various reasons, including where external event sources (e.g. NVTX) assume process-global identifiers are unique enough. Also useful for when e.g. statistical profilers are attached that need to be correlated with quent events, etc. etc.

johanpel:
# Description

Add canonical records and schema constraints for associating Quent entities with operating-system processes and threads.

- Define annotated `quent::os::Process` and `quent::os::Thread` identity records.
- Require exact record shapes, direct once-event fields, one identity per entity, and process-scoped thread entities.
- Resolve qualified record paths and automatically add referenced OS records when parsing YAML.
- Update the instrumentation example and add constraint and YAML validation coverage.

## Related Issues

Closes #522 

## Testing
- `pixi run --frozen cargo fmt --all -- --check`
- `pixi run --frozen cargo clippy -p quent-constraints -p quent-os -p quent-yaml -p quent-instrumentation-build -p quent-instrumentation-build-example --all-targets --all-features --locked -- -D warnings`
- `pixi run --frozen cargo test -p quent-constraints -p quent-os -p quent-yaml -p quent-instrumentation-build -p quent-instrumentation-build-example --all-targets --all-features --locked`

_Written by Codex._


crates/yaml/src/lower.rs
```diff
@@ -461,15 +464,56 @@ fn entity_ref_type(
 
 /// A bare name that is not a [`BuiltinType`], lowered as a record reference.
 fn record_ref(name: &str, path: &str, sink: &mut Diagnostics) -> Option<DataType> {
-    match Identifier::try_new(name) {
-        Ok(id) => Some(DataType::Record(id.into())),
+    match name.parse::<Path>() {
```
9prady9, line 471: Please preserve `quent::os::*` paths in the WASM namespace preprocessing, regenerate the checked-in bundle, and add an OS parsing test.

crates/instrumentation-build/example/src/lib.rs
```diff
@@ -96,3 +112,48 @@ fn emit_events(context: Context<Demo>) -> Result<Uuid, Box<dyn std::error::Error
...
+        std::io::ErrorKind::Unsupported,
+        "native thread IDs are unsupported on this platform",
+    ))
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn native_thread_id_is_nonzero() {
+        assert_ne!(super::current_native_thread_id().unwrap(), 0);
```
9prady9, line 160: Please gate `native_thread_id_is_nonzero` to Linux/macOS/Windows and assert `Unsupported` elsewhere instead of unconditionally calling `unwrap()`.

9prady9 (commented):
Shall I go ahead and use these for NVTX, or do you have changes in the pipeline for that?

crates/instrumentation-build/example/src/lib.rs
mbrobbel: Maybe we should document the supported platforms?

crates/os/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+///
+/// ## Requirements
+///
+/// 1. A canonical process or thread record may be used only as a direct field
+///    of an entity event.
+/// 2. An entity may carry each canonical record in only one field of one event,
+///    whose cardinality must be [`Cardinality::Once`].
+/// 3. An entity may not carry both canonical records.
+/// 4. A thread entity must be transitively scoped under a process entity by
+///    tree-forming entity references.
```
mbrobbel: This makes sense, but I wonder if it's maybe too strict?

crates/os/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+        record: Path,
+        events: Vec<Identifier>,
+    },
+    #[error("{entity}.{event}: OS record `{record}` may be carried by only one event field")]
+    OsRecordUsedMultipleTimesByEvent {
+        entity: Path,
+        record: Path,
+        event: Identifier,
+    },
+    #[error("{entity}: an entity cannot represent both an OS process and an OS thread")]
```
mbrobbel: I'm also wondering if we could/should relax this requirement.

mbrobbel (approved):

crates/yaml/src/lower.rs
johanpel: https://github.com/rapidsai/quent/pull/525/commits/63a334d3efe65fd513e7fbfe41fe9436ab220938

crates/instrumentation-build/example/src/lib.rs
johanpel: https://github.com/rapidsai/quent/commit/6910f2611f0774046fd245f2444f85867c9d5e05
https://github.com/rapidsai/quent/pull/525/commits/4e4ddad67bdc78873bb099f657fb23cc1c8a3e93

crates/os/src/lib.rs
johanpel: I think this is a good idea.

https://github.com/rapidsai/quent/pull/525/commits/8361948cfa8f34a26e6e3e52f61fb2b0523970b4

crates/os/src/lib.rs
johanpel: Since the main purpose of adding these platform-specific IDs is correlation with externally provided events (e.g. by NVTX), I think it's best to require this correlation to be unambiguous, especially when dealing with distributed processes.

I can't come up with a use-case where this requirement would get in the way, but maybe you have something in mind?

mbrobbel: A single entity that tracks the main thread of a process

johanpel: What would prevent that entity from adding the process id field?

mbrobbel: Oh, I didn't realize that would work. So with the other requirement removed this should be possible. Do you still need to create a self-reference?

johanpel: > Do you still need to create a self-reference?

No, this shouldn't be necessary, see: https://github.com/rapidsai/quent/blob/422d98e31562e112cb1ec96cb625de6792f769c4/crates/os/tests/os-constraint.rs#L339-L356

mbrobbel (approved):

johanpel:
I'm working on some more mods in parallel, e.g. #686 #683, but I realized the YAML syntax used there is a bit more consistent w.r.t. what I'm proposing here, which feels kind of weird compared to the other mods.

I think it's best to consolidate on
```yaml
<whatever scope>:
  <mod name>:
    <scope-specific mod key>: <value>
```

So here I'm thinking about moving to:
```yaml
entities:
  Foo:
    events:
      bar:
        attributes:
          my_field_name_for_pid: { os: process }
          my_field_name_for_tid: { os: thread }
          
```

Which would make it consistent with the other approaches

johanpel:
> I'm working on some more mods in parallel, e.g. #686 #683, but I realized the YAML syntax used there is a bit more consistent w.r.t. what I'm proposing here, which feels kind of weird compared to the other mods.
> 
> I think it's best to consolidate on
> 
> ```yaml
> <whatever scope>:
>   <mod name>:
>     <scope-specific mod key>: <value>
> ```
> 
> So here I'm thinking about moving to:
> 
> ```yaml
> entities:
>   Foo:
>     events:
>       bar:
>         attributes:
>           my_field_name_for_pid: { os: process }
>           my_field_name_for_tid: { os: thread }
>           
> ```
> 
> Which would make it consistent with the other approaches

This is processed in https://github.com/rapidsai/quent/pull/525/commits/dd06efbe9bfda2e48daed2bbde0c0676bf62982a

johanpel:
Minor fixes since review and #191 got merged

- WASM build: https://github.com/rapidsai/quent/pull/525/commits/bc0ab4a5c6c694c0fe01d2b2be307f9387573d5e
- Slight improvement to diagnostic: https://github.com/rapidsai/quent/pull/525/commits/05cdd44e6c35e102d46c692370a5ee0ca56bfece
- Doc improvements: https://github.com/rapidsai/quent/pull/525/commits/8988da525aba73c7af17f58c944a914c534d95c4 + https://github.com/rapidsai/quent/pull/525/commits/5a05e992b43ebf0175278ad49ef804dc75187c7f

johanpel:
@9prady9 can you please take another pass? Additional changes are mentioned above.
This will be very useful for [NVTX](https://github.com/rapidsai/quent/pull/740) integration (and many other third party event sources)

crates/yaml/Cargo.toml
```diff
@@ -9,6 +9,7 @@ autoexamples = false
 indexmap = { workspace = true, features = ["serde"] }
 quent-constraints = { path = "../constraints" }
 quent-fsm = { path = "../fsm" }
+quent-os = { path = "../os" }
```
9prady9, line 12: Please regenerate `experimental/vibe/codegen/Cargo.lock`. The new dependency makes clean codegen builds with `--locked` fail; CI masks this by updating the lockfile during Clippy.

experimental/vibe/ui/yaml-wasm/src/lib.rs
```diff
@@ -17,27 +17,38 @@ pub fn parse_schema_json(source: &str) -> Result<String, JsValue> {
 }
 
 fn parse_schema_value(source: &str) -> Result<Value, String> {
-    let encoded = source.replace("::", NAMESPACE_SEPARATOR);
+    let namespace_separator = collision_free_namespace_separator(source);
+    let encoded = source.replace("::", &namespace_separator);
```
9prady9: Please preserve `quent::os::*` paths during preprocessing. Qualified references pass the native parser but fail in the browser with unresolved references. Add a qualified-path test and regenerate the WASM bundle.

crates/os/src/lib.rs
```diff
@@ -0,0 +1,429 @@
...
+/// leader. On macOS, [`pthread_threadid_np`] returns a thread ID separately from
+/// the process ID, and [`pthread_main_np`] identifies the main thread. On Windows,
+/// [`GetCurrentProcessId`] and [`GetCurrentThreadId`] return separate system-wide
+/// IDs valid for the lifetime of the process or thread.
+///
+/// An event producer assigning both roles must verify that the thread is the
+/// process's main or primary thread. Producers and consumers must distinguish
+/// the IDs by their canonical record paths and must not rely on numeric equality.
+///
+/// ## Requirements
```
9prady9, line 65: nit: perhaps also a note that this validator should run alongside `RefTargetConstraint` and `RefTreeConstraint`, checking base validation and all results. 

We have added that info in doc string, but one might just jump to requirements and just use it. Eventually they will figure it out, might be worth making the requirements comprehensive.

crates/os/tests/os-constraint.rs
```diff
@@ -0,0 +1,421 @@
...
+                error,
+                OsError::InvalidOsRecord(RecordValidationError::UnexpectedField { .. })
+            ))
+            .count(),
+        2
+    );
+}
+
+#[test]
+fn thread_may_be_transitively_scoped_under_process() {
```
9prady9, line 199: perhaps add tests for scope references inside nested/shared records and reference payloads, plus cyclic scopes, to guard those behaviors.

crates/yaml/Cargo.toml
johanpel: https://github.com/rapidsai/quent/pull/525/commits/6c00a70cbf8359de529a60dc4319a852ee7a8b34

Also fixed the CI not using --locked

experimental/vibe/ui/yaml-wasm/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/525/commits/f173d1af12024e9d2a1d067989bebd98b5f21bdd

crates/os/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/525/commits/c6fff5aa9f0a7c0858886f2fd6e30f709c610a70

crates/os/tests/os-constraint.rs
johanpel: https://github.com/rapidsai/quent/pull/525/commits/53c8d36d9204bd46cfa814e9098dd0e61d58f872

9prady9 (approved):
cool, awaiting this for #618 🦅

PR #526, opened 2026-08-03
feat(ui): entities swimlanes 3

johallar:
# Description
Implements the entities gantt chart component as part of the resource timelines + pagination

Unit tests are like ~400 of these lines, sorry about the size regardless i'm happy to split if you want

## Related Issues
relates to #215 

## Testing
* Entities should only be shown with the same states as the FSM in the timeline above
* Operator filters should filter entities
* If you want to test the entities paging with a much lower limit:
  * Change const ENTITIES_PER_PAGE = 100; to be much lower (~10)
  * Generate a query with tons of tasks: `cargo run -p quent-simulator -- --num-query-groups 1 --num-queries 1 --num-workers 2 --num-threads 4 --num-tasks 2048 --exporter collector --collector-address http://localhost:7836`
* OperatorGanttChart now has a tooltip as well, which shows all operators at the current xaxis mark

## Screenshots

https://github.com/user-attachments/assets/b42bf441-8355-488f-ae06-e876f72f154b


ui/src/components/LongEntitiesRow.tsx
```diff
@@ -0,0 +1,123 @@
...
+  resourceId,
+  durationSeconds,
+  fsmTypes,
+  isDark,
+  fsmStateScope = 'all',
+}: LongEntitiesRowProps) {
+  const selectedNodeIds = useSelectedNodeIds();
+  const debouncedZoomRange = useDebouncedZoomRange();
+  const operatorIds = useMemo(() => [...selectedNodeIds], [selectedNodeIds]);
+  const window =
```
cmatzenbach: This is going to shadow the browser global `window`. We should rename into something like `zoomWindow`

ui/packages/@quent/components/src/timeline/TimelineSettingsPopover.tsx
```diff
@@ -0,0 +1,23 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Settings } from 'lucide-react';
+import { Popover, PopoverContent, PopoverTrigger } from '../ui/popover';
+
+export function TimelineSettingsPopover() {
```
cmatzenbach, line 7: I'm assuming this component will be used at a later date? Do we have a plan for it? Seems very generic and not being used anywhere

ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
cmatzenbach: This isn't related to this PR but I happened to spot it while looking at the code, we should remove that extra `>` at the end as I believe that invalidates the tailwind class

ui/packages/@quent/components/src/gantt-chart/utils.ts
```diff
@@ -20,25 +20,45 @@ export function clipRectByRect(target: GanttRect, bounds: GanttRect): GanttRect
...
     }
-    if (row === rowEndMs.length) {
-      rowEndMs.push(entry.endMs);
-    } else {
-      rowEndMs[row] = Math.max(rowEndMs[row], entry.endMs);
-    }
+
+    if (row === rows.length) rows.push([]);
+    rows[row]!.splice(insertionIndex, 0, entry);
     entry.rowIndex = row;
```
cmatzenbach: Do we have an immutability rule?  Wasn't sure, but if so this would violate it

ui/packages/@quent/components/src/timeline/TimelineSettingsPopover.tsx
johallar: Yeah you got it, configurable long entity threshold etc. will go here later in this series of PRs. Moved to component so its easy to exclude for now

ui/packages/@quent/components/src/gantt-chart/utils.ts
johallar: Yeah probably best we don't mutate in here, refactoring

cmatzenbach (approved):

PR #527, opened 2026-08-03
fix(ui): create map and pass down for legend formatter

cmatzenbach:
Quickfix follow-up to PR #471 - creates statQuantitySpecs map and passes down to legend for formatting


johallar (approved):

johallar (changes requested):
I think the audit failure was fixed in master, try merging that maybe before using the override

cmatzenbach:
> I think the audit failure was fixed in master, try merging that maybe before using the override

It hasn't been fixed yet, I just tested on latest `main` and running `audit` there shows the failure. I think this is the correct fix since the package in question is a transient dependency, but lmk what you think.

johallar:
> > I think the audit failure was fixed in master, try merging that maybe before using the override
> 
> It hasn't been fixed yet, I just tested on latest `main` and running `audit` there shows the failure. I think this is the correct fix since the package in question is a transient dependency, but lmk what you think.

Dang this package has come up multiple times then. In any case I think we should prefer updating pnpm-lock.json within the dependencies semver range, instead of the explicit "overrides" section wherever possible. 

`pnpm update brace-expansion` will do it

johallar (approved):

PR #528, opened 2026-08-03
fix(ui): create independent play-head indicator

### LINKED ISSUE #487: refactor(ui): implement independent play-head indicator
Re-using the xaxis pointer for the playhead causes some glitchy behavior if the user is also mousing over the timelines when playing through the data flow feature. 

Implement a separate indicator for the data flow play head to avoid the clash

cmatzenbach:
# Description

Creates a separate play-head line indicator so that it does not interfere with the current x-axis pointer that shows where the user is currently focused on the timelines. 

## Related Issues

Closes #487 

## Testing

1. Start the data flow player
2. Ensure a new solid line plays over the timelines for the data flow
3. Hover over the timelines, ensuring a separate dotted line appears to show current hover state
4. Ensure both lines function independently and do not interfere with one another

## Screenshots

https://github.com/user-attachments/assets/8ff100fb-0331-4f27-b547-92f0ab736512


ui/packages/@quent/components/src/lib/timeline.utils.ts
```diff
@@ -512,6 +512,28 @@ export function hideSyncedPointer() {
   broadcastHidePointer(null);
 }
 
+type PlayheadListener = (timestampMs: number | null) => void;
```
johallar: Could we use an atom for this to stay reactive without setting up a custom pub/sub system?

ui/packages/@quent/components/src/timeline/Timeline.tsx
```diff
@@ -373,12 +376,14 @@ export function Timeline({
...
   const opts = useMemo(() => ({ renderer: 'svg' }) as Opts, []);
 
-  const { handleChartReady } = useChartConnect({
+  const { handleChartReady, instanceRef } = useChartConnect({
     durationSeconds,
     chartGroup: CHART_GROUP,
     onReady: onChartReady,
   });
 
+  const playheadPixelX = usePlayheadLinePixel(instanceRef, 0, readyTick);
```
johallar: I'm on the fence about this living in `Timeline`. To keep this component as generic as possible, I think i'd prefer this stay in ResourceTimeline. 

Worth a quick experiment I think. If it's too much plumbing then I could see a playhead being part of this component, but we should confirm that this fails gracefully if the Timeline component is used elsewhere without any kind of playhead hooked up to it.

ui/packages/@quent/components/src/lib/timeline.utils.ts
cmatzenbach: Yes, we can go ahead and move that to an atom instead of keeping the pub/sub block, that should make it cleaner.

One caveat though - the ECharts event listeners (dataZoom, finished) in `usePlayheadLinePixel` would still be necessary regardless, since the pixel position goes stale on zoom/resize even when the timestamp hasn't changed - that part of the complexity doesn't go away with the atom approach.

ui/packages/@quent/components/src/timeline/Timeline.tsx
cmatzenbach: No, that's a good idea, and ended up being a fairly easy move. All we needed to do was add an optional `onReady` prop to `Timeline`, so anyone using it without playhead hooked up is unaffected.

ui/packages/@quent/components/src/lib/timeline.utils.ts
johallar: Makes sense, I just didn't want to set up the custom pub/sub + global to track if atoms do work for our purposes

ui/packages/@quent/components/src/timeline/Timeline.tsx
```diff
@@ -282,6 +285,8 @@ export function Timeline({
   }, [gridOptions, minZoomSpanPct, xAxisOptions, yAxisOptions, seriesOptions]);
 
   const isDraggingRef = useRef(false);
+  const onReadyRef = useRef(onReady);
```
johallar: Why store this in a ref? Can we just call onReady(instance) below on L366?

cmatzenbach: I was thinking `onChartReady` only runs once per chart instance, but looking at it further, you're right, it's recreated on each render, so the ref is unnecessary.

johallar (approved):
🙇 ty for all the updates

PR #531, opened 2026-08-04
feat(ui): higher level operators

### LINKED ISSUE #49: Higher-level plan operator filtering

johallar:
# Description
Allows crossfiltering by "higher level" operators. In practice with query engines this is usually a logical plan, that will then filter by all related physical plan operators. 

## Related Issues
Fixes #49 

## Testing
* All "lower level" operator filtering should work as before, although you can still only select 1 operator at a time

* Select a node from a higher level/logical plan in the DAG view
* Any "related" nodes from child DAG trees should be selected. Navigate to the sub plans and verify they are selected
* The node info panel should list ALL selected dag nodes (logical plan node, and related physical plan nodes)
* operator timelines should highlight all related nodes from physical plans (in simulated data this means there should be an operator selected in each worker, timelines should be filtered accordingly)
* Filter badge in the query toolbar should show the logical node only
* Operator table should filter to all rows associated with that top level operator

## Screenshots
https://github.com/user-attachments/assets/7f6dc396-b8a9-4fd0-a3a8-587c65406a42


ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
```diff
@@ -291,7 +291,7 @@ export function OperatorGanttChart({
         if (params.seriesName !== 'operator-span') return;
         const op = operators[params.dataIndex];
         if (!op) return;
-        if (selectedNodeIds.has(op.operatorId)) {
+        if (selectedNodeIds.size === 1 && selectedNodeIds.has(op.operatorId)) {
```
johallar, line 195: Prep for multi selecting operators, just remove it from the set instead of checkng for size===1 and deselecting everything do setSelectedNodeIds(new Set(selectedNodeIds.remove(op.operatorId)));

ui/packages/@quent/components/src/services/query-plan/query-bundle-transformer.ts
```diff
@@ -88,6 +124,10 @@ export const getPlanDAG = (
 
   const nodeMap = new Map<string, DAGNode>();
   const edges: DAGEdge[] = [];
+  const operators = Object.values(bundle.entities.operators).filter(
+    (operator): operator is Operator => operator !== undefined
```
johallar: Learn about this ts syntax

ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx
johallar: Removing this enables multiple operator selectors from the operator gantt chart which is cool but should come in a follow up because other things need to be updated to account for > 1 operator selection

ui/packages/@quent/components/src/services/query-plan/query-bundle-transformer.ts
johallar: cool:
```
This is a TypeScript type predicate (or type guard):

(operator): operator is Operator => operator !== undefined
Breakdown:

operator — callback parameter
operator is Operator — return type predicate
operator !== undefined — runtime boolean check
It tells TypeScript: “When this function returns true, treat operator as an Operator.”

Here, Object.values(...) likely produces:

(Operator | undefined)[]
After .filter(...), TypeScript narrows the result to:

Operator[]
Without the predicate, TypeScript may retain the possible undefined type.

It does not validate that the value has the full Operator structure—it only checks for undefined. The predicate is valid because the original type already guarantees every defined value is an Operator.

A reusable equivalent would be:

function isOperator(value: Operator | undefined): value is Operator {
  return value !== undefined;
}
Unlike as Operator, a predicate narrows conditionally based on a runtime check rather than unconditionally asserting a type.
```

ui/packages/@quent/components/src/timeline/ResourceTimeline.tsx
```diff
@@ -114,10 +115,10 @@ export function ResourceTimeline({
...
   if (operatorTimelineData !== undefined) {
     lastOverlayRef.current = operatorTimelineData;
-  } else if (!operatorId) {
+  } else if (!hasOperatorFilter) {
     lastOverlayRef.current = undefined;
   }
-  const overlayPreloadedData = operatorId
+  const overlayPreloadedData = hasOperatorFilter
     ? (operatorTimelineData ?? lastOverlayRef.current)
     : undefined;
```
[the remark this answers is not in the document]
johallar: https://github.com/rapidsai/quent/pull/531/commits/f69de768936a1229a9699165d34ceadd4f64135f

ui/packages/@quent/components/src/services/query-plan/query-bundle-transformer.ts
```diff
@@ -42,6 +52,32 @@ const getNodeEntity = (bundle: QueryBundle<EntityRef>, id: string): DAGNode | un
...
+      const childId = stack.pop()!;
+      if (childId === operator.id || related.has(childId)) continue;
+      related.add(childId);
+      stack.push(...(childrenByParentId.get(childId) ?? []));
+    }
+    relatedById.set(operator.id, [...related]);
+  }
+
+  return relatedById;
+};
```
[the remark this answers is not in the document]
johallar: https://github.com/rapidsai/quent/pull/531/commits/f69de768936a1229a9699165d34ceadd4f64135f

ui/packages/@quent/hooks/src/atoms/timeline.ts
```diff
@@ -28,7 +28,7 @@ export function timelineCacheKey(params: TimelineCacheParams): string {
   return [
     params.resourceId,
     params.resourceTypeName,
-    params.operatorId ?? '',
+    JSON.stringify([...new Set(params.operatorIds ?? [])].sort()),
```
cmatzenbach: Not the biggest deal in the world, but you did make a helper that does this below in `timeline.utils.ts` - if there's no circular dependency issues, would be preferable to use the new helper.

cmatzenbach (approved):
One slight nitpick, other than that code looks good and works great!

PR #534, opened 2026-08-04
refactor(ui): data flow ui/ux cleanup

### LINKED ISSUE #486: refactor(ui): "follow measure" placeholder should be migrated to default option
Instead of having "follow measure" as a placeholder and implicit default value, it should be an actual select option (selected by default)

Reference: https://github.com/rapidsai/quent/pull/477#issuecomment-5121480387

### LINKED ISSUE #488: bug(ui): data-flow bars distort DAG layout
DAG Nodes change height when data flow is turned on/off, we should re-layout to account for it or otherwise keep the layout consistent (see wiggly arrows below):

Image

### LINKED ISSUE #489: chore(ui): data flow grid wrapping improvements
Headers and values can wrap currently which can make it a little hard to parse, explore ways to mitigate the text wrapping. 

note: there can be 'n' measures so I think a table like this may not scale well in the long run, explore other layouts or locations for the data flow table. 

Image

### LINKED ISSUE #532: refactor(ui): move task activity out of stats panel

### LINKED ISSUE #533: refactor(ui): don't show data flow bars during inactivity

cmatzenbach:
# Description

This takes care of several items related to data flow cleanup:

- Makes "Follow measure" within the "Bar labels" dropdown a selectable (and default) option, rather than just placeholder text - closes #486 
- Hides data flow bars on DAG nodes when empty (rather than displaying empty bars) - closes #533 
- Improves DAG layout to ensure it's reactive to whether or not data flow is active - closes #488 
- Adds tabs to `DAGNodeInfoPanel` and moves data flow activity to second tab, so it is not conditionally rendered on top of the node statistics - closes #532 
- Improves data flow activity grid wrapping (removes wrapping, adds x-min-width and scrolls below that) - closes #489*  

*Arguably this does not close out #532 completely, since part of the ticket was to explore other layout options, but I think we can break that into a second ticket, as for now this fixes the poor text wrapping.

## Related Issues

Closes #486, #533, #488, #532, #489

## Testing

Issue #486:
1. Ensure "Follow measure" is now the default option
2. Ensure the "Follow measure" option works as expected - change the "Flow measure" field and ensure "Bar labels" reacts accordingly

Issue #533
1. Ensure flow bars are no longer visible when there is no data to display and phantom bars are no longer rendered
2. Ensure bars are rendered and display data when there is activity for that node

Issue #488 
1. Verify DAG nodes and edges respond appropriately when data flow is enabled and disabled
2. Make sure edges no longer get weird positioning, as posted through screenshot in issue

Issue #532 
1. Verify `DAGNodeInfoPanel` now has two tabs and that you can switch between them
2. Ensure tabs _do not_ appear when data flow is disabled
3. Verify data flow state activity is now moved to second tab
4. Verify display text appears when there is no activity for the selected node

Issue #489 
1. Reduce the size of the left panel, make sure table becomes horizontally scrollable
2. Verify data flow activity table no longer wraps, and no longer distorts column headers 

## Screenshots

Issue #486 

https://github.com/user-attachments/assets/944daf72-a677-4868-a5ac-2473d6f795ee

Issue #533:

Screenshot_20260805_105821

Issue #488:

Data flow enabled:
Screenshot_20260805_110022

Data flow disabled:
Screenshot_20260805_110101

Issue #532:
image

Issue #489:

https://github.com/user-attachments/assets/d189a726-fc73-4562-9441-2c8e3caeb521


johallar:
for #533 I think we may want to go the placeholder route, instead of removing completely. With logic something like: if there is no per operator data then get rid of the bars and the space at the bottom of the node where the bars would go. If there is operator data but just none for the current operator/timestamp then show some kind of "no data" placeholder.

I had left some comments earlier in the main ticket, all up for discussion of course https://github.com/rapidsai/quent/issues/433

> Showing a placeholder text in the bar or something might be a good option as well, I think the problem is its not clear what the bars are if there's no data displayed.

ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
```diff
@@ -67,55 +117,43 @@ export const DAGNodeInfoPanel = ({
...
+            value={activeTab}
+            onValueChange={setActiveTab}
+            className="border-t overflow-visible"
+          >
+            <TabsList className="h-7 py-0 px-1 rounded-none">
+              <TabsTrigger value="stats" className="text-xs px-2 py-0.5">
+                Stats
+              </TabsTrigger>
+              <TabsTrigger value="activity" className="text-xs px-2 py-0.5">
+                Activity
```
johallar: What do you think about label being "Data Flow" or similar? 

I think we should also potentially switch to this tab whenever the play button is clicked, or indicate that this tab is "active" in some way.

ui/packages/@quent/components/src/dag/layout.ts
```diff
@@ -6,21 +6,19 @@ import type { Node, Edge } from '@xyflow/react';
 import { DAG_LAYOUT_DIRECTION, type DagLayoutDirection } from '@quent/utils';
 
 export const NODE_LAYOUT_WIDTH = 200;
-const NODE_LAYOUT_HEIGHT = 60;
+export const NODE_LAYOUT_HEIGHT = 60;
+// NodeFlowBar fixed height
+export const FLOW_BAR_HEIGHT = 44;
```
johallar: We should use this to actually size the flow bars to future proof this, if the height changes here or there then the layout will get thrown off again I believe

ui/packages/@quent/components/src/dag/DataFlowMatrix.tsx
```diff
@@ -71,60 +71,69 @@ export const DataFlowMatrix = ({
           · {measureDecl?.display_name ?? frame.measure} during this bin
         </span>
       </div>
-      <table className="mt-1 text-xs w-full border-separate border-spacing-0">
+      <table className="mt-1 text-xs min-w-full border-separate border-spacing-0">
```
johallar: I _think_ you can just apply the `whitespace-nowrap` here instead of on every sub-element, unless that messes some other things up.

ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
johallar: This would align it with the "settings" title anyways

johallar (approved):
Couple mostly minor things, I think we should iterate on the data flow bars placeholder vs hidden thing and switching to the data flow tab when playing. Approving so you're not held up, love all the tweaks!

PR #537, opened 2026-08-05
fix(io): make event import iteration fallible

johanpel:
# Description

Make importers yield `Result` items so read, framing, and decoding failures are not mistaken for EOF. Recover when the next record boundary is known, otherwise terminate after returning the error.

Generated model and query-engine consumers now propagate item errors. Simulator bridges temporarily collect streams until a follow-up introduces the fallible streaming interface in #520.

## Related Issues

Split off as a small incremental change from #520

_Written by Codex._


dhruv9vats (approved):
LGTM, thanks.

PR #540, opened 2026-08-06
feat(nvtx): capture wide-char (*W) NVTX surface as UTF-8

9prady9:
## Summary

Implements the NVTX `*W` (wide-char / Unicode) callback surface on Linux, where `wchar_t` is 32-bit UTF-32. Each code unit maps directly to a Unicode scalar value via `char::from_u32`; the result enters the event stream as a plain `NvtxMessage::String` — identical to the ASCII surface from the downstream perspective.

- `copy_wchar` iterates a NUL-terminated `wchar_t*` and builds an owned UTF-8 `String`
- All five CORE `*W` callbacks (`MarkW`, `RangeStartW`, `RangePushW`, `NameCategoryW`, `NameOsThreadW`) are now full implementations instead of warn-once stubs
- `NVTX_MESSAGE_TYPE_UNICODE` in event attributes (`nvtxDomainRangePushEx` etc.) is handled in `decode_message` alongside the existing ASCII and registered-string cases
- Only `nvtx-injection` is touched — `nvtx-events`, `nvtx-bridge`, and `nvtx-analyzer` are unchanged since wide strings arrive as `NvtxMessage::String` downstream

## Test plan

- [x] `pixi run cargo test -p nvtx-injection` — 30 tests pass (5 new: `copy_wchar` for ASCII/Unicode/null, `mark_w`, `range_push_w`, `range_start_w`, `unicode_message_type_in_event_attributes_is_decoded`)
- [x] `pixi run cargo test -p nvtx-analyzer -p nvtx-example` — all pass, no downstream changes

### DIFF
--- a/integrations/nvtx/injection/src/callbacks.rs
+++ b/integrations/nvtx/injection/src/callbacks.rs
@@ -14,7 +14,6 @@
 
 use std::os::raw::{c_char, c_int, c_void};
 use std::panic::AssertUnwindSafe;
-use std::sync::atomic::{AtomicBool, Ordering};
 
 use crate::bindings::{
     nvtxDomainHandle_t, nvtxEventAttributes_t, nvtxRangeId_t, nvtxResourceAttributes_t,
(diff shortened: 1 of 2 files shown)


johanpel:
Is there a reason we couldn't just support the wide char APIs, e.g. using: https://github.com/VoidStarKat/widestring-rs to convert?

I have doubts about this work-around since the sentinel values clash with NVTX' spec (e.g. [id uniqueness](https://github.com/NVIDIA/NVTX/blob/dev/c/include/nvtx3/nvToolsExt.h#L871)) and it adds complexity that doesn't need to exist if we were to simply support those APIs.

9prady9:
> Is there a reason we couldn't just support the wide char APIs, e.g. using: https://github.com/VoidStarKat/widestring-rs to convert?
> 
> I have doubts about this work-around since the sentinel values clash with NVTX' spec (e.g. [id uniqueness](https://github.com/NVIDIA/NVTX/blob/dev/c/include/nvtx3/nvToolsExt.h#L871)) and it adds complexity that doesn't need to exist if we were to simply support those APIs.

Its not used anywhere in our current targets and it adds a lot more code which isn't used anywhere. Adding this when there is a consumer is easier with the tools we have now.

> I have doubts about this work-around since the sentinel values clash with NVTX' spec (e.g. [id uniqueness](https://github.com/NVIDIA/NVTX/blob/dev/c/include/nvtx3/nvToolsExt.h#L871)) and it adds complexity that doesn't need to exist if we were to simply support those APIs.

I didn't understand what you mean. Fix is to address the current behavior of no matching start/push for a end/pop, respectively.

johanpel:
> Its not used anywhere in our current targets and it adds a lot more code which isn't used anywhere. Adding this when there is a consumer is easier with the tools we have now.

If the wide-character APIs are not used anywhere, what requires this workaround now?
 
> I didn't understand what you mean. Fix is to address the current behavior of no matching start/push for a end/pop, respectively.

Is the missing captured start/push event caused by something outside our control, or by our choice of not dispatching an event for the *W callbacks?

9prady9:
> > Its not used anywhere in our current targets and it adds a lot more code which isn't used anywhere. Adding this when there is a consumer is easier with the tools we have now.
> 
> If the wide-character APIs are not used anywhere, what requires this workaround now?

Without the surface API to capture these, the push/pops incorrectly get hold of any other push/pop and/or silently don't say anything. We surface a warning when we encounter these by providing the entry points at least.

> 
> > I didn't understand what you mean. Fix is to address the current behavior of no matching start/push for a end/pop, respectively.
> 
> Is the missing captured start/push event caused by something outside our control, or by our choice of not dispatching an event for the *W callbacks?

These are actual nvtx API calls out of our control right, handling them incorrectly is a bug in this crate not the other way around.

johanpel:
> handling them incorrectly is a bug in this crate not the other way around.

Indeed, the bug is that we're not emitting events we're supposed to be emitting. Why is using sentinel values and an additional thread local stack to filter out events better than actually emitting those events?

9prady9:
> > handling them incorrectly is a bug in this crate not the other way around.
> 
> Indeed, the bug is that we're not emitting events we're supposed to be emitting. Why is using sentinel values and an additional thread local stack to filter out events better than actually emitting those events?

You're right, I got it wrong. Updated the PR — `*W` callbacks are now full implementations converting UTF-32 to UTF-8 via `char::from_u32`. Sentinel and suppression machinery gone.

johanpel (approved):
Nice, thanks for adding wide char support @9prady9!

PR #544, opened 2026-08-07
feat(codegen): add default-on NVTX capture

9prady9:
## Summary
- make NVTX capture default-on for `model!` models emitted through the CXX bridge
- allow `model!` callers to disable CXX NVTX capture with `nvtx: false`
- keep the generated CXX context's NVTX observer pipeline alive and install the process-global injection hook
- add the NVTX runtime dependencies to generated CXX bridge consumers

## Scope
- this PR is intentionally limited to the CXX codegen path needed by `sirius-db/sirius#1439`
- NVTX is not exposed through the YAML model format
- schema-driven unification across Rust, CXX, and Python is tracked in #618

## Reconciliation
- rebase the codegen change onto current `main` (`e99109e`)
- drop the superseded mock UI commit; the NVTX UI is now provided by #586
- move the legacy C++ bridge dependency update to the path introduced by #627

This supplies the remaining Quent codegen support needed by sirius-db/sirius#1439.

## Testing
- `pixi run cargo check -p quent-bridge -p quent-qe-bridge --locked`
- `pixi run cargo test -p quent-codegen -p quent-model -p quent-model-macros -p quent-yaml --locked`

### DIFF
--- a/crates/codegen/src/cxx_bridge.rs
+++ b/crates/codegen/src/cxx_bridge.rs
@@ -478,6 +478,41 @@ fn emit_context_bridge(
     let build_wraps: Vec<&TokenStream> = observer_fields.iter().map(|o| &o.wrap).collect();
     let field_inits: Vec<syn::Ident> = observer_fields.iter().map(|o| o.field.clone()).collect();
 
+    let nvtx_imports = model.nvtx.then(|| {
+        quote! {
+            use nvtx_bridge;
+            use nvtx_injection;
+        }
+    });
+    let nvtx_struct_field = model.nvtx.then(|| {
+        quote! {
+            _nvtx_pipeline: Option<Box<dyn std::any::Any + Send + Sync>>
+        }
+    });
+    let nvtx_tuple_binding = model.nvtx.then(|| quote! { _nvtx_pipeline, });
+    let nvtx_noop_value = model.nvtx.then(|| quote! { None, });
+    let nvtx_active_setup = model.nvtx.then(|| {
+        quote! {
+            let pipeline = inner
+                .observer::<nvtx_bridge::NvtxEventEntity>(&options)
+                .await
+                .map_err(|e| e.to_string())?;
+            let sender = pipeline.sender();
+            let context_id = id;
+            // The injection hook is process-global and one-shot. A later
+            // generated context keeps its own pipeline alive but cannot replace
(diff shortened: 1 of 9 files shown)


crates/codegen/src/cxx_bridge.rs
```diff
@@ -478,6 +478,41 @@ fn emit_context_bridge(
...
+            // generated context keeps its own pipeline alive but cannot replace
+            // the first context's capture destination.
+            let _ = nvtx_injection::install_hook(move |event| {
+                sender.emit(context_id, nvtx_bridge::NvtxEventEntity::from(event))
+            })
+            .ok();
+            let _nvtx_pipeline: Option<Box<dyn std::any::Any + Send + Sync>> =
+                Some(Box::new(pipeline));
+        }
+    });
```
johanpel, line 512: Why should we scope this to the CXX bridge only? If this would be moved to the model-macros crate then Rust and Python clients can also leverage it.

crates/yaml/src/lib.rs
```diff
@@ -28,6 +29,8 @@ pub use diag::{Diagnostic, Diagnostics, Origin};
 pub struct Parsed {
     /// The parsed schema.
     pub schema: Schema,
+    /// Whether CXX instrumentation generated for this model should capture NVTX.
+    pub nvtx: bool,
```
johanpel: If this remains scoped to the CXX bridge we shouldn't add it to the YAML DSL yet since CXX bridge code generation isn't schema-based yet, so this would set the wrong expectations.

9prady9: In that case, we should remove it for now and make it CXX only for now.

Doing it for all three is much larger change and I don't want to rush it now while we are trying to get the nvtx in UI. It should be part of the already created follow up #618

crates/codegen/src/cxx_bridge.rs
9prady9: https://github.com/rapidsai/quent/pull/544#discussion_r3870499858

crates/yaml/src/lib.rs
9prady9: in fact it seems it fits 618 more. - extended 618 scope accordingly.

johanpel (approved):

PR #561, opened 2026-08-10
fix(ui): resolve nanoid audit advisory

9prady9:
Pins the transitive nanoid dependency to 3.3.17 to resolve GHSA-2v37-7h3g-55p8, which is currently failing the shared UI Audit job across PRs.

Verification:
- pixi run pnpm --dir ui install --frozen-lockfile
- pixi run pnpm --dir ui audit --audit-level=high

### DIFF
--- a/ui/pnpm-lock.yaml
+++ b/ui/pnpm-lock.yaml
@@ -58,6 +58,9 @@ catalogs:
       specifier: ^6.0.3
       version: 6.0.3
 
+overrides:
+  nanoid: 3.3.17
+
 importers:
 
   .:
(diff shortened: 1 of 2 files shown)


joosthooz (approved):
Ok to merge, but just checking is it needed to have this as an override?

9prady9:
> Ok to merge, but just checking is it needed to have this as an override?

Honestly thats what codex came up with for the failure - its a UI audit job, so I can definitely wait if you want to let UI team look into it.

dhruv9vats (approved):

Issue #373, opened 2026-07-13
Phase 3: See It in the UI (Server & Swim Lanes)


9prady9:
**What this delivers:** the reconstructed NVTX ranges become visible and interactive in the Quent UI — completing the end-to-end slice.

**In scope**
- HTTP endpoint(s) that serve the reconstructed ranges, marks, and statistics
- Swim-lane timeline view grouped by domain and thread, on a shared zoom/pan time axis
- Range colors, instant marks, named-thread lane labels
- Hover detail and filtering by domain/category

**Out of scope (later phases)**
- Running alongside Nsight/other tools
- Validation against real GPU workloads
- Decoding payload/extension data

**Deliverables**
- Analyzer HTTP endpoint(s) following the existing server layout
- Swim-lane UI for NVTX ranges

**Done when**
- [ ] Endpoint(s) return ranges, marks, and statistics as typed views
- [ ] Ranges render as nested spans in swim lanes grouped by domain and thread
- [ ] Marks show as instant events; NVTX colors honored (with a fallback palette); named threads label their lanes
- [ ] Hovering a range shows its message, duration, thread, domain, and category
- [ ] User can filter ranges by domain and category

---
Part of #76 · Phase 3 of 5

9prady9:
UI changes will likely land in #586

PR #562, opened 2026-08-10
feat(nvtx): add UI viewport contracts

9prady9:
## Summary

Adds the Rust-owned NVTX exchange contract used by the server and Timeline UI.

The new `nvtx-ui` crate:

- defines catalog, selection, viewport, lane, item, color, and statistics types
- validates and canonicalizes domain/category selections
- converts reconstructed NVTX models into deterministic, UI-ready viewport responses
- owns filtering, lane grouping, depth, ordering, colors, clipping, and statistics

Related to #373 

## Verification

```text
pixi run cargo test -p nvtx-ui --locked
pixi run cargo clippy -p nvtx-ui --all-targets --locked -- -D warnings
pixi run cargo fmt --all -- --check
```

### DIFF
--- a/integrations/nvtx/ui/src/lib.rs
+++ b/integrations/nvtx/ui/src/lib.rs
@@ -12,14 +12,17 @@ use std::error::Error;
 use std::fmt;
 
 use nvtx_analyzer::{NvtxColor, NvtxModel, NvtxSpan, SpanId, SpanKind};
+use quent_time::{TimeUnixNanoSec, to_nanosecs, to_secs, to_secs_relative};
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 
 /// Stable metadata for one NVTX stream.
-#[derive(TS, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(TS, Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct NvtxCatalog {
-    pub trace_start: u64,
-    pub trace_end: u64,
+    /// Trace start in seconds relative to the query start.
+    pub trace_start: f64,
+    /// Trace end in seconds relative to the query start.
+    pub trace_end: f64,
     pub domains: Vec<NvtxCatalogDomain>,
     pub anomalies: NvtxCatalogAnomalies,
 }
@@ -71,15 +74,11 @@ pub struct NvtxCatalogCategory {
     pub name: String,
 }
 
-/// Inclusive viewport bounds in Unix nanoseconds.
-#[derive(TS, Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+/// Inclusive viewport bounds in seconds relative to the query start.
+#[derive(TS, Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
 pub struct NvtxViewportWindow {
-    #[serde(with = "decimal_u64")]
-    #[ts(type = "string")]
-    pub start: u64,
-    #[serde(with = "decimal_u64")]
-    #[ts(type = "string")]
-    pub end: u64,
+    pub start: f64,
+    pub end: f64,
 }
 
 /// One domain's selected categories.
@@ -93,21 +92,21 @@ pub struct NvtxDomainSelection {
 }
 
 /// Request for one atomically-scoped set of lanes and statistics.
-#[derive(TS, Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(TS, Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct NvtxViewportRequest {
     pub viewport: NvtxViewportWindow,
     pub selections: Vec<NvtxDomainSelection>,
 }
 
 /// UI-ready NVTX content for one viewport and selection.
-#[derive(TS, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(TS, Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct NvtxViewportResponse {
     pub viewport: NvtxViewportWindow,
     pub domains: Vec<NvtxDomainLaneGroup>,
     pub statistics: Vec<NvtxRangeStatistics>,
 }
 
-#[derive(TS, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(TS, Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct NvtxDomainLaneGroup {
     pub domain_id: u64,
     pub name: String,
@@ -125,7 +124,7 @@ pub enum NvtxLaneIdentity {
     Marks,
 }
 
-#[derive(TS, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(TS, Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct NvtxLane {
     pub id: String,
     pub label: String,
@@ -141,7 +140,7 @@ pub enum NvtxRangeKind {
     StartEnd,
 }
 
-#[derive(TS, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(TS, Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct NvtxRangeItem {
     pub message: String,
     pub domain_id: u64,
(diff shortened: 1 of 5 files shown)


integrations/nvtx/ui/src/lib.rs
```diff
@@ -0,0 +1,1134 @@
...
+
+/// Reconstruction events that could not be represented faithfully in the model.
+#[derive(TS, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub struct NvtxCatalogAnomalies {
+    pub orphan_range_ends: u64,
+    pub orphan_range_pops: u64,
+    pub orphan_resource_destroys: u64,
+    pub reused_range_ids: u64,
+    pub reused_resource_handles: u64,
+    pub total: u64,
```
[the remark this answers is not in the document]
9prady9: https://github.com/rapidsai/quent/pull/562/commits/cefd560316dc8aa6d1c4edfa19b362baa6c53a5c

integrations/nvtx/ui/src/lib.rs
```diff
@@ -0,0 +1,975 @@
...
+            parent = model
+                .spans()
+                .get(parent_index)
+                .and_then(|span| span.kind.parent());
+            remaining -= 1;
+        }
+        *slot = depth;
+    }
+    depths
+}
```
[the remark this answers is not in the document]
9prady9: https://github.com/rapidsai/quent/commit/cefd560316dc8aa6d1c4edfa19b362baa6c53a5c

9prady9:
### Viewport contracts: purpose and data flow

The contracts in this PR form one pipeline:

`NvtxModel -> NvtxCatalog -> NvtxViewportRequest -> NvtxViewportResponse`

In short:

- **Catalog:** What exists in the trace?
- **Request:** What does the user want to inspect now?
- **Response:** What should the UI render and report?

Contract by contract:

1. **`NvtxCatalog`** — stable trace-wide metadata used to initialize filters and navigation without reconstructing the whole viewport.
2. **`NvtxCatalogAnomalies`** — reports orphan closes, reused identifiers, and whether reconstruction is faithful, so the UI can warn when rendered spans or statistics may not tell the full story.
3. **`NvtxCatalogDomain`** — describes each domain and its available threads and categories.
4. **`NvtxCatalogThread`** — identifies the process/thread choices available within a domain.
5. **`NvtxCatalogCategory`** — identifies category choices and their display colors.
6. **`NvtxViewportWindow`** — the requested time interval. Nanosecond values cross JSON as decimal strings to preserve full `u64` precision.
7. **`NvtxDomainSelection`** — one selected domain plus its optional thread and category filters.
8. **`NvtxViewportRequest`** — combines the time window with domain selections and defines the exact slice requested by the client.
9. **`NvtxViewportResponse`** — contains the stable catalog plus the lane groups and statistics for that slice.
10. **`NvtxDomainLaneGroup`** — groups renderable lanes under a domain.
11. **`NvtxLaneIdentity`** — states what a lane represents: a thread at a nesting depth, process ranges, or marks.
12. **`NvtxLane`** — a renderable row containing ranges and marks.
13. **`NvtxRangeKind`** — distinguishes thread ranges from process ranges.
14. **`NvtxRangeItem`** — a single visible range. Observed bounds preserve captured truth; display bounds are clipped to the viewport. An incomplete range has no observed end or duration.
15. **`NvtxMarkItem`** — a point event visible in the requested window.
16. **`NvtxRangeStatistics`** — aggregates visible ranges by domain/category/message. `count` includes incomplete ranges, while duration statistics use completed observed ranges; min/max are optional when no completed duration exists.
17. **`NvtxViewportError`** — explicit validation failures for malformed windows, decimal identifiers, unknown domains, and invalid selections.

Two implementation properties are important:

- Catalog construction and viewport grouping scan the model once, avoiding work proportional to domains × spans.
- `from_model_with_catalog` allows the server to cache and reuse stable catalog metadata across pans and filter changes.

This separation keeps discovery, interaction state, and rendering payloads independent while preserving reconstruction fidelity and lossless timestamps.

integrations/nvtx/ui/src/lib.rs
```diff
@@ -0,0 +1,1357 @@
...
+#[derive(TS, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct NvtxCatalogCategory {
+    pub category_id: u32,
+    pub name: String,
+}
+
+/// Inclusive viewport bounds in Unix nanoseconds.
+#[derive(TS, Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct NvtxViewportWindow {
+    #[serde(with = "decimal_u64")]
```
johallar: javascript has to rely on bigint for nanosecond epoch timestamps, so we designed the resource timelines to give these start/end timestamps in decimal seconds relative to the query start time to get around that. If possible I would like to design NVTX viewport/range starts and ends to be the same. 

Timeline Request:
Image

Timeline response:
Image

integrations/nvtx/ui/src/lib.rs
```diff
@@ -0,0 +1,1567 @@
...
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+/// Stable metadata for one NVTX stream.
+#[derive(TS, Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct NvtxCatalog {
+    /// Absolute origin used to produce every relative-second field.
+    #[serde(skip)]
+    #[ts(skip)]
+    query_start: TimeUnixNanoSec,
```
joosthooz: Why do we skip `query_start`, wouldn't that cause trouble when roundtripping data?

integrations/nvtx/ui/src/lib.rs
```diff
@@ -0,0 +1,1567 @@
...
+    }
+
+    /// Validate and canonicalize a request.
+    pub fn canonicalize_request(
+        &self,
+        mut request: NvtxViewportRequest,
+    ) -> Result<NvtxViewportRequest, NvtxViewportError> {
+        if !request.viewport.start.is_finite()
+            || !request.viewport.end.is_finite()
+            || request.viewport.start < 0.0
```
joosthooz: Is it never possible for a trace to start before the query start, making this negative?

integrations/nvtx/ui/src/lib.rs
```diff
@@ -0,0 +1,1567 @@
...
+            let domain_items = items_by_domain.entry(span.domain).or_default();
+            match span.kind {
+                SpanKind::PushPop { thread_id, .. } => {
+                    domain_items
+                        .thread_lanes
+                        .entry((thread_id, depths[index]))
+                        .or_default()
+                        .push(item);
+                }
+                SpanKind::StartEnd => domain_items.process_ranges.push(item),
```
joosthooz, line 514: Could this result in overlapping ranges, and is that a problem?

integrations/nvtx/ui/src/lib.rs
```diff
@@ -0,0 +1,1567 @@
...
+                        name: thread_names.get(&thread_id).map_or_else(
+                            || model.thread_name(thread_id),
+                            |name| (*name).to_owned(),
+                        ),
+                    })
+                    .collect();
+                threads.sort_by(|left, right| {
+                    left.name
+                        .cmp(&right.name)
+                        .then(left.thread_id.cmp(&right.thread_id))
```
joosthooz, line 313: Cosmetic nit; it looks like this would sort threads with default names like `thread 1` .. `thread 10` like 
```
thread 1
thread 10
thread 2
...
```

joosthooz (commented):
Just some small questions, I'm not sure if the serde thing is an actual problem

integrations/nvtx/ui/src/lib.rs
9prady9: yeah, but there is no real ordering here to those threads than UI arrangement. Unless we want to put them top to bottom going from start time to end of the query time - perhaps that is useful didn' think of that to be honest until now.

integrations/nvtx/ui/src/lib.rs
9prady9: yes, ranges are process wide and can definitely overlap depending on the how the nvtx range API calls are sprinkled around.  Right now, data isn't altered but the current render will render rectangles on top of each other when there is a overlap.

It's not a bug but visually hard to grasp that - the only I can think of to resolve this to again stack them in that lane. @johallar what do you recommend here ?

I am not sure how much a visual cue will help though even if we try to solve this rendering overlap. The user can still get the actual time range info from the tooltip.

integrations/nvtx/ui/src/lib.rs
9prady9: From the way I understand these two, query start is the absolute 0.0 as per UI timeline (as per other telemetry info apart from NVTX events) for a given query, while the trace start is set to the minimum time ticker of the earliest/first nvtx event for this run; so I doubt trace_start can appear before query_start.

integrations/nvtx/ui/src/lib.rs
9prady9: fixing this one.

integrations/nvtx/ui/src/lib.rs
9prady9: Well, that is true only if my assumption that nvtx events are query scoped is true. We might have events before that also in the code. Let me think about this a bit more.

joosthooz (approved):
Great, thanks

cmatzenbach (approved):

integrations/nvtx/ui/src/lib.rs
johallar: Let's discuss tomorrow morning a bit, FE can do some intra-lane stacking or try to visualize overlapping spans in the same lane in some other way if i'm understanding correctly.

johallar (approved):

PR #563, opened 2026-08-10
feat(nvtx): add context server routes

9prady9:
## Summary

- add context-scoped NVTX catalog and viewport endpoints to the query-engine server
- discover the contexts associated with each engine and expose their resource ownership
- cache reconstructed NVTX models and query-relative catalogs for efficient viewport requests
- run telemetry processing on a bounded blocking pool and validate request size and filter limits

The routes use the viewport contracts introduced in #562. A required `query_start` parameter establishes the time origin for catalog and viewport responses.

## Verification

- `pixi run cargo test -p nvtx-server -p quent-query-engine-server`
- `pixi run cargo clippy -p nvtx-server -p quent-query-engine-server --all-targets -- -D warnings`
- `cargo fmt --all -- --check`

### DIFF
--- a/integrations/nvtx/server/src/lib.rs
+++ b/integrations/nvtx/server/src/lib.rs
@@ -36,7 +36,6 @@ const MAX_CONCURRENT_MODEL_TASKS: usize = 4;
 pub struct NvtxImporterError(String);
 
 impl NvtxImporterError {
-    /// Capture an importer error for server-side logging.
     pub fn new(error: impl fmt::Display) -> Self {
         Self(error.to_string())
     }
@@ -99,7 +98,6 @@ struct CachedNvtx {
 }
 
 impl CachedNvtx {
-    /// Pair a reconstructed model with its origin-specific catalog cache.
     fn new(model: NvtxModel) -> Self {
         Self {
             model,
@@ -125,7 +123,6 @@ struct NvtxTaskLimiter {
 }
 
 impl NvtxTaskLimiter {
-    /// Create a limiter allowing at most `max_concurrency` model tasks.
     fn new(max_concurrency: usize) -> Self {
         Self {
             permits: Arc::new(Semaphore::new(max_concurrency)),
@@ -158,7 +155,6 @@ struct NvtxModelCache {
 }
 
 impl NvtxModelCache {
-    /// Create the context model cache and its shared blocking-work limiter.
     fn new(importer: Box<NvtxImporterFn>) -> Self {
         Self {
             models: AsyncCache::builder()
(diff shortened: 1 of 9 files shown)


integrations/nvtx/server/src/lib.rs
```diff
@@ -0,0 +1,616 @@
...
+    tasks: NvtxTaskLimiter,
+}
+
+impl NvtxModelCache {
+    fn new(importer: Box<NvtxImporterFn>) -> Self {
+        Self {
+            models: AsyncCache::builder()
+                .max_capacity(128)
+                .time_to_idle(Duration::from_hours(24))
+                .build(),
```
joosthooz, line 158: This caching might be a problem when a user accesses something before it was uploaded (caching the failure). Or when something was updated after accessing it.

integrations/nvtx/server/src/lib.rs
```diff
@@ -0,0 +1,616 @@
...
+    if !stream_dir.is_dir() {
+        return Ok(None);
+    }
+    let mut stream_entries = std::fs::read_dir(&stream_dir).map_err(NvtxImporterError::new)?;
+    match stream_entries.next() {
+        None => return Ok(Some(Vec::new())),
+        Some(Err(error)) => return Err(NvtxImporterError::new(error)),
+        Some(Ok(_)) => {}
+    }
+    let format = Format::detect(&context_dir)
```
joosthooz, line 75: Claude says: `Format::detect` runs against `context_dir` and returns the extension of the first file found in any subdirectory, so a context with mixed-format streams selects the wrong codec for NVTX.

integrations/nvtx/server/src/lib.rs
9prady9: IIUC, quent renders archived data/assets by downloading or streaming the information based on viewports being requested. The data the viewer shows is immutable. Yes, the collector can allow mutability but that does not carry forward the mutability towards quent server/viewer, it purely exists at the data source level which isn't exposed via quent. The caching we have here should help alleviate perf issues when server and data source (telemetry) are physically not at the same location - often the case I would think - removes repeated: storage reads, de-serilization, nvtx model reconstruction and then deriving the catalog. 

Under the above lifecycle, both present and absent NVTX results are immutable and safe to cache. Import errors are not cached.

Supporting a viewer over actively collected telemetry would require separate invalidation and refresh semantics.

integrations/nvtx/server/src/lib.rs
9prady9: Serialization format is fixed per Quent context: the generated context uses one exporter provider for all entity streams, and mixed-format context artifacts are not valid.

Until the day we allow mixed formatting for the providers where each provider has different formats, this wouldn't be a problem.

9prady9: nevertheless handling this more strictly I believe, will push soon.

9prady9: Check this https://github.com/rapidsai/quent/pull/563/commits/583d6bf0f4aa33fe2b8abd349f41974b12d390c4

integrations/nvtx/server/src/lib.rs
9prady9: https://github.com/rapidsai/quent/pull/563/commits/a7bb36b6fea71f4fc4d6e09c6c54110c4a09975b

Should handle this more narrowly now. It still the viewer/client side responsibility to request for refresh once it got "not found"

joosthooz (approved):
Thank you!

9prady9:
> Thank you!

Thank you for the thorough reviews :)

johallar (approved):

PR #564, opened 2026-08-10
feat(ui): entities viewer and detail panel

cmatzenbach:
# Description

This brings in the Entities table originally mocked out by @johanpel, with numerous UI/UX improvements. In addition, it adds an entities detail tab, which can be accessed either by clicking on a row in the entities table or by clicking on an entity in the timeline section. 

Entities table changes:
- Added `QueryToolbar` to the top of the view, so that current filters can be seen/cleared/etc
- Revised layout of the dropdowns
- Improved styling 
- Improved verbiage
- Fixed default text rendering
- Added sort options to all columns
- Added a "Sequence" row, which gives a preview of all the states that particular entity goes through

Entities Detail tab:
- Fix styling
- Show total timespan, dominant state (with %), and bar showing % of time spent in each state with tooltip
- Adds chart showing data consumption/flow per state
-  Colors each state detail section to match up with timelines
- Adds bar to each state detail section showing how much time was spent in that state

## Testing

1. Open Entities tab, ensure the table works as expected
2. Click on an entity row, ensure the entity details panel pops up and shows all relevant information
3. Click on an entity from the timeline view, ensure it is highlighted 
4. Ensure entity details panel pops up after clicking on an entity from timeline
5. Verify all information is displayed

## Screenshots

https://github.com/user-attachments/assets/6cd86f9d-9e9e-48a5-984f-08ead8d169aa

https://github.com/user-attachments/assets/29c4dff8-3fbf-40f6-b6e2-64decc9e52b2


examples/simulator/server/ts-bindings/EntityListItem.ts
johallar, line 1: We shouldn't be committing generated typescript bindings anymore after https://github.com/rapidsai/quent/pull/255

ui/packages/@quent/components/src/fsm-chart/FsmCapacityChart.tsx
```diff
@@ -0,0 +1,142 @@
...
+import EChartsReactCore from 'echarts-for-react/lib/core';
+import type { FsmTransition } from '@quent/utils';
+import { formatBytes } from '@quent/utils';
+import { echarts } from '../lib/echarts';
+import { useChartResize } from '../lib/useChartResize';
+import { useTimelineEchartsTheme } from '../timeline/timelineEchartsTheme';
+
+const CHART_HEIGHT = 90;
+
+function isBytesStat(name: string): boolean {
```
johallar: Duplicates `formatters.ts` code

ui/src/components/entities-table/EntitiesToolbar.tsx
```diff
@@ -0,0 +1,232 @@
...
+            className="w-28"
+            value={filters.windowEnd}
+            onChange={value => onFiltersChange({ windowEnd: value })}
+          />
+        </div>
+
+        {/* Display settings */}
+        <div className="flex items-end gap-2.5">
+          <FieldWrapper label="Sort" className="w-44">
+            <SelectField
```
johallar: I think we want the table header sort controls to be a BE sort, instead of sorting the current page, and then remove this from the toolbar. If that means the only sortable column is the fsm span column I think that's fine also.

ui/src/components/entities-table/EntitiesToolbar.tsx
```diff
@@ -0,0 +1,232 @@
...
+              onValueChange={value =>
+                onFiltersChange(
+                  { sortDir: (value as SortDir | null) ?? 'Desc' },
+                  { preserveSelection: true }
+                )
+              }
+              triggerClassName="h-8"
+            />
+          </FieldWrapper>
+          <PageSizeField
```
johallar: I think this may be better placed with the other page controls in the table footer

examples/simulator/server/ts-bindings/EntityListItem.ts
cmatzenbach: Yeah good point, not sure how that made it in there, might've been from starting from Johan's original branch.  Removed.

ui/src/components/entities-table/EntitiesToolbar.tsx
johallar, line 1: Text is a bit inconsistent between the controls here, would be good to match text with the resource/operator text size + styles. 
Image

the options seem to match but not the selected values

ui/src/components/entities-table/EntitiesToolbar.tsx
```diff
@@ -0,0 +1,232 @@
...
+  children,
+  className,
+}: {
+  label: string;
+  children: React.ReactNode;
+  className?: string;
+}) {
+  return (
+    <div className={cn('flex flex-col gap-1', className)}>
+      <span className="text-xs text-muted-foreground">{label}</span>
```
johallar, line 151: This should use a `label` element like the other fields

ui/src/components/entities-table/EntityDetailPanel.tsx
```diff
@@ -0,0 +1,315 @@
...
+import { useTheme, THEME_DARK } from '@/contexts/ThemeContext';
+
+interface EntityDetailPanelProps {
+  fsm: FiniteStateMachine | null;
+  resourceLabel: (id: string) => string;
+  operatorLabel: (id: string) => string;
+  stateColorFn?: (name: string) => string;
+}
+
+function isBytesStat(name: string): boolean {
```
johallar: Duplicated again here

cmatzenbach: Yup, just removed that one when addressing your comment from earlier

ui/src/components/entities-table/EntityDetailPanel.tsx
johallar, line 1: It would be good to adjust the precision here to be able to tell differences, it can show the same timestamp for all states
Image

johallar: `@5.626s`

johallar:
Throughout these components I think the general rule we want to stick to is that any data coming from the analyzer we should display with the `<DataText>` component

ui/src/components/EntityDetailDrawer.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+interface EntityDetailDrawerProps {
+  fsm: FiniteStateMachine | null;
+  resourceLabel: (id: string) => string;
+  operatorLabel: (id: string) => string;
+  onClose: () => void;
+  stateColorFn?: (name: string) => string;
+}
+
+export function EntityDetailDrawer({
```
johallar, line 26: We should use the shadcn ui drawer fr this if possible, which will handle a lot of the accessibility and mobile events for us: 
https://ui.shadcn.com/docs/components/base/drawer#non-modal

ui/src/components/QueryResourceTree.tsx
```diff
@@ -133,6 +136,29 @@ function QueryResourceTreeContent({ queryBundle, engineId }: QueryResourceTreePr
   const [selectedTypes, setSelectedTypes] = useAtom(selectedTypesAtom);
   const [selectedFsmTypes, setSelectedFsmTypes] = useAtom(selectedFsmTypesAtom);
 
+  const [drawerFsm, setDrawerFsm] = useState<FiniteStateMachine | null>(null);
+  const closeDrawer = useCallback(() => setDrawerFsm(null), []);
+
+  const stateColorFn = useMemo(
+    () => createFsmTypeColorFn(entities.fsm_types, isDark ? 'dark' : 'light'),
```
johallar: Should use `THEME_LIGHT` and `THEME_DARK`

ui/src/components/LongEntitiesRow.tsx
```diff
@@ -77,6 +84,15 @@ export function LongEntitiesRow({
   );
   const totalEntities = data?.pages[data.pages.length - 1]?.total ?? entities.length;
 
+  const handleEntityClick = useCallback(
```
johallar, line 117: If clicking an already highlighted entity, it should deselect IMO

ui/src/components/entities-table/EntityDetailPanel.tsx
```diff
@@ -0,0 +1,307 @@
...
+                    style={{ width: `${pct}%`, backgroundColor: stateColor }}
+                  />
+                </div>
+              )}
+
+              {transition.usages.length > 0 && (
+                <ul className="mt-1 space-y-0.5 text-xs text-muted-foreground">
+                  {transition.usages.map((usage, usageIndex) => (
+                    <li key={usageIndex} className="flex flex-wrap items-baseline gap-x-2">
+                      <span className="font-mono">{resourceLabel(usage.resource)}</span>
```
johallar: Could use `<DataText>`?

ui/src/components/entities-table/EntityDetailPanel.tsx
```diff
@@ -0,0 +1,307 @@
...
+                      {usage.capacities.map(([name, capacity], capacityIndex) => (
+                        <span key={capacityIndex} className="tabular-nums">
+                          {name}
+                          {capacity != null
```
johallar: This should use the `formatQuantity` function looking up the `capacity_specs` from the query bundle, with maybe a fallback to `inferFieldFormatter`

ui/src/components/entities-table/EntityDetailPanel.tsx
```diff
@@ -0,0 +1,307 @@
...
+                    <span
+                      className={`tabular-nums text-sm font-medium ${
+                        isBottleneck ? 'text-orange-500 dark:text-orange-400' : ''
+                      }`}
+                    >
+                      {formatDuration(durationMs)}
+                    </span>
+                  )}
+                  <span className="tabular-nums text-xs text-muted-foreground">
+                    @{transition.timestamp.toFixed(3)}s
```
johallar: Needs a little bit more doing, for many shorter tasks (when zoomed in) this shows the same timestamp for each

johallar:
the fsm capacity chart tooltip overflow is hidden/cutoff, you may need to go the react component route + a portal:
Screenshot 2026-08-11 at 10 34 01 AM

ui/src/components/entities-table/EntitiesTable.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+    },
+    {
+      name: 'finished',
+      timestamp: 1,
+      usages: [],
+      attributes: [],
+      derived_attributes: [],
+    },
+  ],
+};
```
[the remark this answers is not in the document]
cmatzenbach: Skipping this one. QueryBundle<EntityRef> requires ~10 additional deeply-nested fields beyond what the test uses (engine, query_group, query, workers, plans, ports, resource_types, resource_group_types, resource_groups, plan_tree, resource_tree, unique_operator_names, quantity_specs, start_time_unix_ns). Filling those in with stubs would be more noise than the as unknown as cast it replaces, and would break silently the moment any of those types evolve.

The correct long-term fix is to narrow EntitiesTable's prop type to Pick<QueryBundle<EntityRef>, 'query_id' | 'duration_s' | 'entities'> so the fixture can satisfy the prop directly without a cast — but that's a component API change that should be its own PR, not a test-only tweak.

ui/src/components/entities-table/EntityDetailPanel.tsx
```diff
@@ -0,0 +1,315 @@
...
+                <AttributeRows
+                  attributes={transition.derived_attributes}
+                  derived
+                  operatorLabel={operatorLabel}
+                />
+              )}
+            </li>
+          );
+        })}
+      </ol>
```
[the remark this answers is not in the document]
cmatzenbach: Skipping - the transition list in the detail panel is for a single selected entity - FSM transition counts are bounded by the state machine definition, typically 5–20 entries. That's well below the threshold where virtualizer overhead pays off; `@tanstack/react-virtual` introduces measurement and absolute-positioning cost that would make this slower for lists this small. Virtualization makes sense if transitions could realistically reach the hundreds, which they can't by nature of the FSM model.

ui/src/components/entities-table/EntityDetailPanel.tsx
johallar, line 1: I think this component or a reusable form of it should live in the @quent/components

ui/src/components/entities-table/EntityResults.tsx
```diff
@@ -0,0 +1,362 @@
...
+          <div className="p-4 text-sm text-muted-foreground">No entities match the filters.</div>
+        )}
+        {isLoading && <div className="p-4 text-sm text-muted-foreground">Loading…</div>}
+      </div>
+
+      <div className="shrink-0 border-t bg-card p-2 flex items-center justify-between text-xs text-muted-foreground">
+        <span>
+          {visibleStart}–{visibleEnd} of {total} {total === 1 ? 'entity' : 'entities'}
+        </span>
+        <div className="flex items-center gap-2">
```
johallar: can also use a shadcn component for this: 
https://ui.shadcn.com/docs/components/base/pagination

ui/src/components/entities-table/EntitiesToolbar.tsx
cmatzenbach: Agree, fixed

ui/src/components/LongEntitiesRow.tsx
cmatzenbach: Agreed, fixed

ui/src/components/entities-table/EntityResults.tsx
cmatzenbach: Good idea, replaced with shadcn pagination components

ui/src/components/entities-table/useEntityTable.ts
```diff
@@ -0,0 +1,191 @@
...
+      windowEnd: debouncedWindowEnd,
+    }),
+    [filters, debouncedMinUsageS, debouncedWindowStart, debouncedWindowEnd]
+  );
+  const request = useMemo(
+    () => buildEntityRequest({ filters: effectiveFilters, operatorId, page, queryId, durationS }),
+    [durationS, effectiveFilters, operatorId, page, queryId]
+  );
+  const query = useEntities({ engineId, request }, { enabled: validationErrors.length === 0 });
+  const isDebouncing =
```
johallar: nit: this ux is a little funky because it displays the loader immediately on change, so if the user is typing then they will see the loader as soon as they start, instead of it loading once when they pause.

ui/src/components/entities-table/useEntityTable.ts
```diff
@@ -0,0 +1,191 @@
...
+  const rows = useMemo(() => entityRows(query.data), [query.data]);
+  const pageSize = normalizePageSize(filters.pageSize);
+  const total = query.data?.total ?? 0;
+  const pageCount = Math.max(1, Math.ceil(total / pageSize));
+  const visibleStart = total === 0 ? 0 : page * pageSize + 1;
+  const visibleEnd = total === 0 ? 0 : Math.min(total, visibleStart + rows.length - 1);
+  const activeFilterCount = activeEntityFilterCount(filters, defaults, operatorId);
+
+  return {
+    activeFilterCount,
```
johallar: nit: I get why this is all one hook, but it's a bulky interface. Consider nesting one level (pagination stuff, filters stuff, query stuff, etc) to keep it a little more readable.

ui/src/components/entities-table/EntitiesToolbar.tsx
```diff
@@ -0,0 +1,203 @@
...
+              className="flex items-center gap-1 text-xs text-muted-foreground"
+            >
+              <LoaderCircle className="size-3.5 animate-spin" />
+              Updating…
+            </span>
+          )}
+        </div>
+      </div>
+
+      {validationErrors.length > 0 && (
```
johallar, line 131: Consider using shadcn mechanisms for this, they come with validation stuff: https://ui.shadcn.com/docs/components/base/input#invalid

ui/src/components/entities-table/useEntityTable.ts
cmatzenbach: Good catch - fixed so that now loader only shows when data is actually loading

ui/src/components/entities-table/EntityDetailPanel.tsx
```diff
@@ -90,6 +92,21 @@ export function EntityDetailPanel({
     };
   }
 
+  // Find data volume from derived attributes (last bytes-stat with a numeric value)
```
johallar: This panel already [shows all derived attributes](https://github.com/rapidsai/quent/pull/589/changes#diff-8efd8132d16172d62c75f916dd61ec4506e51fc90033001f7d6eec163cf68fbcR24), I think we can remove this specific calculation + section

ui/src/components/entities-table/utils.ts
```diff
@@ -0,0 +1,134 @@
...
+export const MAX_PAGE_SIZE = 500;
+
+export function defaultEntityFilters(durationS: number): EntityFilters {
+  return {
+    entityType: null,
+    resourceId: null,
+    minUsageS: '',
+    windowStart: '0',
+    windowEnd: String(durationS),
+    sortDir: 'Desc',
```
johallar: nit: re-usable constants better than using these string literals throughout, packaged with the `SortableHeader` ideally

ui/src/components/entities-table/useEntityTable.ts
cmatzenbach: Refactored in [73c70ca](https://github.com/rapidsai/quent/pull/564/commits/73c70ca1ae090a4f071051db3c8fa74a59854f80)

ui/src/components/entities-table/utils.ts
```diff
@@ -0,0 +1,164 @@
...
+): boolean {
+  return (
+    activeFilterCount > 0 ||
+    filters.sortDir !== defaults.sortDir ||
+    filters.pageSize !== defaults.pageSize
+  );
+}
+
+function parseOptionalNumber(value: string): number | null {
+  if (value.trim() === '') return null;
```
johallar: You'll need to merge in main and run lint:fix, since main will enforce the brackets rule now

cmatzenbach (commented):
Commenting for now, as I haven't gotten a chance to do a thorough UI test

ui/packages/@quent/components/src/lib/sliderField.utils.ts
```diff
@@ -0,0 +1,27 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export function clamp(value: number, min: number, max: number): number {
```
johallar, line 4: Should probably live in top level utils packages/@quent/utils/something

johallar (approved):
Changes look great! One organizational comment and one thought on the duration slider:

The duration slider should probably not have max of the whole range, but should be the longest entity (maybe query for a single entity sorted by longest duration to get the max of this slider?)

johallar:
Screenshot 2026-08-26 at 9 22 18 PM

Bubbling this up actually, i'm not sure if we want to have the operator dropdown here since the DAG is on the left and probably a better way to select this crossfilter.

johanpel (approved):
Rust changes LGTM. Very nice feature, thanks!

ui/packages/@quent/components/src/lib/sliderField.utils.ts
cmatzenbach: I moved it to the `ui/packages/@quent/components/src/lib/` folder, where other helper files containing pure functions consumed by sibling components live (and renamed according to the convention of other files in there).

cmatzenbach:
> Changes look great! One organizational comment and one thought on the duration slider:
> 
> The duration slider should probably not have max of the whole range, but should be the longest entity (maybe query for a single entity sorted by longest duration to get the max of this slider?)

Agree - fixed. I ended up using the top entity's end timestamp, since that gives us an absolute time range.

EDIT: On second thought, we may want to revisit this in the future. The problem is, there are a decent amount of entities with very short ranges above that max. So setting the slider's max at the max of the longest entity automatically excludes these values. They can be restored by clicking the "Reset" button, but that behavior may not be desired - a point for discussion later on. For now how it works is, on page load all the entities are shown. If you drag the top end of the slider, it will treat the proposed max as the range of the slider. Thus, if you drag it back to the full range, you will be looking at a filtered result set - it does show one active filter, so hitting Reset will clear this filter and get you back to showing all the entities, but the user has to realize that.

cmatzenbach:
> Screenshot 2026-08-26 at 9 22 18 PM
> Bubbling this up actually, i'm not sure if we want to have the operator dropdown here since the DAG is on the left and probably a better way to select this crossfilter.

I dunno, I could see some use for it. Using the search option within the combobox to find specific operators is a lot quicker than having to navigate through the DAG tree, especially when you have multiple physical plans. Also, users _could_ collapse the DAG tree if they want to focus on the other items, in which case the Operator field would be very useful. I say we leave it for now and if it doesn't get much use we can remove it later - for example, if we find people are used to navigating solely from the DAG.

johallar:
> I dunno, I could see some use for it. Using the search option within the combobox to find specific operators is a lot quicker than having to navigate through the DAG tree, especially when you have multiple physical plans. Also, users _could_ collapse the DAG tree if they want to focus on the other items, in which case the Operator field would be very useful. I say we leave it for now and if it doesn't get much use we can remove it later - for example, if we find people are used to navigating solely from the DAG.

Fair enough re collapsing the dag panel. It may be nice to add more detail here (like we I'd like to do for the resources dropdown) so you can see associated parent plans or workers, etc

PR #569, opened 2026-08-11
feat(ui):  long entities swimlanes 3.5

### LINKED ISSUE #98: Handle many overlapping tasks in timeline tooltip
Image

* This both locks up the browser AND looks bad, fix both. 
* Aggregate on state? Truncate the list somehow? Make tooltip pinnable + scroll?
* Also have a look at stats/custom stats that can be a part of these fsm entities, and how to best display in the tooltip

johallar:
# Description
This iteration is to improve usability when handling large numbers of entities. 
* Adds expand/collapse controls for long entities swimlanes
* Adds loader and persistent page size for "show more" functionality. Reduces jumpyness by not hiding/showing the button while loading, and keeps page size when Zooming/Panning so swimlanes jump less.
* Update tooltip to truncate after showing a certain number of entities when many are overlapping. Not being able to show all information should be mitigated by the detail panel https://github.com/rapidsai/quent/pull/564. Future iterations may also add aggregate stats about the entities if too many to show all individually.
* More informative "No Entities" message, shows the threshold being used
* Adds rough setting for long entities threshold (show less <--> show more)

## Related Issues
Fixes: #98 
Relates to: #215 

## Testing
* Simulate a query with many entities and many threads so there are many concurrent entities in the chart. 
`cargo run -p quent-simulator -- --num-query-groups 1 --num-queries 1 --num-workers 2 --num-threads 24 --num-tasks 65536 --exporter collector --collector-address http://localhost:7836`
* Open query and open any resource with entities that have usage on that resource
* Change the entities slider to show more/less entities (changes the long entity threshold in the request)
* Entity rows are expandable/collapsible
* Resource tree virtualization works as expected (opening every resource will not overwhelp the browser), charts render as they scroll into view.

## Screenshots
Demo with an absurd (i think?) amount of entities. Every interaction should be able to handle huge amounts of entities. It can get a little jumpy when there are so many, but the price we pay RN. 

https://github.com/user-attachments/assets/b245bc12-6774-4467-8f30-afe063321313

### DIFF
--- a/ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
+++ b/ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
@@ -199,43 +199,62 @@ function MarkDetailRow({ name, value }: { name: string; value: string }) {
   );
 }
 
+const DETAILED_ACTIVE_MARK_LIMIT = 6;
+const COMPACT_ACTIVE_MARK_LIMIT = 8;
+
 function ActiveMarksSection({ marks }: { marks: ActiveMark[] }) {
   if (marks.length === 0) return null;
+  const isCompact = marks.length > DETAILED_ACTIVE_MARK_LIMIT;
+  const visibleMarks = isCompact ? marks.slice(0, COMPACT_ACTIVE_MARK_LIMIT) : marks;
+  const hiddenCount = marks.length - visibleMarks.length;
+
   return (
     <div className="mt-1 pt-1 border-t border-border">
-      {marks.map((m, i) => (
+      {visibleMarks.map((m, i) => (
         <div key={i}>
           <div className="flex items-center gap-1">
             <ColorSwatch color={m.color} />
             <DataText className="text-muted-foreground">{m.label}</DataText>
             <DataText className="text-foreground font-medium ml-auto">{m.stateName}</DataText>
           </div>
-          {m.durationMs !== undefined && (
-            <MarkDetailRow name="duration" value={formatDuration(m.durationMs)} />
-          )}
-          {m.attributes?.map(attr => (
-            <MarkDetailRow
-              key={attr.key}
-              name={attr.key}
-              value={formatAttributeValue(attr.key, attr.value)}
-            />
-          ))}
-          {m.derivedAttributes && m.derivedAttributes.length > 0 && (
+          {!isCompact && (
             <>
-              <DataText as="div" className="pl-3 pt-0.5 text-muted-foreground italic opacity-70">
-                derived
-              </DataText>
-              {m.derivedAttributes.map(attr => (
+              {m.durationMs !== undefined && (
+                <MarkDetailRow name="duration" value={formatDuration(m.durationMs)} />
+              )}
+              {m.attributes?.map(attr => (
                 <MarkDetailRow
                   key={attr.key}
                   name={attr.key}
                   value={formatAttributeValue(attr.key, attr.value)}
                 />
               ))}
+              {m.derivedAttributes && m.derivedAttributes.length > 0 && (
+                <>
+                  <DataText
+                    as="div"
+                    className="pl-3 pt-0.5 text-muted-foreground italic opacity-70"
+                  >
+                    derived
+                  </DataText>
+                  {m.derivedAttributes.map(attr => (
+                    <MarkDetailRow
+                      key={attr.key}
+                      name={attr.key}
+                      value={formatAttributeValue(attr.key, attr.value)}
+                    />
+                  ))}
+                </>
+              )}
             </>
           )}
         </div>
       ))}
+      {hiddenCount > 0 && (
+        <DataText as="div" className="pt-1 text-muted-foreground">
+          {hiddenCount} more {hiddenCount === 1 ? 'entity' : 'entities'} not shown
+        </DataText>
+      )}
     </div>
   );
 }
@@ -347,7 +366,7 @@ export function EntityTooltipContent({
   activeMarks: ActiveMark[];
 }) {
   return (
-    <div className="px-2 py-1.5 bg-popover rounded text-[11px] text-foreground leading-tight shadow-md z-50">
+    <div className="max-h-[50vh] overflow-hidden px-2 py-1.5 bg-popover rounded text-[11px] text-foreground leading-tight shadow-md z-50">
       <DataText as="div" className="font-semibold mb-1 text-muted-foreground">
         {formatDurationForWindow(timestamp, windowMs)}
       </DataText>
--- a/ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
+++ b/ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
@@ -357,7 +357,7 @@ export function EntityTooltipContent({
   activeMarks: ActiveMark[];
 }) {
   return (
-    <div className="max-h-[50vh] overflow-hidden px-2 py-1.5 bg-popover rounded text-[11px] text-foreground leading-tight shadow-md z-50">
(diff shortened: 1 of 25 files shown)


ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
```diff
@@ -347,7 +357,7 @@ export function EntityTooltipContent({
   activeMarks: ActiveMark[];
 }) {
   return (
-    <div className="px-2 py-1.5 bg-popover rounded text-[11px] text-foreground leading-tight shadow-md z-50">
+    <div className="max-h-[50vh] overflow-hidden px-2 py-1.5 bg-popover rounded text-[11px] text-foreground leading-tight shadow-md z-50">
```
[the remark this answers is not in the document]
johallar: d25d6c2

ui/packages/@quent/hooks/src/timeline/useTimelineAtoms.ts
```diff
@@ -11,19 +11,43 @@ import {
...
 }
 
+function useReturnedTimelineData(resourceId: string): SingleTimelineResponse | undefined {
+  const timelineDataMap = useAtomValue(timelineDataMapAtom);
+  const visibleEntries = useAtomValue(visibleEntriesAtom);
+  const request = visibleEntries[resourceId];
+  if (!request) return undefined;
+  const key = timelineCacheKey({
+    resourceId,
+    resourceTypeName: getResourceTypeName(request),
```
9prady9, line 41: Fallback timeline responses never reach this atom. `applyBulkTimelineResponse` skips non-OK entries, while `ResourceTimeline` may recover with its individual `fetchedData`; because that result is not written to `timelineDataMapAtom`, this hook stays `undefined` and `LongEntitiesRow` renders its loading skeleton indefinitely. Please either seed the atom from fallback results or derive the bin count from the effective timeline query.

ui/packages/@quent/hooks/src/timeline/useTimelineAtoms.ts
```diff
@@ -11,19 +11,43 @@ import {
...
+    fsmTypeName: getFsmTypeName(request),
+  });
+  return timelineDataMap[key];
+}
+
+export function useReturnedTimelineNumBins(resourceId: string): number | undefined {
```
9prady9, line 53: This cached response is keyed only by resource/type/FSM, not viewport. After a zoom or pan, the visible request updates while the map still contains the prior response, so the new window is divided by the old `num_bins` and an incorrectly thresholded entity request is enabled until the bulk fetch completes. Please tie the response to its request configuration/span, or return `undefined` until they match.

ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
```diff
@@ -347,7 +357,7 @@ export function EntityTooltipContent({
   activeMarks: ActiveMark[];
 }) {
   return (
-    <div className="px-2 py-1.5 bg-popover rounded text-[11px] text-foreground leading-tight shadow-md z-50">
+    <div className="max-h-[50vh] overflow-y-auto overflow-x-hidden px-2 py-1.5 bg-popover rounded text-[11px] text-foreground leading-tight shadow-md z-50">
```
9prady9: `overflow-y-auto` is not actually interactive here: `PointerTooltipPortal` wraps this content in `pointer-events-none`. Wheel, mouse, and touch events therefore target the chart underneath, so attributes beyond `50vh` remain inaccessible. Please make the tooltip interactive/pinnable or truncate the overflowing details.

9prady9 (commented):
Three inline issues: fallback loading, stale viewport bin counts, and inaccessible tooltip overflow.

ui/packages/@quent/components/src/timeline/TimelineTooltip.tsx
johallar: True, this should be mitigated further when we have a detail panel: https://github.com/rapidsai/quent/pull/564, for now i'll stick with limiting the number of things that can show in the tooltip, but remove the max-h and overflow styles so it will grow w the (limited) content.

After we have a detail panel, we can truncate or summarize tooltip items further.

ui/packages/@quent/hooks/src/timeline/useTimelineAtoms.ts
johallar: The number of bins does not change based on the viewport, but I could see it being controlled by a UI setting at some point in the future. I am going to setup a fallback value that uses the same const that is used to populate num_bins. If it moves to a setting in the future, it should update this as well.

ui/packages/@quent/hooks/src/timeline/useTimelineAtoms.ts
johallar: This should also be resolved by using the MAX_TIMELINE_BINS const as a fallback value.

ui/packages/@quent/hooks/src/timeline/useTimelineAtoms.ts
```diff
@@ -11,19 +11,43 @@ import {
   startTimeMsAtom,
   bulkInitializedAtom,
   visibleEntriesAtom,
+  longEntityDensityAtom,
+  timelineCacheKey,
 } from '../atoms/timeline';
-import type { ZoomRange, SingleTimelineResponse } from '@quent/utils';
+import { type ZoomRange, type SingleTimelineResponse } from '@quent/utils';
+import { getFsmTypeName, getResourceTypeName } from './timeline.utils';
```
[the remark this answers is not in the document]
johallar: fd877cef

ui/packages/@quent/utils/src/timeline.ts
```diff
@@ -0,0 +1,15 @@
...
+export function getResourceTypeName(request: TimelineRequest<OperatorFilter> | undefined): string {
+  if (!request || !('ResourceGroup' in request)) return '';
+  return request.ResourceGroup.resource_type_name;
+}
+
+export function getFsmTypeName(request: TimelineRequest<OperatorFilter>): string | null {
+  return 'ResourceGroup' in request
+    ? request.ResourceGroup.entity_filter.entity_type_name
+    : request.Resource.entity_filter.entity_type_name;
+}
```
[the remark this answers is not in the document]
johallar: 721edc0

ui/packages/@quent/hooks/src/timeline/useTimelineAtoms.ts
9prady9: Re-checking this against the timeline cache, the actual response bin count does vary with the viewport. The server combines the overlapping chunk bins and explicitly returns that actual count; for example, a full viewport can yield 400 bins while a 75% viewport yields about 300. Because `timelineDataMapAtom` is keyed without the viewport, this hook returns the previous count during the new fetch, so `LongEntitiesRow` combines the new window with the old divisor and issues an incorrectly thresholded request. Can we associate the cached response with its request span, or return `undefined` until the response matches the active viewport?

9prady9 (approved):

ui/packages/@quent/hooks/src/timeline/useTimelineAtoms.ts
johallar: oo yeah ok, i was wrong, ok lemme address this

johallar: 635c585e6356246ce2afc902051342a6d740a33f

Adds check for stale state, only falls back to default if request completes, it's NOT stale, and there's still no number of bins.

PR #585, opened 2026-08-17
feat(simulator): add nvtx ranges to simulated queries

johallar:
# Description
⚠️ Vibe Coded

Implements nvtx range generation in the simulator


johallar:
Add:
* categories
* generate all nvtx entity types (push/pop, marks, etc)

johanpel:
I think we'll need to hold off a bit for https://github.com/rapidsai/quent/issues/618 so everything can follow the same export / import path and a lot of code here will no longer be necessary.

PR #588, opened 2026-08-17
feat(ui): deep linking part 2

johallar:
# Description
* Adds the rest of the "useful" state to deep linking context, this should allow users to share most of the existing UI state in the URL.
* Figures out the versioned schema plan in more depth. This should allow new version of quent to be fully backwards compatible with old shared links. Schemas are defined with a version, this version determines how the payload will be decoded and used to populate the app state. 
* Removes the v dated /node/:id route

## Related Issues
#

## Testing
1. Test sharing/generation of all state here: https://github.com/rapidsai/quent/pull/588/changes#diff-50bba7a20bf246d6a0819f863cc4355c298524fff331876755610a43dbf75ccbR100-R139
2. Pasted URLs should hydrate all state as it was
3. Copy a share URL from the `main` branch (expand resources, and zoom to specific area), checkout this feature branch, paste url in
4. Pasted URL should still populate the zoom range start/end and expanded resource IDs

## Screenshots

https://github.com/user-attachments/assets/82eee292-60c6-4174-840a-0b1827b53005


ui/docs/deep-links.md
```diff
@@ -7,32 +7,61 @@ Quent deep links are snapshots created by the **Copy Link** action. Opening a
...
 
 ```text
-/profile/engine/ENGINE/query/QUERY/timeline?s=v1.COMPRESSED_STATE
+/profile/engine/ENGINE/query/QUERY/timeline?s=v2.COMPRESSED_STATE
 ```
 
 Incoming state is treated as untrusted data and validated with the same Zod
-schema used by the UI and command-line tool. The complete absolute URL is
-limited to 2,048 characters.
+schema used by the UI and command-line tool. Arrays and identifier lengths have
```
cmatzenbach: Let use "resource id" instead of "identifier". It's not clear what that refers to.

ui/packages/@quent/components/src/dag/DAGChart.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    metadata: { relatedOperatorIds: ['physical-1', 'physical-2'] },
+  },
+  {
+    id: 'other',
+    label: 'Other operator',
+    type: 'scan',
+  },
+];
+
+describe('resolveInspectedNodeData', () => {
```
cmatzenbach, line 22: As I mentioned on a previous PR, if we're going to start a test I think we should either add the full suite for the component or at least file a ticket so we know to go back and fill out the rest. It'll be easier for auditing later on when we try to figure out what still needs tested (granted, programs will tell us where coverage is missing...). Let me know if you disagree.

ui/packages/@quent/hooks/src/index.ts
```diff
@@ -138,6 +138,16 @@ export type {
 // Utility hooks
 export { useDeferredReady } from './dag/useDeferredReady';
 
+// Serializable state bridge for application deep links
```
cmatzenbach: Unnecessary comment, the import and file names are clear enough

ui/src/components/operator-table/OperatorTable.tsx
```diff
@@ -25,11 +25,15 @@ import {
 } from '@quent/hooks';
 import type { QueryBundle, EntityRef } from '@quent/utils';
 import { useTheme, THEME_DARK } from '@/contexts/ThemeContext';
-import type { OperatorTableRow } from './types';
+import {
+  DEFAULT_OPERATOR_TABLE_ENABLED,
+  OPERATOR_TABLE_INDEX_ORDER,
```
cmatzenbach, line 30: Nice refactor

ui/packages/@quent/hooks/src/deepLink/useSerializableViewState.ts
```diff
@@ -0,0 +1,218 @@
...
+  dataFlow?: Partial<SerializableDataFlowState>;
+  operatorTable?: SerializableOperatorTableState;
+}
+
+interface SerializableViewStateOptions {
+  operatorTablePersistKey: string;
+  operatorTableGroupKeys: readonly string[];
+}
+
+export function useSerializableViewState({
```
cmatzenbach, line 88: I played around with some different ways to possibly auto-generate this instead of having to hand-code this and `hydrate`, but I see the issue, some of the values aren't simple scalars (e.g. `dataFlow.playheadS`). I did find a way to generate most of this, but the special fields have to be taken care of separately, so in the end, it's not worth it. We will have to remember to update `read`, `hydrate`, the interfaces, and the default omission logic in `DeepLinkBoundary.tsx`'s `copyLink` function if we ever add more fields to these areas of state that are needed for deep linking.

cmatzenbach:
Found one bug:

1. Go to timeline page, scrub the timeline
2. Click Copy Link while still on timeline
3. Open that copied timeline link
4. Click the Operators tab
5. Now refresh the page, or copy the address bar URL into a private window

The issue is that in `src/routes/profile.engine.$engineId.query.$queryId.tsx` the current query param is passed into tab links; however, `DeepLinkBoundary.tsx` rejects payloads whose saved tab doesn't match the active route. I wonder if we should be rebuilding `s` for the destination tab (maybe this is too much work for a fairly obscure bug, feel free to punt off for later).

cmatzenbach:
AI had one finding that it found while I was reading the code that I think may be worthwhile:

In `src/routes/profile.engine.$engineId.tsx:64`, `isQueryReady` is `false` until `queryBundle != null`. Then `src/features/deep-link/DeepLinkBoundary.tsx:91` starts `isHydrated` as `false` when `encodedState` exists, line 103 keeps intake unresolved while `!isQueryReady`, and line 393 renders `null` instead of `children`.

The issue is that `children` includes the layout and `Outlet`, so the query route’s own pending/error UI from `src/routes/profile.engine.$engineId.query.$queryId.tsx:15` may not render during encoded-link loads. On a slow query fetch, the user can see a blank app instead of `QueryLoading`. On a loader failure, the boundary may stay unhydrated because `queryBundle` never becomes non-null, so the route error UI can also be hidden.

Maybe either render the pending/error route UI while intake is unresolved, or make the boundary distinguish “query is pending” from “query failed” and only defer the stateful timeline/DAG children that need hydration.

ui/docs/deep-links.md
cmatzenbach: Old comment, you're going to re-word per our conversation. Disregard

johallar:
> The issue is that in `src/routes/profile.engine.$engineId.query.$queryId.tsx` the current query param is passed into tab links; however, `DeepLinkBoundary.tsx` rejects payloads whose saved tab doesn't match the active route. I wonder if we should be rebuilding `s` for the destination tab (maybe this is too much work for a fairly obscure bug, feel free to punt off for later).

Posting here as well for posterity: This _is_ a bug, I will either remove the deep link encoded state from the URL once it's loaded, or address this in a follow up that keeps the encoded state up to date as app state changes

ui/packages/@quent/hooks/src/deepLink/useSerializableViewState.ts
johallar: We could maybe formalize this a little bit with a wrapper function (`encodeState`), that would keep track of what to encode... i'll make a ticket for this as it's probably good to formalize this part a little bit.

johallar: Wrote up ideas here: https://github.com/rapidsai/quent/issues/633

ui/packages/@quent/components/src/dag/DAGChart.test.ts
johallar: I think adding tests relevant to the feature/bug is probably the best policy, to keep PRs focused. If we want to make a concerted push to increase coverage I think an "epic" (parent issue) with areas we want to test is probably in order

johallar:
> > The issue is that in `src/routes/profile.engine.$engineId.query.$queryId.tsx` the current query param is passed into tab links; however, `DeepLinkBoundary.tsx` rejects payloads whose saved tab doesn't match the active route. I wonder if we should be rebuilding `s` for the destination tab (maybe this is too much work for a fairly obscure bug, feel free to punt off for later).
> 
> Posting here as well for posterity: This _is_ a bug, I will either remove the deep link encoded state from the URL once it's loaded, or address this in a follow up that keeps the encoded state up to date as app state changes

@cmatzenbach I am just removing the state from the url once its loaded, since it will almost always be stale immediately. https://github.com/rapidsai/quent/issues/634 to take this to the next step

ui/src/features/deep-link/DeepLinkBoundary.tsx
```diff
@@ -125,30 +206,156 @@ export function DeepLinkBoundary({
...
+
+    const dataFlow: NonNullable<DeepLinkStateV2['dataFlow']> = {};
+    if (!sharedView.dataFlow.enabled) dataFlow.enabled = false;
+    if (sharedView.dataFlow.measure) dataFlow.measure = sharedView.dataFlow.measure;
+    if (sharedView.dataFlow.labelMeasure) {
+      dataFlow.labelMeasure = sharedView.dataFlow.labelMeasure;
+    }
+    if (sharedView.dataFlow.dimensions) {
+      dataFlow.dimensions = sharedView.dataFlow.dimensions;
+    }
```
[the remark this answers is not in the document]
johallar: https://github.com/rapidsai/quent/pull/588/commits/ef0a5a469074a9bbdd2fca44e8961ed0cd79e41e

johallar:
> AI had one finding that it found while I was reading the code that I think may be worthwhile:
> 
> In `src/routes/profile.engine.$engineId.tsx:64`, `isQueryReady` is `false` until `queryBundle != null`. Then `src/features/deep-link/DeepLinkBoundary.tsx:91` starts `isHydrated` as `false` when `encodedState` exists, line 103 keeps intake unresolved while `!isQueryReady`, and line 393 renders `null` instead of `children`.
> 
> The issue is that `children` includes the layout and `Outlet`, so the query route’s own pending/error UI from `src/routes/profile.engine.$engineId.query.$queryId.tsx:15` may not render during encoded-link loads. On a slow query fetch, the user can see a blank app instead of `QueryLoading`. On a loader failure, the boundary may stay unhydrated because `queryBundle` never becomes non-null, so the route error UI can also be hidden.
> 
> Maybe either render the pending/error route UI while intake is unresolved, or make the boundary distinguish “query is pending” from “query failed” and only defer the stateful timeline/DAG children that need hydration.

I've updated this fix so that an outlet/error will still render which should allow query loader + errors to be displayed and not leave the user hanging.

cmatzenbach (approved):

PR #591, opened 2026-08-21
feat(ui): resource tree filter

### LINKED ISSUE #169: ResourceTree search/filter
Had this thought the other day @johanpel, thoughts?

Not sure how useful this in practice, but seems like if you had many resources, some way to filter would be useful (only show me resource of type T)

johallar:
# Description
Implements resource tree filtering by fsm type, resource type, plain text, or id. 

Integrates with deep linking so users or agents can highlight or filter resources to share.

Depends on https://github.com/rapidsai/quent/pull/588

## Related Issues
Fixes #169 

## Testing
1. Open any query, open the search options. Try all options provided, verify that tree filters as expected. All sub rows should come along with search results (operators, entities, etc)
2. Search should apply to NVTX tree
3. Search box can also handle comma delimited IDs, fish some resource IDs out of the network request and try comma delimited list of IDs
4. Use search term that finds nothing, verify empty message
5. "Show all" checkbox should show results in context of the full tree
6. Clear should clear all active filters
7. Badge next to search icon should show number of active filters
8. Verify that deep linking is fully integrated, all resource filters + show all should be deep link-able

## Screenshots

https://github.com/user-attachments/assets/361e8cae-5caf-4dfd-a20a-7be21a8a63d3


cmatzenbach:
When trying a bogus search term, I noticed all the timeline stuff is filtered out, but the NVTX rows still appear. Are they not being filtered by search term?

Screenshot_20260901_161021

cmatzenbach:
UI nit (take it or leave it): I do wonder if we should make the menu "pop" a bit more. You can see it is an overlay on the right, as it differentiates against the backdrop of the timeline scrubber, but on the left side it blends into the background. It's not imperative as it doesn't take up too much vertical real estate now, but it may be nice visually to have it stand out more (and may be helpful should the menu expand to multiple rows).

image

cmatzenbach:
So when you have an active filter, it is not letting me collapse the worker rows. Was this intentional? I think users should still be able to collapse the rows if they want, it's not clear that they would need to clear filters to be able to do this again (if this is intentional, maybe a pop-up saying "Clear filters to collapse worker groups" or something)

https://github.com/user-attachments/assets/81ae9c2c-448c-47f0-b9ae-6cb4f8f4d964

cmatzenbach:
So on queries _without_ NVTX data, when I add a filter that has 0 results, I get a nice message like this:

Screenshot_20260901_163722


This doesn't happen when NVTX data is present - there must be some check that's failing due to NVTX data still being present that is preventing that from showing.

ui/scripts/deep-link.ts
```diff
@@ -52,7 +64,16 @@ async function readState(
             : {}),
         }
       : { ...raw, route };
-  const parsed = DeepLinkStateV2Schema.safeParse(candidate);
+  const resourceFilter = {
+    ...(typeof values['resource-search'] === 'string' ? { search: values['resource-search'] } : {}),
+    ...(parseList(values['resource-types'])
+      ? { resourceTypes: parseList(values['resource-types']) }
```
cmatzenbach: Rather than calling this twice (both here and below), we should hoist to a local const for clarity/efficiency, e.g.:

```
const resourceTypes = parseList(values['resource-types']);  
const fsmTypes = parseList(values['fsm-types']);
```

johallar:
> When trying a bogus search term, I noticed all the timeline stuff is filtered out, but the NVTX rows still appear. Are they not being filtered by search term?

Yeah this is a bit of an oversight, NVTX is not included in the search for now, and eventually the NVTX rows will become sub-rows of the resources. The short/medium term fix is to probably show the empty message for each sub-tree.

I am working on a follow up that allows each sub tree (resources/nvtx) to determine its own filtering capabilities.

https://github.com/user-attachments/assets/ebedbd22-6c43-46c6-a335-e5ad672cb0bc

johallar:
> UI nit (take it or leave it): I do wonder if we should make the menu "pop" a bit more. You can see it is an overlay on the right, as it differentiates against the backdrop of the timeline scrubber, but on the left side it blends into the background. It's not imperative as it doesn't take up too much vertical real estate now, but it may be nice visually to have it stand out more (and may be helpful should the menu expand to multiple rows).
> 
> image

I agree, i'll explore some options

johallar:
@cmatzenbach Addressed here:
* text search now searches both nvtx AND resources
* Resource rows expand/collapse in "show all" mode
* Added border + a little extra padding for search box to pop more. Think it could still use more, but best I could do for now.

cmatzenbach (approved):
Looks great!

PR #608, opened 2026-08-24
chore: group Dependabot dependency updates

9prady9:
# Description

Group Dependabot **minor and patch** version updates per ecosystem, so each of GitHub Actions, npm, and Cargo opens a single weekly pull request instead of one per dependency.

Major updates are deliberately left out of the groups and continue to open individually, so a breaking bump is always reviewed on its own and one bad update cannot block the rest of the ecosystem.

## Related Issues

N/A

## Testing

- Parsed `.github/dependabot.yml` and confirmed each of the three ecosystem entries defines a group scoped to `version-updates` with `update-types: [minor, patch]` and an explicit commit message prefix
- Ran `git diff --check`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,37 +1,60 @@
 version: 2
 
-multi-ecosystem-groups:
-  quent-dependencies:
-    schedule:
-      interval: "weekly"
-
 updates:
-  - cooldown:
+  - commit-message:
+      prefix: "chore(deps)"
+    cooldown:
       default-days: 14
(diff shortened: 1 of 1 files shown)


9prady9:
I am working on separating major version updates which is not a good idea to club with other dependencies.

johanpel (approved):
LGTM, tnx!

9prady9:
@mbrobbel Any feedback ?

mbrobbel (approved):

PR #622, opened 2026-08-25
refactor(ui): xaxis pointer using jotai

johallar:
# Description
xaxis pointer (vertical dotted line, indicating current mouse position) was getting pretty buggy when going through echarts. This busts out of echarts and uses an atom to publish/listen for the xaxis pointer location.

## Related Issues
N/A

## Testing
* Open dashboard with (ideally) resources, entities, operator gantt charts, nvtx ranges.
* Hovering over timeline controller (especially when zoomed in), displays xaxis pointer on all resource/nvtx charts, and sub row charts (operators swimlanes, entities)
* Hovering over any sub chart type (resources, nvtx, entities, operators) shows xaxis pointer on all other charts AND timeline controller (within datazoom), and timeline ruler.

## Screenshots

https://github.com/user-attachments/assets/f2caf34f-e43a-412f-a023-88e87358ca04

PR #623, opened 2026-08-25
refactor(ui): animate playhead line when playing

johallar:
# Description
Animates the playhead on the resource timeline view when data-flow is playing. Already looked smooth when dragging and don't want the delay in movement so only animate when playing.

## Related Issues
N/A

## Testing
* Click "play" on data-flow 
* Verify that data-flow playhead is moving v smoothly through the timelines
* Click and drag the data-flow slider handle
* Verify that there's no delay when dragging, or clicking to jump to a different spot on the slider

## Screenshots

https://github.com/user-attachments/assets/0a90e080-03fc-4846-b717-8ecbbb8cb1a8

### DIFF
--- a/ui/packages/@quent/components/src/timeline/PlayheadLine.test.tsx
+++ b/ui/packages/@quent/components/src/timeline/PlayheadLine.test.tsx
@@ -2,22 +2,44 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { render } from '@testing-library/react';
+import type { EChartsInstance } from 'echarts-for-react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
+import { usePlayheadLinePixel } from '../lib/usePlayheadLinePixel';
 import { PlayheadLine } from './PlayheadLine';
 
-const mocks = vi.hoisted(() => ({ isPlaying: false }));
+const mocks = vi.hoisted(() => ({
+  isPlaying: false,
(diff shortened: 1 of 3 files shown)


cmatzenbach:
Not sure if you want to address it here or in the next NVTX PR part, but there should be a hover line over the NVTX gantt charts.

ui/packages/@quent/components/src/timeline/PlayheadLine.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+vi.mock('@quent/hooks', () => ({
+  useDataFlowIsPlaying: () => mocks.isPlaying,
+}));
+
+vi.mock('../lib/usePlayheadLinePixel', () => ({
+  usePlayheadLinePixel: () => 24,
+}));
+
+describe('PlayheadLine', () => {
```
cmatzenbach, line 23: This test (obviously) only covers the playhead line transition behavior. Since we're making it, I think it'd be worth it to add all the tests for this component, just in case we don't get back to this for awhile.

cmatzenbach (approved):
Would like to see full component coverage in the new unit test created, but other than that looks great.

johallar:
> Not sure if you want to address it here or in the next NVTX PR part, but there should be a hover line over the NVTX gantt charts.

Fixed in https://github.com/rapidsai/quent/pull/622

PR #626, opened 2026-08-26
build: replace protoc with protox

mbrobbel:
Use [protox](https://docs.rs/protox/latest/protox/) instead of `protoc`.


johanpel (approved):
Nice, thanks

PR #630, opened 2026-08-26
build(nvtx): move the example to nvtx 2.0, NVIDIA's official bindings

9prady9:
## What

The `nvtx` crate changed hands. Versions 1.1–1.3 were a community wrapper by Spencer Imbleau (`simbleau/nvtx`); **2.0.0, published 2026-08-18, is NVIDIA's own binding**, released from `github.com/NVIDIA/NVTX` under `Apache-2.0 WITH LLVM-exception`. This moves the example's producer onto it.

Only the producer side changes — `nvtx-events`, `nvtx-injection`, and the analyzer's reconstruction are untouched.

## API translation

2.0 replaces the macro API with a function API:

| Before (1.3) | After (2.0) |
|---|---|
| `nvtx::name_thread!("…")` | `nvtx::name_thread(current_thread_id(), c"…")` |
| `nvtx::mark!("…")` | `nvtx::mark(c"…")` |
| `nvtx::range_push!` / `range_pop!` | `nvtx::LocalRange::new(…)` + drop |
| `nvtx::range!("…")` | `nvtx::Range::new(c"…")` |

## Notes

- **`default-features = false, features = ["std"]`.** 2.0 is `no_std`-capable; the default set also enables `cuda`, `cuda_runtime`, `tracing`, and `color-name`. `std` is the only one the example needs, and it keeps the CUDA toolkit out of the build.
- **The thread id comes from `SYS_gettid` directly** rather than the crate's `name-current-thread` feature. That feature pulls the `gettid` crate, which calls the glibc `gettid()` wrapper — exported only by glibc >= 2.30, while the pixi env is `sysroot_linux-64-2.28`, so it would not link. This mirrors the existing rationale in `integrations/nvtx/injection/src/init.rs`, and yields the same tid the injection layer stamps on events.
- **No new platform constraint.** `nvtx-sys` 2.0 runs bindgen, but `nvtx-injection` — already an example dependency — required bindgen and `libclang` before this change.
- **Dependency footprint shrinks.** 2.0 drops `derive_builder`, which takes `darling`, `darling_core`, `darling_macro`, and `ident_case` with it; only `spin` is added.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,6 +133,9 @@ cc = "1.2"
 http = "1"
 indexmap = "2"
 log = {version = "0.4.28", features = ["kv"]}
+# NVIDIA's own NVTX bindings, published from github.com/NVIDIA/NVTX. Only `std`
+# is wanted; the defaults also pull in CUDA, `tracing`, and colour-name support.
+nvtx = { version = "2", default-features = false, features = ["std"] }
 petgraph = "0.8.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
 prettyplease = "0.3.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,8 +133,6 @@ cc = "1.2"
 http = "1"
 indexmap = "2"
 log = {version = "0.4.28", features = ["kv"]}
-# NVIDIA's own NVTX bindings, published from github.com/NVIDIA/NVTX. Only `std`
-# is wanted; the defaults also pull in CUDA, `tracing`, and colour-name support.
 nvtx = { version = "2", default-features = false, features = ["std"] }
 petgraph = "0.8.3"
 postcard = { version = "1", default-features = false, features = ["alloc"] }
--- a/integrations/nvtx/example/Cargo.toml
+++ b/integrations/nvtx/example/Cargo.toml
@@ -14,15 +14,16 @@ path = "src/main.rs"
 [dependencies]
 nvtx-injection = { path = "../injection", features = ["static-injection"] }
 nvtx-bridge = { path = "../bridge" }
+libc = "0.2"
 # `io-callback` selects the callback exporter used to print (bin) and collect
 # (test) captured events.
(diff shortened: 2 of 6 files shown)


Cargo.toml
```diff
@@ -133,6 +133,9 @@ cc = "1.2"
 http = "1"
 indexmap = "2"
 log = {version = "0.4.28", features = ["kv"]}
+# NVIDIA's Rust bindings are not published on crates.io. Pin the NVTX v3.5.0
+# release commit so the example uses the official producer reproducibly.
```
mbrobbel: It is since last week https://github.com/NVIDIA/NVTX/releases/tag/v3.6.0
- https://crates.io/crates/nvtx

9prady9: refresh again :) I pushed something in the meanwhile

Cargo.toml
```diff
@@ -133,6 +133,9 @@ cc = "1.2"
 http = "1"
 indexmap = "2"
 log = {version = "0.4.28", features = ["kv"]}
+# NVIDIA's own NVTX bindings, published from github.com/NVIDIA/NVTX. Only `std`
+# is wanted; the defaults also pull in CUDA, `tracing`, and colour-name support.
```
johanpel:
```suggestion
```

integrations/nvtx/example/Cargo.toml
```diff
@@ -14,15 +14,16 @@ path = "src/main.rs"
...
 # (test) captured events.
 quent-instrumentation = { path = "../../../crates/instrumentation", features = [
     "io-callback",
 ] }
 uuid = { workspace = true }
-# The NVTX producer: a header-only Rust binding that makes the NVTX calls the
-# injection layer captures. Self-contained (vendors the NVTX v3 headers).
-nvtx = "1.3"
+# The official NVIDIA NVTX Rust producer makes the calls the injection layer
+# captures.
```
johanpel:
```suggestion
```

johanpel (approved):
LGTM

PR #637, opened 2026-08-26
fix(ui): truncate and add hovercard to long nav bar selects

### LINKED ISSUE #631: bug: long query names should truncate
### Summary

Image

Long query name stretches and overflows container, it should truncate and have title or other hover to see full query text. Obviously should also extend to query group and engines.

### Version / commit

main

1. Open <ask dhruv for benchmark db run id>
2. See query name overflows and stretches dropdown container

johallar:
# Description
Fixes bug with engine/query group/query navigation overflow

* Adds overflow detection, and shows hovercards for the landing page selects + items, and the breadcrumbs selects + items IF there is overflow.
* Adds padding around the nav bar navigation items so they don't bump into the logo or actions on the right when the screen is very small

## Related Issues
Fixes #631 

## Testing
I used this patch to test manually, adds a bunch of really long names + sql queries for the query names:
[long-names.patch](https://github.com/user-attachments/files/31488200/long-names.patch)

## Screenshots

https://github.com/user-attachments/assets/56ee46b0-7e6d-49fd-886c-7247ca75a60e


ui/packages/@quent/components/src/ui/overflow-hover-card.tsx
```diff
@@ -0,0 +1,55 @@
...
+      side={side}
+      align="start"
+      className="pointer-events-none w-auto max-w-sm bg-background p-2 text-foreground"
+    >
+      <DataText className="break-all text-xs">{label}</DataText>
+    </HoverCardContent>
+  );
+}
+
+export function OverflowingItemLabel({ label, className }: { label: string; className?: string }) {
```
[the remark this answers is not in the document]
johallar: 2ad604ee6ff490d0a6470cb4d19f4555611e398e

ui/packages/@quent/components/src/ui/select.tsx
```diff
@@ -124,7 +124,9 @@ const SelectItem = React.forwardRef<
       </SelectPrimitive.ItemIndicator>
     </span>
 
-    <SelectPrimitive.ItemText>{children}</SelectPrimitive.ItemText>
+    <span className="min-w-0 flex-1 truncate">
```
cmatzenbach, line 127: Easy win, nice

ui/src/components/NavBarNavigator.tsx
```diff
@@ -97,9 +126,25 @@ export function NavBarNavigator() {
     return null;
   }
 
+  const engineItems =
+    engines?.map(engine => ({
+      id: engine.id,
+      label: engine.instance_name ?? engine.id,
```
cmatzenbach, line 124: Should we be falling back to `id` on these or just rendering something like "Not available" or "None"? Thinking through what it would be like if a user didn't populate these in the backend - they would see the `id` duplicated and probably think it was a bug, rather than understanding that `label` is a separate field and they don't have it populated. I can see the advantages of rendering it - always present, easy fallback - but I'm not sure repeating it shows the user anything worthwhile and may obscure missing information.

ui/src/components/NavBarNavigator.tsx
```diff
@@ -30,27 +35,51 @@ function BreadcrumbDropdown({
   items: { id: string; label: string }[] | undefined;
   onSelect: (id: string) => void;
 }) {
+  const triggerRef = useRef<HTMLButtonElement>(null);
```
cmatzenbach: This re-implements `OverflowingItemLabel` by hand instead of just reusing it - we're already importing it in `NavBarNavigator` for the item labels. `EngineSelectionPage.tsx` has yet a third version of this same "tooltip on truncated text" logic with a different hover delay (that one is 300ms vs the ~700ms default here). Can we consolidate these into one shared hook/component so the behavior and timing are consistent?

ui/src/components/NavBarNavigator.tsx
johallar: ID will be unique per engine wont it?

ui/src/components/NavBarNavigator.tsx
cmatzenbach: Consolidated into shared hook

cmatzenbach (approved):

PR #638, opened 2026-08-27
feat(codegen, vibe): add schema-based C++ and Python generators

johanpel:
# Description

Adds experimental schema-driven C++ and Python instrumentation generators under `experimental/vibe/codegen`.

The generators consume `quent_schema::Schema` directly and live in an isolated Cargo workspace with their own lockfile, examples, tests, and build support.

## Generated APIs

The C++ generator emits internal CXX bridges and a public `quent.hpp` façade with:

- Scoped, copyable observer pointers.
- Strongly typed entity IDs.
- Entity-specific `Handle<Entity>` specializations.
- `FsmHandle<Entity, State>` typestate specializations that expose only legal transitions.
- Generated records, event payloads, optional values, lists, UUIDs, and typed entity references.
- Standard C++ public types while keeping CXX representations internal.
- Configurable no-op, file, and collector exporters.

FSM transitions consume the source handle and return the target-state specialization. Generated-name validation rejects schemas that would produce conflicting bridge or façade symbols.

The Python generator emits a PyO3 module and PEP 561 type information with:

- Pythonic contexts, observers, and generated entity handles.
- State-specific FSM handle classes.
- Standard-library `uuid.UUID` values.
- Keyword-only event attributes.
- Mapping-based records and entity-reference payloads.
- Typed exporter constructors and generated exception classes.
- Runtime protection against reusing consumed FSM handles.

## Dynamic attributes

Both languages support the complete dynamic-value model: every numeric width, strings, nulls, structures, typed lists, structure lists, and recursively nested lists.

C++ provides width-specific `DynamicAttributes` methods and a move-only `DynamicList` builder. Python provides natural conversions plus `DynamicValue` constructors when exact types are required.

Attribute insertion order, nested structure order, and list element order are preserved through serialization.

## Examples and build integration

Adds equivalent C++ and Python README examples and independently buildable versions of every tutorial under `crates/yaml/examples`. Tutorial builds reference the canonical YAML schemas directly rather than copying them.

CI builds and exercises the generated bindings and tutorials. It also covers configurable Cargo target directories and Linux/macOS archive linking.

## Testing

Validated with:

- Full locked workspace tests.
- Clippy with warnings denied.
- Rust formatting checks.
- C++ compilation and integration tests.
- Python runtime and type checks.
- Markdown and diff checks.

Coverage includes FSM type constraints, dynamic-value fidelity and ordering, recursive lists, moved-from builder reuse, schema-list
conversions, generated-name collisions, and stale generated-file cleanup.

_Written by Codex._


.github/workflows/cpp.yml
```diff
@@ -61,3 +63,21 @@ jobs:
...
+        run: pixi run cargo test --manifest-path experimental/vibe/codegen/Cargo.toml -p quent-demo-cpp-test -p quent-cpp-list-test
+      - name: Build and run schema codegen C++ tutorials
+        run: |
+          for manifest in experimental/vibe/codegen/cpp/example/tutorial/*/bridge/Cargo.toml; do
+            tutorial="$(basename "$(dirname "$(dirname "$manifest")")")"
+            target_dir="${CARGO_TARGET_DIR:-experimental/vibe/codegen/target}"
+            library="${target_dir}/debug/libquent_tutorial_${tutorial//-/_}_cpp_bridge.a"
+            executable="${RUNNER_TEMP}/quent-tutorial-${tutorial}"
+            pixi run cargo build --manifest-path "$manifest" --locked
+            if [[ "${RUNNER_OS}" == "macOS" ]]; then
```
mbrobbel: We use `ubuntu-latest`?

.github/workflows/rust.yml
```diff
@@ -66,6 +66,9 @@ jobs:
       - run: pixi run cargo fmt --manifest-path crates/yaml/examples/Cargo.toml -- --check
       - run: pixi run cargo clippy --manifest-path crates/yaml/examples/Cargo.toml --bins --locked -- -D warnings
       - run: pixi run cargo check --manifest-path crates/yaml/examples/Cargo.toml --bins --locked
+      - run: pixi run cargo fmt --manifest-path experimental/vibe/codegen/Cargo.toml --all -- --check
+      - run: pixi run cargo clippy --manifest-path experimental/vibe/codegen/Cargo.toml --workspace --all-targets -- -D warnings
+      - run: pixi run env LD_LIBRARY_PATH="$PWD/.pixi/envs/default/lib" cargo test --manifest-path experimental/vibe/codegen/Cargo.toml --workspace --locked
```
mbrobbel: 🤨 why do we need `LD_LIBRARY_PATH="$PWD/.pixi/envs/default/lib"`?

experimental/vibe/codegen/cpp/example/readme/bridge/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#[allow(unused, clippy::all)]
+mod bridge {
```
mbrobbel, line 5: Nit: maybe we can add something like https://docs.rs/tonic/latest/tonic/macro.include_proto.html?

experimental/vibe/codegen/cpp/example/tutorial/finite-state-machine/main.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "quent-tutorial-finite-state-machine-cpp-bridge/gen/quent.hpp"
+
+int main() {
+  auto context = quent::Context::none();
+
+  quent::FsmHandle<quent::Job> job = context.job_observer()->handle();
+
+  // A transition consumes its move-only state handle and returns the handle
+  // type for the new state. Named C++ objects require std::move to be consumed.
+  quent::FsmHandle<quent::Job, quent::job_state::Queued> queued =
```
mbrobbel, line 15: Would `auto` work here?

experimental/vibe/codegen/cpp/src/facade.rs
```diff
@@ -0,0 +1,1343 @@
...
+      value_ = std::move(other.value_);
+      other.value_.root_count = 0;
+    }}
+    return *this;
+  }}
+
+  void add_null(std::string key) {{
+    append(make_attribute(detail::DynamicAttributeKind::Null, std::move(key)));
+  }}
+  void add_u8(std::string key, std::uint8_t value) {{
```
mbrobbel: Should we do function overloading for this instead?

experimental/vibe/codegen/cpp/src/lib.rs
```diff
@@ -0,0 +1,923 @@
...
+                F32List,
+                F64List,
+                StringList,
+                StructList,
+                ListList,
+                Bool,
+            }
+
+            #[derive(Debug)]
+            pub struct DynamicAttribute {
```
mbrobbel: 😬

mbrobbel (approved):
Nice :shipit:

experimental/vibe/codegen/cpp/example/tutorial/finite-state-machine/main.cpp
johanpel: It would work, but I intentionally made the types explicit here so users can see that the handle type changes between transitions (unless it's a self-transition of course). This matches the Rust tutorial. This is an important difference between `FsmHandle`s and regular `Handle`s.

experimental/vibe/codegen/cpp/example/readme/bridge/src/lib.rs
johanpel: This has some caveats because right now codegen is only a build dep so we can't use any macros it would provide here. Will park this for later.

mbrobbel: Yes this would have to be provided via another crate.\

johanpel: https://github.com/rapidsai/quent/issues/728

.github/workflows/cpp.yml
johanpel: Whoops, removed in https://github.com/rapidsai/quent/pull/638/commits/75d9f634d4487467e14d242c359c35c27b3e3b18

.github/workflows/rust.yml
johanpel: This was because we were running Python test from within Rust tests, which seems a bit convoluted, so I've isolated those Python tests to be tested through regular pytest here: https://github.com/rapidsai/quent/pull/638/commits/8392f41cbb354b985f8ef0cd84e1122bf25b744a

experimental/vibe/codegen/cpp/src/lib.rs
johanpel: Perhaps a lesser evil: https://github.com/rapidsai/quent/pull/638/commits/6eaf2e206dedc153a08be387c8d8b1731cead53e

experimental/vibe/codegen/cpp/src/facade.rs
johanpel: https://github.com/rapidsai/quent/pull/638/commits/bde170e2b031282d8d87dbf50301301524b18357

PR #642, opened 2026-08-28
fix(ui): hide the entity lane when a resource has zero utilization

### LINKED ISSUE #640: feat(ui): hide "long entity" lane if there is no resource activity at all
### Summary

Image

If binned resource utilization data is completely 0 within some window of time, then there are also no usages from entities, in which case zooming in will never expose more entities. We should hide the full entity lane in this case as it takes up a lot of vertical space and the text currently implies there might be something there while there isn't.

Fixing this first would be great already. We can furthermore add a counter for "short entities" to bins, such that we can draw something visually in that full entity lane to make users understand where they need to zoom in. Like a vertical line or something with an increasingly visible gradient relative to the other bins and depending on the amount of "short entities" within a bin.

cmatzenbach:
# Description

- Hide the "Entities" row (label + lane) entirely for resources whose binned utilization is all-zero in the current zoom window, reusing the utilization data already fetched for the sparkline chart - no new network requests.
- Rows with non-zero utilization but no entities above the current long-entities threshold are unaffected - the "zoom to see more" message still applies there, since zooming can still surface shorter entities

## Related Issues

Closes: #640 

## Testing

1. Open a query's profile page, expand the resource tree, confirm the Entities lane appears under a resource with utilization
2. Zoom into a time window where a resource's utilization is completely flat/zero, confirm its Entities row disappears entirely (no label, no "No Matching Entities" text, no reserved space)
3. Zoom back out to the full window, confirm the Entities row reappears for that resource
4. Zoom into a window with utilization but only entities shorter than the threshold, confirm the "No Matching Entities / Showing entities..." message still shows
5. Continue zooming in on that same window until entities exceed the threshold, confirm the Entities lane populates with entries
6. Check a resource group row (not just a single resource) behaves the same as steps 2-3
7. Collapse and re-expand the resource tree, confirm Entities rows still hide/show correctly without stale state
8. Switch between two different resources in the tree, confirm each one's Entities row hides/shows independently based on its own utilization

## Screenshots

https://github.com/user-attachments/assets/cc3c25c3-8999-4787-bf29-27fc58d575f5


johallar:
To me this is a tradeoff because timelines jump up and down when zooming/panning as data goes on/off the screen, making it quite hard to track what you were looking at. I would like to take a step to at least mitigate that before merging, open to ideas but heres one:

Rows with no timeline data cause this jumpyness especially, because the entities skeleton appears below when users zoom or pan which pushes timelines below it down while loading, and then jumps back up when its done loading. 

So one thing we could do is not show the entities skeleton loader (or attempt to reload entities at all maybe) when the resource timeline above has no usage.

https://github.com/user-attachments/assets/47dcb57d-7c14-463e-b136-db8b3f58455d

cmatzenbach:
@johallar Ok, refactored this so now we only strip out entity rows that don't have any entity data at all for the query duration. So we now have three states (aside from the one actually displaying entities):
  - Full-duration zero entities -> row is hidden entirely (new behavior)
  - Current-window zero entities, but some usage elsewhere in the query -> "No entities in this time range" (new message)
  - Current-window non-zero utilization, but no entities above the current threshold -> the old "No Matching Entities / Showing entities longer than X. Zoom to see more." message (unchanged)

johallar (commented):
🙇 thx for all the changes, looks great!

johallar (approved):
🙇 thx for all the changes, looks great!

PR #595, opened 2026-08-21
feat(ui): multi-operator selection

johallar:
# Description
BE allows filtering by multiple operators, and this is used to filter by high level operators. This takes it to the next step and allows users to select multiple operators in any view (DAG, timeline) to filter timelines. 

## Related Issues
N/A

## Testing
* In DAG view, click to select multiple operators in one plan. Try logical and physical plans. 
* Timelines on the right should be filtered by all selected operators
* Operator badges should appear in the filter area above timelines, Clear all button should be present. 
* When many operators are selected, they are truncated in the timeline toolbar

Screenshot 2026-08-24 at 12 42 14 PM

* Select/deselect operators from the timeline view
* Node detail panel below dag should be populated with _all_ selected operator stats. Changes were also made to color code + nest these operators
Screenshot 2026-08-24 at 1 11 03 PM

## Screenshots

https://github.com/user-attachments/assets/feee0474-1a3b-427a-9973-a8128bc97805

PR #654, opened 2026-08-31
feat(ui): entities tab operator filtering, multiple operator selection, dropdown descriptions

cmatzenbach:
# Description

1. Adds context to each value in the "Operator" and "Resource" dropdowns (in the Entities tab). Previously, just the operator/resource name was shown, and it was impossible to distinguish options with the same name; now, plan (operator dropdown only) and worker information has been added to each option, allowing users to differentiate.
2. Ties in global operator filtering in with the Entities table. Now any selections made within the "Operator" dropdown on the Entities tab will be reflected in the DAG, timelines, pivot table, etc
3. Changes the Operator dropdown to use a multi-select, in order to choose multiple operators (and support multiple operator filters applied elsewhere in the app).
4. Small UI fixes - aligns field styles more closely with those from Query Plan Settings section, including adding the clear buttons where necessary, matching hover styles, and icon size tweaks.

## Testing

1. Go to Entities tab, open "Operator" searchable select dropdown
2. Ensure `plan` and `worker` information has been added in the new `description` field
3. Use the search bar to narrow down the list by one of the new fields in the `description`
4. Check the Resource dropdown, you should see a description added there as well
5. Select multiple operators from list; ensure this crossfilters all relevant components
6. Click to filter by operator(s) elsewhere in the app; ensure this also filters the entities tab 

## Screenshots

Dropdown changes:

https://github.com/user-attachments/assets/3e901e53-d191-4f13-bd9d-3c8011fb5f99

Operator crossfiltering:

https://github.com/user-attachments/assets/2401bebf-4ad5-47ec-9343-6863623f7aea

### DIFF
--- a/ui/packages/@quent/components/src/ui/option-multi-select.tsx
+++ b/ui/packages/@quent/components/src/ui/option-multi-select.tsx
@@ -33,8 +33,8 @@ interface OptionMultiSelectProps {
   showSelectedBadges?: boolean;
   className?: string;
   triggerClassName?: string;
-  /** Render option labels in a monospace font, e.g. for raw field/column identifiers. Defaults to true. */
-  monospaceLabels?: boolean;
+  /** className applied to each option's label and its selected badge. */
+  optionClassName?: string;
 }
 
 function optionLabel(option: OptionMultiSelectOption): string {
@@ -57,7 +57,7 @@ export function OptionMultiSelect({
   showSelectedBadges = true,
   className,
   triggerClassName,
-  monospaceLabels = true,
+  optionClassName,
 }: OptionMultiSelectProps) {
   const [search, setSearch] = useState('');
 
@@ -199,7 +199,7 @@ export function OptionMultiSelect({
                     {checked && <Check className="size-2.5" strokeWidth={3} />}
                   </span>
                   <span className="min-w-0 flex-1 text-left">
-                    <span className={cn('block truncate', monospaceLabels && 'font-mono')}>
+                    <span className={cn('block truncate', optionClassName)}>
                       {optionLabel(option)}
                     </span>
                     {option.description && (
@@ -229,7 +229,7 @@ export function OptionMultiSelect({
                   variant="outline"
                   className={cn(
                     'px-1.5 py-0 text-data bg-primary/10 border-primary/40 hover:bg-primary/15',
-                    monospaceLabels && 'font-mono'
+                    optionClassName
                   )}
                 >
                   <span className="truncate">{optionLabel(option)}</span>
--- a/ui/packages/@quent/components/src/ui/option-multi-select.tsx
+++ b/ui/packages/@quent/components/src/ui/option-multi-select.tsx
@@ -9,10 +9,19 @@ import { Button } from './button';
 import { Input } from './input';
 import { Popover, PopoverContent, PopoverTrigger } from './popover';
 
+export interface OptionMultiSelectOption {
+  value: string;
+  label?: string;
+  description?: string;
+}
+
 interface OptionMultiSelectProps {
-  label: string;
+  /** Prefix label rendered before the trigger, e.g. "Columns:". Omit when the field already has an external label (e.g. via a wrapping <FieldWrapper>). */
+  label?: string;
+  /** Accessible label for the trigger button when the visible label is rendered externally. */
+  ariaLabel?: string;
   triggerText: string;
-  options: string[];
+  options: OptionMultiSelectOption[];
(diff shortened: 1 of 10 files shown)


johallar:
One thought about the operator dropdown: I think the operator crossfilter should sync everywhere or not. It threw me a little bit that when you select a filter from the DAG it selects it in the entities toolbar, but if you select from the entities toolbar it doesn't select it everywhere else.

eg. If you click an operator box from the operator swimlanes timeline chart it selects it in the dag and operator table views

johallar:
We'll want to plan for multi operator selection as well (https://github.com/rapidsai/quent/pull/595), which it looks like the entities API allows. I think this needs to become a multi-select, and should be synced with the current selected operators crossfilter.

ui/packages/@quent/components/src/ui/option-multi-select.tsx
```diff
@@ -37,13 +55,20 @@ export function OptionMultiSelect({
   noneSelectedText = 'None selected',
   maxVisibleBadges = 6,
   showSelectedBadges = true,
+  className,
+  triggerClassName,
+  monospaceLabels = true,
```
johallar: Too specific for this component, would prefer `labelClassName` to go along with `className` + `triggerClassName`

ui/packages/@quent/components/src/ui/option-multi-select.tsx
```diff
@@ -114,33 +172,42 @@ export function OptionMultiSelect({
...
+                  key={option.value}
                   type="button"
                   role="option"
                   aria-selected={checked}
                   data-state={checked ? 'checked' : 'unchecked'}
-                  onClick={() => onToggleOption(option)}
+                  onClick={() => onToggleOption(option.value)}
                   className={cn(
-                    'relative flex w-full cursor-pointer select-none items-center gap-2 rounded-sm px-2 py-1 text-xs font-mono outline-none',
+                    'relative flex w-full cursor-pointer select-none items-start gap-2 rounded-sm px-2 py-1 text-xs outline-none',
```
johallar: prefer this centered

ui/packages/@quent/components/src/ui/option-multi-select.tsx
```diff
@@ -158,18 +225,21 @@ export function OptionMultiSelect({
             <div className="flex flex-wrap items-center gap-1">
               {visibleSelectedOptions.map(option => (
                 <Badge
-                  key={option}
+                  key={option.value}
                   variant="outline"
-                  className="px-1.5 py-0 font-mono text-data bg-primary/10 border-primary/40 hover:bg-primary/15"
+                  className={cn(
+                    'px-1.5 py-0 text-data bg-primary/10 border-primary/40 hover:bg-primary/15',
+                    monospaceLabels && 'font-mono'
```
johallar: Oh i see its in a bunch of places and defaulted to `font-mono`, lemme think about this one

johallar: Still doable as suggested I think, `optionClassName` would just get added here: https://github.com/rapidsai/quent/pull/654/changes#diff-e4dcb9cf3de94b29ab5ae7236dd2c33f86dbcf3228a2eda65efbf55d70cab5fcR184

ui/packages/@quent/components/src/ui/option-multi-select.tsx
```diff
@@ -52,12 +77,27 @@ export function OptionMultiSelect({
...
-    <div className="flex items-center gap-1 px-3 py-1.5 border-t border-border/50">
-      <span className="text-xs text-muted-foreground shrink-0 mr-1">{label}:</span>
+    <div
+      className={cn(
+        'flex items-center gap-1',
+        label ? 'px-3 py-1.5 border-t border-border/50' : 'min-w-0',
+        className
+      )}
+    >
+      {label && <span className="text-xs text-muted-foreground shrink-0 mr-1">{label}:</span>}
```
johallar: My fault, but could/should use `<label>` here

johallar (approved):
Looks good! Couple ideas for option multi select to keep it max flexible

ui/packages/@quent/components/src/ui/option-multi-select.tsx
cmatzenbach: agreed, i prefer it centered between the two lines as well. the only justification for `start` would be if some items don't have a second line, as then it wouldn't be consistent among all the options, but since for our current use they all have description lines, I think `center` will work fine.

PR #658, opened 2026-09-01
feat(components): improve operator inspection UI

johallar:
# Description

* Adapt [DAGNodeInfoPanel.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx) to use main’s useSelectedNodeData() while adding parent/child accordions, operation-type bars, nested stats, and nested data-flow matrices.

* Include panel tests in [DAGNodeInfoPanel.test.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.test.tsx), optional zero-filled frames in [DataFlowMatrix.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/dag/DataFlowMatrix.tsx), and higher-level/data-flow coverage in [DataFlowOverlay.test.tsx](https://github.com/rapidsai/quent/pull/ui/src/components/DataFlowOverlay.test.tsx).

* Reset per-operator collapse state when the selected root operator changes, so closed children do not leak between selections.

* Do not include multi-selection atoms, toolbar changes, DAG click changes, or multiple top-level panel titles.

* Extract the pure opacity helper and tests in [nodeOpacity.ts](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/query-plan/nodeOpacity.ts) and [nodeOpacity.test.ts](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/query-plan/nodeOpacity.test.ts), then use it from [QueryPlanNode.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/query-plan/QueryPlanNode.tsx).

* Include only the edge-interaction dimming hunk from [DAGChart.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/dag/DAGChart.tsx); defer all click/selection-state rewrites.

## Testing
Result after merge: one selected logical operator still behaves as on main, its physical operators are nested and color-coded, and selected nodes/edges remain visible during hover highlights.

Selected node(s) remain highlighted when hovering others:

https://github.com/user-attachments/assets/4d1d148f-4b9f-4355-b1e9-e01444d6aa57

Nested node detail panel on higher level operator selection:

https://github.com/user-attachments/assets/7dd449e6-648c-4791-87b7-6f4177d87502

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
+++ b/ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { useEffect, useState } from 'react';
+import { useEffect, useState, type ReactNode } from 'react';
 import { ChevronUp, ChevronDown } from 'lucide-react';
 import {
   useSelectedNodeData,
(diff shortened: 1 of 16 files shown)


ui/packages/@quent/components/src/query-plan/nodeOpacity.test.ts
```diff
@@ -0,0 +1,31 @@
...
+        hoveredStatValues: null,
+        highlightedNodeIds: new Set(['hovered']),
+        operatorId: 'selected',
+        isDimmed: false,
+        isSelected: true,
+      })
+    ).toBe('opacity-100');
+  });
+
+  it('dims nodes that are neither selected nor hovered', () => {
```
cmatzenbach, line 20: I think there's couple additional cases we should test for here:
 - dims an edge that is neither selected nor highlighted
- keeps an edge touching a selected node visible during an active highlight elsewhere

The latter seems very specific, but I think it's worthy of a test because this PR changes that functionality (for the better) and it'd be good to ensure future changes don't revert it. Might be worth pulling `dimFromInteraction`'s logic into a small helper (same pattern as `getNodeOpacityClass` in `nodeOpacity.ts`) so it's easy to unit test directly rather than through a full DAG render.

ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
```diff
@@ -109,18 +68,31 @@ export const DAGNodeInfoPanel = ({
...
-          operator={operator}
-          quantitySpecs={quantitySpecs}
-          showHeader
-        />
-      ))}
+      <OperatorDetailsBlock
+        operator={selectedNodeData}
+        quantitySpecs={quantitySpecs}
+        isOpen={isOperatorOpen}
+        onOpenChange={setOperatorOpen}
```
cmatzenbach, line 75: Since `setOperatorOpen` is passed down to both `OperatorDetailsBlock` and `OperatorDataFlowBlock` and shares the same state, collapsing an item on one tab automatically collapses it on the other tab as well. My question is, was this intentional? I personally don't have a strong opinion on which is better UX (I can see valid arguments for both), so you can decide which you prefer, just wanted to point that outl.

cmatzenbach (approved):
A couple small items, but nothing that holds it back from approval. Love the new components! They really help to clean up the parents and keep it more focused.

ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
johallar: Good question I hadn't fully thought through. I think keeping the same selections _does_ make sense... if you're looking at specific operators in the details tab you'd want to see the same in the data flow tab. 

This may be something to get feedback on in the future but I think makes sense for now.

ui/packages/@quent/components/src/query-plan/nodeOpacity.test.ts
johallar: https://github.com/rapidsai/quent/pull/658/commits/008f13a4ac241178d34bc78fd23e5b4cab680c37

PR #670, opened 2026-09-04
feat(simulator): deploy the simulator example to GitHub pages

mbrobbel:
This enables anyone to try the quent simulator example without building it from source. It deploys the example and runs the "server" with wasm in the browser.

- Add a static simulator demo that runs the analyzer in a Web Worker using WASM.
- Share a typed server contract between the native server and browser implementation.
- Generate the Postcard demo recording and WASM bundle during the build.
- Deploy the simulator alongside the Schema Explorer on GitHub Pages.
- Add a dedicated simulator UI entrypoint with hash-based routing.


examples/simulator/wasm/src/bin/generate.rs
```diff
@@ -0,0 +1,42 @@
...
+    let output = PathBuf::from(args.next().ok_or("usage: generate-demo <output>")?);
+    if args.next().is_some() {
+        return Err("usage: generate-demo <output>".into());
+    }
+
+    let events = Arc::new(Mutex::new(Vec::<Event<SimulatorEvent>>::new()));
+    let callback = {
+        let events = Arc::clone(&events);
+        EventCallback::new(move |event| events.lock().unwrap().push(event))
+    };
```
johanpel, line 26: This seems a little bit convoluted because we could also just export with the postcard exporter directly and drop the context to flush it.

johanpel (approved):
Rust changes LGTM but one thing we might want to simplify:

ui/packages/@quent/client/src/client.ts
```diff
@@ -0,0 +1,69 @@
...
+  ): Promise<EntityListResponse>;
+  fetchDataFlow(
+    engineId: string,
+    queryId: string,
+    config: TimelineConfig,
+    measures?: string[]
+  ): Promise<DataFlowTimelineBinned | null>;
+}
+
+let client: ApiClient | undefined;
```
johallar: This is very cool! Two changes i'd like to see:
* Set `client` up with a closure to avoid relying on a global
* Create an `httpApiClient`, `getApiClient` will return http or wasm client instead of falling back to `apiFetch` when client is falsey.

Gonna make a patch real quick

johallar: [http-client.patch](https://github.com/user-attachments/files/31845212/http-client.patch)

Something along these lines, or I can push stuff if easier

mbrobbel: Thank you! Feel free to push to my branch.

cmatzenbach (approved):

ui/src/lib/simulatorClient.ts
```diff
@@ -0,0 +1,98 @@
...
+    fetchEngineContexts: engineId =>
+      call<EngineContexts>({ operation: 'engineContexts', engineId }),
+    fetchListCoordinators: engineId =>
+      call<QueryGroup[]>({ operation: 'listCoordinators', engineId }),
+    fetchListQueries: (engineId, queryGroupId) =>
+      call<Query[]>({ operation: 'listQueries', engineId, queryGroupId }),
+    fetchQueryBundle: (engineId, queryId) =>
+      call<QueryBundle<EntityRef>>({ operation: 'queryBundle', engineId, queryId }),
+    fetchSingleTimeline: (engineId, request) =>
+      call<SingleTimelineResponse>({ operation: 'singleTimeline', engineId, request }),
```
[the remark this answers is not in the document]
johallar: We should remove from the http endpoint in a follow up https://github.com/rapidsai/quent/issues/675

johallar (approved):
🆒

PR #673, opened 2026-09-04
feat(ui): [3] multi operator select ux

johallar:
# Description

Lotta lines spent just changing language from node -> operator and from inspected -> selected to keep things consistent across all this new code

## Design Overview
This consolidates all operator selection logic into these main parts:

```
onst operatorSelection = useOperatorSelection();
const updateOperatorSelection = useOperatorSelectionActions();

// Derived inspection/presentation data
useSelectedOperatorIds() — union of every concrete operator ID represented by all selection groups.
useSelectedOperatorsData() — detail data for selected operator groups, including statistics and related operators. It reads separately stored state synchronized with operator selection updates.
```

### `updateOperatorSelection`
This is the value returned by `useOperatorSelectionActions()`. All reusable components should mutate selection through it.

Actions:

`add` — adds/collapses a selection and stores its inspected data.
`remove` — removes an exact selectionId group.
`replace` — atomically rebuilds groups after hierarchy changes.
`hydrate` — supplies inspected data without changing selection.
`clear` — clears both selection and inspected data.

Then:

* Use `selectedOperatorIds` for highlighting/filtering.
* Use `operatorSelection.selections` for badges/groups.
* Dispatch add with inspectedData for a new standalone selection.
* Use hierarchy `toggleOperatorSelection` plus replace for parent/child toggling.
* Use `useSelectedOperatorsData()` only when rendering inspection details.

## Code Changes

* Switch DAG and operator-timeline clicks to toggle complete logical selections in [DAGChart.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/dag/DAGChart.tsx) and [OperatorGanttChart.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/operator-timeline/OperatorGanttChart.tsx); pane click and Clear remove all selections.

* Ship toolbar representation atomically: add [truncated-badge-list.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/ui/truncated-badge-list.tsx), its tests/export, update [QueryToolbar.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/timeline/QueryToolbar.tsx) for bounded badges and per-item removal, and reuse the primitive in [option-multi-select.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/ui/option-multi-select.tsx).

* Adapt PR 1’s panel to useSelectedNodesData() in [DAGNodeInfoPanel.tsx](https://github.com/rapidsai/quent/pull/ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx), retaining nested higher-level operators while allowing multiple top-level accordions and titles.

* Include the corresponding toolbar, panel, DAG, action-atom, and selection tests. Keep parent/child dedupe in this activation path so users never see a temporary duplicate-badge state.

## Testing
All selected operators should rely on one source of truth state (`useOperatorSelection`) and be mutated through one hook `useOperatorSelectionActions`. This should act as a global crossfilter.
* Select/Deselect operators from any of the following locations: DAG, Query Toolbar, Resource operator swimlanes, entities multiselect
* "Clear" operator filters from the query toolbar - expect to clear all selected operators always
* Click negative space in DAG to clear all selections (not sure if we should keep this, but this is pre-existing)
* Click "None" in entities toolbar multi select to clear all selected operators

### Hierarchical select/deselect
* Select a parent operator (from logical plan in simulated query)
* Navigate to and deselect a child operator in the DAG view, resource timeline DAG swimlanes, entities operator multi select.
* Parent operator should "deselect", any remaining child operators should remain selected. This will "explode" the parent badge into the remaining children in the query toolbars filters list.

## Screenshots

https://github.com/user-attachments/assets/4c616aa8-92ab-48df-b075-1c2b1aaa7325


ui/packages/@quent/components/src/ui/option-multi-select.tsx
```diff
@@ -222,10 +221,14 @@ export function OptionMultiSelect({
           {selectedOptions.length === 0 ? (
             <span className="text-xs text-muted-foreground italic">{noneSelectedText}</span>
           ) : (
-            <div className="flex flex-wrap items-center gap-1">
-              {visibleSelectedOptions.map(option => (
+            <TruncatedBadgeList
```
cmatzenbach, line 224: Has this integration been tested out? The only usage of `OptionMultiSelect` is in `EntityToolbar` and `ResourceFilterSearch`, neither of which has `showSelectedBadges` as `false`. I set it to `true` and the UI isn't bad, but may not be what is desired either (it stacks everything vertically next to the dropdown). I thought maybe it would stack them horizontally and truncate the names (similar to its use in `QueryToolbar`). Is there something planned for this by adding it in?

ui/packages/@quent/hooks/src/dag/inspectedNodeData.ts
```diff
@@ -0,0 +1,25 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { InspectedNodeData } from '@quent/utils';
+
+export function upsertInspectedNodeData(
```
cmatzenbach: Is there a typo in the function name - `upsert`? Or is it a combo of updating and inserting, hence the name?

ui/packages/@quent/hooks/src/dag/operatorSelection.ts
```diff
@@ -0,0 +1,83 @@
...
+  state: OperatorSelectionState,
+  selectionId: string,
+  label: string,
+  operatorIds: Iterable<string>
+): OperatorSelectionState {
+  const selectedIds = new Set(operatorIds);
+  selectedIds.add(selectionId);
+
+  for (const [existingId, existing] of state.selections) {
+    if (existingId !== selectionId && containsAll(existing.operatorIds, selectedIds)) {
```
cmatzenbach, line 37: Is `existing.operatorIds` a `Map` type? It will need to be to work with `containsAll`

cmatzenbach:
I found a bug with selecting logical operators and interacting with them via the gant chart. If you select a logical operator (either via the DAG or via the Entities table), and then go to the timeline and try to click on the operator to de-select it, you are unable to do so - clicking on it does nothing. The attached screencast shows the bug (same thing happens if you select via the DAG). To make life a bit easier (since I figured this might be hard to track down), I asked claude about it, and here's what it said:


"The bar's highlight state checks the flattened selectedNodeIds (includes every operator in every group), but the click handler decides add-vs-remove by checking operatorSelection.selections.has(op.operatorId) (only top-level selection keys). If a logical operator was selected as a group elsewhere (e.g. the Entities table dropdown, which groups parent + physical children into one selection), its child bars render fully highlighted in the timeline — but clicking them does nothing, because addOperatorSelection silently no-ops when the clicked id is already fully covered by an existing group."

https://github.com/user-attachments/assets/68b379be-9fb0-473b-8280-90d5a31013f2

ui/packages/@quent/components/src/dag/dagSelection.ts
```diff
@@ -35,3 +37,42 @@ export function resolveInspectedNodeData(
...
+      label: node.label,
+      operatorIds,
+      inspectedData: inspectNode(node),
+    });
+  }
+
+  return { selections, unresolvedOperatorIds };
+}
+
+export function resolveInspectedNodeData(
```
cmatzenbach: I think this function is now dead - it doesn't seem to be used anywhere besides in tests.

ui/packages/@quent/utils/src/operatorHierarchy.ts
```diff
@@ -0,0 +1,98 @@
...
+      related.add(childId);
+      stack.push(...(childrenByParentId.get(childId) ?? []));
+    }
+    relatedById.set(operatorId, [...related].sort());
+  }
+
+  return relatedById;
+}
+
+export function resolveOperatorSelections(
```
cmatzenbach, line 97: This function duplicates the greedy grouping logic in `resolveInspectedNodeSelections` (dag/dagSelection.ts) - same "collapse into largest covered parent" logic, just over `Operator[]` instead of `DAGNode[]`. Worth extracting a shared generic helper so a future fix to one doesn't silently miss the other.

ui/packages/@quent/components/src/dag/DAGChart.tsx
```diff
@@ -402,46 +430,30 @@ const FlowLayout = ({
 
   const handleNodeClick = useCallback(
     (_event: MouseEvent, node: Node<QueryPlanNodeData>): void => {
-      if (selectedNodeIds.has(node.id)) {
-        setSelectedNodeIds(new Set());
-        setSelectedOperatorLabel(null);
-        setSelectedNodeData(null);
-        onSelectionChange?.([]);
+      let nextSelectedIds: Set<string>;
+      if (operatorSelection.selections.has(node.id)) {
```
cmatzenbach: From claude's response to the bug I found, it looks like this has the same logic which will trigger the same bug that `OperatorGantChart` has - worth fixing here as well.

ui/packages/@quent/hooks/src/dag/inspectedNodeData.ts
johallar: Yeah this comes from database lingo, a combo operator that updates if the thing exists, or inserts it if it doesn't

ui/packages/@quent/components/src/ui/option-multi-select.tsx
johallar: `OptionMultiSelect` is also used in the pivot table to show selected stats for the table, which shows these badges, so this just tries to capture the badge list as a re-usable component.

Image

Entities multiselect passes className of `w-full`, and is displayed flexed with other inputs so the input tries to fill the space it has and the badges stack vertically cause there's no space. 

`maxVisibleBadges` controls the number of badges to show before truncating, so it'll just stack them vertically before adding the "+2 more" truncation badge.

johallar: To use the badges part of this component you'd put it on its own line:

https://github.com/user-attachments/assets/a0054482-5d46-4664-b4bf-7b13d0fdbdbb

Maybe responsibility could be shifted back to the caller to determine where the badges go? Or how they are displayed? not sure, maybe it's just too much in the multi select and it should be composed/controlled 100% by the caller

ui/packages/@quent/hooks/src/dag/operatorSelection.ts
johallar: typescript will enforce this is a `ReadonlySet`

```
export interface OperatorSelectionState {
  readonly selections: ReadonlyMap<string, OperatorSelection>;
  readonly activeId: string | null;
}
```

```
export interface OperatorSelection {
  readonly label: string;
  readonly operatorIds: ReadonlySet<string>;
}
```

```

function containsAll(container: ReadonlySet<string>, contained: ReadonlySet<string>): boolean {
  for (const id of contained) {
    if (!container.has(id)) {
      return false;
    }
  }
  return true;
}
```

ui/packages/@quent/utils/src/operatorHierarchy.ts
johallar: Should be fixed from prior PR @ https://github.com/rapidsai/quent/commit/f2e3f245f65a025c2d73664f5d2f70caf58771fe

lemme know if you're looking for something different though

ui/packages/@quent/components/src/dag/dagSelection.ts
johallar: Found some other dead "single operator select" state/logic, tacking that removal onto here.

ui/src/components/entities-table/EntitiesTable.tsx
```diff
@@ -32,66 +32,68 @@ export function EntitiesTable(props: EntitiesTableProps) {
...
-            hasValidationErrors={table.filters.validationErrors.length > 0}
-            page={table.pagination.page}
-            pageCount={table.pagination.pageCount}
-            pageSize={table.filters.values.pageSize}
-            paginationDisabled={table.pagination.disabled}
-            total={table.pagination.total}
-            visibleStart={table.pagination.visibleStart}
-            visibleEnd={table.pagination.visibleEnd}
-            sortDir={table.filters.values.sortDir}
+    <div className="flex flex-col h-full">
```
johallar, line 35: Dramatic... just moving query toolbar to the top to match the other tabs layouts:
Image

ui/packages/@quent/components/src/dag/dagSelection.ts
johallar: Carried a way a bit, but also normalized language across all of this so its easier to read. Node becomes operator everywhere, and "inspected" becomes "selected"

ui/packages/@quent/components/src/dag/DAGChart.tsx
johallar: Indeed it did, fixed in both places!

johallar:
> I found a bug with selecting logical operators and interacting with them via the gant chart. If you select a logical operator (either via the DAG or via the Entities table), and then go to the timeline and try to click on the operator to de-select it, you are unable to do so - clicking on it does nothing. The attached screencast shows the bug (same thing happens if you select via the DAG). To make life a bit easier (since I figured this might be hard to track down), I asked claude about it, and here's what it said:
> 
> "The bar's highlight state checks the flattened selectedNodeIds (includes every operator in every group), but the click handler decides add-vs-remove by checking operatorSelection.selections.has(op.operatorId) (only top-level selection keys). If a logical operator was selected as a group elsewhere (e.g. the Entities table dropdown, which groups parent + physical children into one selection), its child bars render fully highlighted in the timeline — but clicking them does nothing, because addOperatorSelection silently no-ops when the clicked id is already fully covered by an existing group."
> 
>  logical-ops-gant-chart-bug.mp4

Addressed this as well, whenever you get around to re-review

cmatzenbach (approved):

ui/packages/@quent/components/src/ui/option-multi-select.tsx
cmatzenbach: Yeah I think it might be a good idea to let the caller control that, as it could vary based on the situation. It's ok for now though, if we run into issues we can do a follow-up later.

ui/packages/@quent/hooks/src/dag/operatorSelection.ts
cmatzenbach: Perfect - just wanted to be sure (should've just looked at the type - my bad)

cmatzenbach (approved):

PR #686, opened 2026-09-08
feat(log): add traditional log sink constraint + YAML syntax

johanpel:
# Description

Adds first-class `log:` syntax to `quent-yaml` for declaring entity-scoped logging sinks.

- Lowers ordered levels into repeatable events with an implicit `message: string` field.
- Supports optional target and source fields plus common and level-specific attributes.
- Adds the `quent-log` constraint crate, builder APIs, analyzer inspection APIs, and validation for generated and manually constructed schemas.
- Adds YAML, constraint, and instrumentation-generation coverage.

## Related Issues

Based off #716 which needs to merge first.

## Testing

- `cargo test -p quent-log -p quent-yaml -p quent-instrumentation-build`
- `cargo clippy -p quent-log -p quent-yaml -p quent-instrumentation-build --all-targets -- -D warnings`
- `cargo fmt --all -- --check`

_Written by Codex._

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2915,6 +2915,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "quent-log"
+version = "0.1.0"
+dependencies = [
+ "quent-constraints",
+ "quent-schema",
+ "serde",
+ "serde_json",
+ "thiserror",
+]
+
+[[package]]
 name = "quent-model"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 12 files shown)


crates/yaml/src/extensions/mod.rs
```diff
@@ -75,19 +78,16 @@ impl Elaborator {
...
         let mut records = Vec::new();
         for (name, spec) in &model.fsms {
-            if model.entities.contains_key(name) {
-                sink.error(
-                    &format!("fsms.{name}"),
-                    format!("`{name}` is declared as both an entity and an FSM"),
-                    None,
-                );
-                continue;
-            }
```
johanpel: Removed this because the final schema build will catch these types of errors and if we keep adding shorthands for creating entire entities this will blow up.

mbrobbel (approved):
This is really nice!

PR #690, opened 2026-09-09
refactor(analyzer): decouple FSM storage from legacy events (#191 - 1)

johanpel:
# Description

Introduce `FsmDescriptor<T>` to separate analyzed FSM storage from the legacy `TransitionInfo` trait.

This generalizes `TransitionEvent`, `FsmEventsBuilder`, and `FsmEvents` over a descriptor that provides state names, attributes, resource usages, and runtime declarations. `LegacyFsmDescriptor` preserves the existing macro-generated event path, while schema-generated events can use `push_transition` directly. This is scheduled for removal at the end of the stack.

## Related Issues

PR 1 / ? - final stack of PRs for #191

## Testing

- `pixi run cargo test -p quent-analyzer`

_Written by Codex._


crates/analyzer/src/fsm/events.rs
```diff
@@ -20,18 +20,47 @@ use crate::{
     resource::{CapacityValue, Usage, Using},
 };
 
+/// Interprets model-specific FSM transition data.
+pub trait FsmDescriptor<T> {
```
mbrobbel, line 24: Can you elaborate on the intended use of this trait in the rest of the stack?

johanpel: This is necessary to migrate away from `TransitionInfo` which is defined in the model crate (scheduled for removal at the end of the stack) for which the implementation is currently generated by the legacy model macro crates.

I want to deal with generating the implementation of this trait for FSM event enums separately in a later store/analysis-build crate, but is is not strictly necessary for the migration now since application-specific analyzers can for now hand-roll these.

mbrobbel: I guess I'm still a little bit confused. Can't we capture this as data?

johanpel: As discussed, since I've done pretty much all of the work that could have been done in isolation, I'm going to migrate in one big PR instead of us spending time on these intermediate artifacts. Closing

Issue #696, opened 2026-09-09
Phase 3b: Route NVTX events to multiple contexts in one process


9prady9:
The process-global NVTX hook currently binds capture to the first Quent context. Events from another live context are attributed to the first, while contexts created after the first is destroyed cannot capture NVTX.

## Scope

- Keep one process-global NVTX hook and route events to registered Quent contexts.
- Add safe context registration, unregistration, and a nestable scoped route guard.
- Preserve route ownership across range and resource lifecycles.
- Replay process-wide NVTX metadata to contexts registered later.
- Define deterministic behavior for events without an active route and during teardown.

## Done when

- Concurrent contexts receive only events emitted within their active route scopes.
- Destroying one context does not prevent a later context from capturing NVTX.
- Cross-thread and nested ranges remain assigned to their originating context.
- Later contexts can resolve existing domains, strings, categories, and thread names.
- GPU-independent tests cover concurrent, nested, and sequential context lifetimes.
- Existing single-context behavior remains unchanged.

Related to #618, which covers shared event transport and analysis.

9prady9:
Sirius needs a followup after this one - concerned issue can be found here https://github.com/sirius-db/sirius/issues/1735

PR #697, opened 2026-09-09
fix(nvtx): use nvtx-sys tooling

9prady9:
`nvtx-injection` owned a Conda-specific bindgen path for NVTX's injection ABI even though `nvtx-sys` now bundles the headers and exposes that ABI through its `tools` feature.

## Description

Use `nvtx-sys = { version = "2", features = ["tools"] }` as the source of NVTX producer and injection ABI bindings.

- Use the upstream callback tables, callback IDs, and exact implementation-function aliases for all 26 subscribed CORE/CORE2 callbacks.
- Remove Quent's direct bindgen build dependency, wrapper header, binding allowlist, and generated-bindings maintenance path.
- Remove the external `nvtx-c` package from the x86_64 and aarch64 Pixi environments because `nvtx-sys` ships the NVTX headers it binds against.
- Keep Quent's Linux 64-bit target validation and `static-injection` strong-symbol shim; `nvtx-sys/tools` supplies ABI definitions, not Quent's initializer.
- Preserve raw/future discriminator values while adapting to the upstream generated enums and exact `ResourceCreate`/`wchar_t` callback signatures.
- Update the NVTX documentation to describe upstream binding ownership and downstream toolchain requirements.

This uses the tooling support introduced by [NVIDIA/NVTX#170](https://github.com/NVIDIA/NVTX/pull/170).

### Downstream build requirements

`nvtx-sys` still runs bindgen and native C compilation during consumer builds. Any downstream build graph that includes `nvtx-injection` therefore needs:

- a discoverable `libclang`; and
- a native C compiler.

No external NVTX installation or `CONDA_PREFIX` is required because `nvtx-sys` bundles its headers. The corresponding Sirius distribution-image toolchain update is intentionally handled downstream.

## Related Issues

- Unblocks the distribution build failure exposed by [sirius-db/sirius#1439](https://github.com/sirius-db/sirius/pull/1439) once its build image provides the required native toolchain.
- Deliberately excludes the multi-context routing work tracked by #696.

## Testing

- [x] Clean `nvtx-injection` build with `CONDA_PREFIX` and `LIBCLANG_PATH` unset inside the Pixi environment
- [x] `pixi run --frozen cargo test --locked -p nvtx-injection -p nvtx-example -p nvtx-analyzer --all-features`
- [x] Real NVTX capture round trip and static-injection symbol override
- [x] `pixi run --frozen cargo tree --locked -p nvtx-injection -e normal,build` — confirms the intentional `nvtx-sys -> bindgen + cc` build graph
- [x] `pixi run --frozen cargo fmt --all -- --check`
- [x] `pixi run --frozen cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- [x] `pixi run --frozen cargo test --workspace --all-features --locked --all-targets`
- [x] `pixi run --frozen cargo build --workspace --all-features --locked --release`
- [x] Remaining Rust workflow checks, including the NDJSON-only store tests, build examples, and YAML examples
- [x] `pixi run --frozen cargo deny check`
- [x] `pixi run --frozen rumdl check --diff`
- [x] `pixi run --frozen bash docs/tutorial/build.sh`

A native aarch64 Rust compile was not available locally. The Linux-aarch64 Pixi environment retains `libclang` plus Clang's compiler resource headers for bindgen.

## Screenshots

N/A — build-tooling change only.


johanpel:
What about switching over to:

```
nvtx-sys = { version = "2", features = ["tools"] }
```

?

johanpel:
Looking at the CI failure in Sirius a bit more closely, I think the best course of action is:
- as mentioned above, use nvtx-sys so we don't have to check in anything generated by bindgen
- add the required dependencies of running bindgen to the build env of Sirius
- make any Conda-related shenanigans depend on the presense of CONDA_PREFIX only instead of requiring a Conda env, but I don't expect any since NVTX is header only and `nvtx-sys` ships the headers

9prady9:
In general, its good point to switch over to nvtx-sys.

> Looking at the CI failure in Sirius a bit more closely, I think the best course of action is:
> 
> * as mentioned above, use nvtx-sys so we don't have to check in anything generated by bindgen
> * add the required dependencies of running bindgen to the build env of Sirius
> * make any Conda-related shenanigans depend on the presense of CONDA_PREFIX only instead of requiring a Conda env, but I don't expect any since NVTX is header only and `nvtx-sys` ships the headers

One important trade-off is that nvtx-sys unconditionally runs bindgen and cc, so  this requires libclang and a C compiler in **every downstream build environment** - sirius in this case - but any new downstream quent user else too if they want to use nvtx-injection.

Commited code would make it less work from build perspective though. <ins>I understand it can drift and requires maintenance but will only happen when we update nvtx version, it is not automatic or time bound</ins>.

IMO, the ideal longer-term solution would be for nvtx-sys itself to offer pregenerated bindings or a feature that disables bindgen. Then Quent could depend on it normally without imposing libclang on consumers.

We can add nvtx-sys as is to move this forward, but it will require changes in downstream build envs - there is not doubt about that.

johanpel:
Since the only work from a build perspective required is to add one line with `libclang` to Sirius' `pixi.toml`, this would be my distinct preference.

I am not in favor of adopting solutions for every transitive dependency problem from every third-party event source dependency to every downstream consumer if we can help it, especially not if there are simple solutions either up or downstream.

9prady9:
> Since the only work from a build perspective required is to add one line with `libclang` to Sirius' `pixi.toml`, this would be my distinct preference.

I have already made the change. However, I would like to point out that even though it sounds like a single line change - its a build env change in Quent nvtx injection users we are forcing. Among others, here are few issues future downstream consumers may encounter:
  - consumers already building another crate that depends on nvtx-sys may have already satisfied these requirements, so their incremental burden can be zero. That cannot be assumed for every nvtx-injection consumer.
  - cross-compilation case: an x86-64 consumer targeting aarch64-unknown-linux-gnu needs host libclang, an ARM64 sysroot/libc headers, target compiler and archiver, and linker configuration. Without the sysroot, parsing the bundled header definitely causes an issue - not specific to this use case.
  - Bindgen/cc discovery might depend on LIBCLANG_PATH sometimes.

> I am not in favor of adopting solutions for every transitive dependency problem from every third-party event source dependency to every downstream consumer if we can help it, especially not if there are simple solutions either up or downstream.

`A dependency that can potentially cause build issues` vs `203 lines of generated code without dependency` - I would call that a trade-off not a precedent for handling other transitive dependencies.


We can go ahead with nvtx-sys; I just wanted to highlight that it shifts real build complexity downstream and is not quite as simple as it may appear.

johanpel (approved):
Thanks for greatly reducing some real build complexity, LGTM!

PR #698, opened 2026-09-10
feat(dynamic-attributes)!: support recursive lists and generic insertion

### LINKED ISSUE #79: Support nested lists in attributes
We currently support nesting compound types through structs, e.g.: `list<struct<x: list<...>, ...>>`, but not `list<list<x>>` yet.

### LINKED ISSUE #694: feat(dynamic-attributes): add generic value insertion
Provide a generic key-value insertion API for DynamicAttributes so callers do not need type-specific methods such as add_string and add_u64. Define conversions for every supported dynamic value type.

Raised in this PR review comment (https://github.com/rapidsai/quent/pull/681#discussion_r3966905306).

johanpel:
# Description

- Support recursively nested dynamic attribute lists.
- Add generic `DynamicAttributes::add(key, value)` conversions for supported values.
- Retain explicit `add_null` and `add_attribute` operations while removing redundant type-specific helpers.
- Re-export dynamic attribute value types through generated instrumentation APIs.
- Remove the unused `DynamicValue` to `f64` conversion API and its `thiserror` dependency.
- Update generated bindings, examples, tests, and the instrumentation tutorial.

## Related Issues

Closes #79
Closes #694

## Testing

- `pixi run cargo test -p quent-dynamic-attributes --all-features`
- `pixi run cargo test -p quent-instrumentation-build`
- `pixi run cargo test -p quent-instrumentation --lib`
- `pixi run cargo test -p quent-codegen`
- `pixi run cargo clippy -p quent-dynamic-attributes -p quent-instrumentation-build -p quent-instrumentation -p quent-codegen --all-targets --all-features -- -D warnings`
- `pixi run bash docs/tutorial/build.sh`

## Screenshots

Not applicable.

_Written by Codex._

### DIFF
--- a/crates/dynamic-attributes/src/lib.rs
+++ b/crates/dynamic-attributes/src/lib.rs
@@ -39,6 +39,7 @@ pub enum DynamicList {
     F64(Vec<f64>),
     String(Vec<String>),
     Struct(Vec<DynamicStruct>),
+    List(Vec<DynamicList>),
 }
 
 /// A [`DynamicAttribute`] value.
@@ -225,6 +226,10 @@ impl DynamicAttributes {
         });
     }
 
+    pub fn add_list(&mut self, key: impl Into<String>, value: DynamicList) {
+        self.0.push(DynamicAttribute::list(key, value));
+    }
+
     pub fn into_vec(self) -> Vec<DynamicAttribute> {
         self.0
     }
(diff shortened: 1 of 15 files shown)


crates/dynamic-attributes/src/lib.rs
```diff
@@ -194,35 +261,19 @@ impl DynamicAttributes {
...
+        self.0.push(DynamicAttribute {
+            key: key.into(),
+            value: Some(value.into()),
+        });
     }
 
-    pub fn add_f64(&mut self, key: impl Into<String>, value: f64) {
-        self.0.push(DynamicAttribute::f64(key, value));
+    pub fn add_attribute(&mut self, attr: DynamicAttribute) {
+        self.0.push(attr);
```
mbrobbel: Do we need this?

crates/dynamic-attributes/src/lib.rs
```diff
@@ -194,35 +261,19 @@ impl DynamicAttributes {
...
-    pub fn add_bool(&mut self, key: impl Into<String>, value: bool) {
-        self.0.push(DynamicAttribute {
-            key: key.into(),
-            value: Some(if value {
-                DynamicValue::U8(1)
-            } else {
-                DynamicValue::U8(0)
-            }),
-        });
+    pub fn add_null(&mut self, key: impl Into<String>) {
```
mbrobbel: What about using `()` or some `Null` unit struct with an `Into` impl, so we just have `add`?

crates/dynamic-attributes/src/lib.rs
johanpel: no this can also be removed: https://github.com/rapidsai/quent/pull/698/commits/bcc5c2af793d7789ba88ae42de8651736fd094e3

crates/dynamic-attributes/src/lib.rs
johanpel: Added here: https://github.com/rapidsai/quent/pull/698/commits/bcc5c2af793d7789ba88ae42de8651736fd094e3

This is not the only possible solution, we could alternatively encode an explicit DynamicValue::Null too and make the DynamicAttribute::value field non-optional in order to prevent multiple encodings for null values. I don't really have a preference here myself.

mbrobbel: Yes let's do `Option` then, with https://doc.rust-lang.org/stable/std/option/enum.Option.html#impl-From%3CT%3E-for-Option%3CT%3E it should still be easy to use.

mbrobbel (approved):

Issue #492, opened 2026-07-30
Introduce query engine entity analysis traits


johanpel:
This will help migrate to schema-based events, and ultimately allow the backing storage to be interchangable (e.g. to leverage Narrow etc.)

johanpel:
Solved in #493

PR #699, opened 2026-09-10
refactor(instrumentation)!: Schema-based instrumentation architecture

### LINKED ISSUE #191: Schema-based instrumentation architecture

johanpel:
# Description

Completes the migration from the legacy model and macro stack to schema-generated instrumentation and storage.

The legacy `quent-model`, `quent-model-macros`, `quent-codegen`, `quent-stdlib`, and `quent-query-engine-model` crates have been removed together with their legacy examples, integration tests, and documentation.

## Application-agnostic analyzer crate (`quent-analyzer`)

The analyzer has been refactored to match the modular constraints defined by `quent-ref-target`, `quent-ref-tree`, `quent-fsm`, and `quent-resource`. Previously intertwined entity, FSM, reference-tree, and resource semantics are now separated.

- Entities
  - No longer require an instance name.
  - Expose storage-independent analysis traits, with Rust-native in-memory implementations isolated under
  `entity::native`.
- FSMs
  - No longer require a single exit state; any number of states may be final.
  - Preserve the final transition that closes an FSM.
  - Validate timestamp and sequence-number ordering.
  - Validate transition sequences against the statically defined FSM topology.
  - Isolate Rust-native reconstruction under `fsm::native`.
- Reference trees
  - Provide generic analysis of schema-defined scoped references.
  - Separate entity scope relationships from resource semantics.
- Resources
  - Remove explicit resource groups in favor of free-form scopes expressed through `quent-ref-tree`.
  - No longer require resources to be leaves in a reference tree.
  - No longer prescribe a fixed FSM topology; both plain entities and FSMs may be resources.
  - Retain derived resource trees for UI hierarchy and aggregated resource timelines.
- Contexts
  - Add application-agnostic indexing from analysis targets to the runtime contexts contributing telemetry.
- UI contracts
  - Move runtime FSM declarations and related UI-specific representations into `quent-ui`.

## Query-engine model

- Remove `quent-query-engine-model` and the legacy concrete `InMemoryQueryEngineModel`.
- Retain schema-independent query-engine semantics as traits in `quent-query-engine-analyzer`, including entity access, traversal, plan relationships, resource timelines, and UI conversion.
- Require applications to implement those traits for their own schema-generated event types instead of depending on a shared concrete event model.
- Generalize server-side context discovery around `ContextInventory` rather than query-engine-specific worker indexing.
- Remove simulator dependencies from the query-engine server.
- Move query-engine UI bindings out of the simulator and into the query-engine domain.

## `quent-open`

- Select generated viewer compatibility from the Quent revision recorded in `model.qmi`.
- Use Git ancestry boundaries for:
  - the `quent-exporter` to `quent-io` rename;
  - NVTX server routes;
  - query-engine-specific indexing versus application-agnostic context inventories.
- Preserve support for artifacts produced from older and divergent branches without detecting compatibility through failed builds.
- Use the Git CLI for fetching pinned dependencies so existing SSH configuration and credentials are respected.
- Add a compatibility test and repository guidance for preserving historical wrapper contracts across future breaking changes.

## Simulator

- Define the simulator application model in `model.yaml`.
- Generate instrumentation and stored event types through `quent-instrumentation-build` and `quent-store-build`.
- Replace legacy model macros and standard-library resource types with schema-defined entities, FSMs, resources, usages, and scoped references.
- Implement the query-engine analyzer traits directly for simulator entities.
- Keep the repetitive schema adapters in explicit `boilerplate` modules pending generated analyzer implementations tracked by #288.
- Derive resource hierarchy from schema reference relationships while preserving existing binned resource-timeline behavior.
- Model tasks, executors, threads, memories, channels, GPUs, storage, and networking as independent schema entities with their actual resource and FSM constraints.
- Move the fixed query-engine integration tests under the simulator.

## Related issues

Relates to #288.
Closes #191 #491 #492

WIP Exploration of upstream changes needed in SiriusDB: https://github.com/sirius-db/sirius/pull/1795

## Testing

- `pixi run cargo test -p quent-open --all-features`
- `pixi run cargo clippy -p quent-open --all-features --all-targets -- -D warnings`

_Written by Codex._


crates/open/src/compatibility.rs
```diff
@@ -0,0 +1,159 @@
...
+/// Cargo package used before [`IO_PACKAGE`] was introduced.
+pub(crate) const LEGACY_IO_PACKAGE: &str = "quent-exporter";
+/// Cargo package that provides the optional NVTX HTTP routes.
+pub(crate) const NVTX_SERVER_PACKAGE: &str = "nvtx-server";
+
+/// Boundary whose descendants provide the `quent-io` package.
+const IO_PACKAGE_BOUNDARY: &str = "aa1e9b1b394f5f978215b69cd5c526291c4b4723";
+/// Boundary whose descendants provide NVTX routes and the extensible analyzer router.
+const NVTX_ROUTES_BOUNDARY: &str = "f40e69c2d4405c765c6270221e2a58e58ef704a6";
+/// Last `upstream/main` revision before context-inventory indexing was introduced.
```
mbrobbel: Maybe add links to the commits here for convenience.

crates/open/src/compatibility.rs
```diff
@@ -0,0 +1,159 @@
...
+///
+/// PR #699 is required to be the next change merged after this revision. Only strict
+/// descendants use context-inventory indexing; this revision, its ancestors, and
+/// revisions on older branches use query-engine indexing.
+const CONTEXT_INVENTORY_PREDECESSOR: &str = "5e6818e89ccde0a41dc08a12819938a093969665";
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum NvtxRoutes {
+    Enabled,
+    Disabled,
+}
```
mbrobbel: `bool`?

crates/open/src/revision.rs
```diff
@@ -0,0 +1,223 @@
...
+
+/// A pinned revision fetched into a local bare repository for ancestry checks.
+pub struct PinnedRevision<'a> {
+    repository: &'a Path,
+}
+
+impl<'a> PinnedRevision<'a> {
+    /// Fetch `pin` into `repository` without checking out its file contents.
+    pub async fn fetch(repository: &'a Path, pin: &GitPin) -> Result<Self> {
+        std::fs::create_dir_all(repository)?;
```
mbrobbel: nit: tokio fs

AGENTS.md
```diff
@@ -1,28 +1,24 @@
...
+crates, package names, features, or APIs used by the generated wrapper change.
 
-- [#638: schema-based C++ and Python generators][pr-638]
-- [#699: schema-based instrumentation architecture][pr-699]
+Before making such a breaking change:
 
-Before starting work that affects the areas below, check whether both pull
-requests have merged. Until both have merged, do not make significant changes
-to:
+- Add an immutable full commit SHA from `upstream/main` immediately before the
```
mbrobbel, line 11: Maybe add a note for what this looks like when we have releases?

mbrobbel (commented):
Nice

crates/open/src/compatibility.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/fdf836950eab4df71189780b60ed1a5f34013c23

crates/open/src/compatibility.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/fdf836950eab4df71189780b60ed1a5f34013c23

crates/open/src/revision.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/fdf836950eab4df71189780b60ed1a5f34013c23

AGENTS.md
johanpel: https://github.com/rapidsai/quent/pull/699/commits/fdf836950eab4df71189780b60ed1a5f34013c23

crates/analyzer/src/entity/mod.rs
```diff
@@ -1,83 +1,32 @@
...
-            .field("earliest_timestamp", &self.earliest_timestamp)
-            .field("latest_timestamp", &self.latest_timestamp)
-            .finish_non_exhaustive()
-    }
+/// Trait for analysis-time types that represent an entity.
+pub trait Entity {
+    /// Return the universally unique identifier.
+    fn id(&self) -> Uuid;
+    /// Return the type name.
+    fn type_name(&self) -> &str;
```
mbrobbel, line 18: Nit; this could be an associated `const`?

crates/analyzer/src/entity/native.rs
```diff
@@ -0,0 +1,159 @@
...
+    ///
+    /// Returns [`AnalyzerError::Validation`] if `id` is nil.
+    pub fn try_from_event(event: Event<A::Event>) -> AnalyzerResult<Self> {
+        if event.id.is_nil() {
+            Err(AnalyzerError::Validation(
+                "entity id cannot be nil".to_owned(),
+            ))
+        } else {
+            let mut accumulator = A::default();
+            accumulator.push(event.data);
```
mbrobbel, line 56: This would benefit from preallocation. Maybe we can use adopt `FromIterator` or `Extend`? Or we add a prealloc (`with_capacity`) method in addition to `Default` to `Entity`?

crates/analyzer/src/entity/mod.rs
```diff
@@ -1,83 +1,32 @@
...
-impl<M: EntityData> std::fmt::Debug for EntityEvents<M> {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("EntityEvents")
-            .field("id", &self.id)
-            .field("earliest_timestamp", &self.earliest_timestamp)
-            .field("latest_timestamp", &self.latest_timestamp)
-            .finish_non_exhaustive()
-    }
+/// Trait for analysis-time types that represent an entity.
+pub trait Entity {
```
mbrobbel, line 14: What is the relation between this and `quent_event::Entity`?

crates/analyzer/src/entity/native.rs
```diff
@@ -0,0 +1,159 @@
...
+//! Analysis-time entities in Rust-native in-memory storage.
+
+use quent_events::Event;
+use quent_time::TimeUnixNanoSec;
+use uuid::Uuid;
+
+use crate::{AnalyzerError, AnalyzerResult};
+
+/// Accumulates events of its associated entity event type.
+pub trait EntityEventAccumulator: quent_events::Entity + Default {
```
mbrobbel: See my question above, why is this not using the analyzer `Entity` trait as bound?

crates/analyzer/src/entity/native.rs
```diff
@@ -0,0 +1,159 @@
...
+        Ok(())
+    }
+
+    /// Returns the event accumulator.
+    pub fn accumulator(&self) -> &A {
+        &self.accumulator
+    }
+}
+
+impl<A: EntityEventAccumulator> crate::entity::Entity for AnalyzedEntity<A> {
```
mbrobbel, line 91: Ok, I see this now. I wonder if we need the `earliest` + `latest` fields materialized, or if we can just scan the `accumulator`?

I wonder what this adds or why we need it?

crates/analyzer/src/fsm/mod.rs
```diff
@@ -130,55 +105,16 @@ where
         self.fsm.transition(self.index).unwrap().name()
     }
 
+    /// Returns the transition that closes this state.
+    ///
+    /// For the last state of a complete FSM, this is the final transition.
+    pub fn next_transition(&self) -> &T {
+        self.fsm.transition(self.index + 1).unwrap()
```
mbrobbel, line 203: I assume this panics when called on the `FsmStateRef` returned by `last`?

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+use crate::{
+    AnalyzerError, AnalyzerResult, Entity,
+    fsm::{Fsm, FsmUsages, Transition},
+    resource::{AnalyzedUsage, CapacityValue, Usage, Using},
+};
+
+/// Trait for application-specific payloads of FSM transition events.
+pub trait TransitionEvent: EntityEvent {
+    /// Return the name of the state transitioned into.
+    fn name(&self) -> &'static str;
```
mbrobbel, line 26: Nit: associated const?

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+
+    /// Returns the per-FSM wrapping sequence number assigned in transition order.
+    fn sequence(&self) -> u16;
+
+    /// Returns whether this transition ends the FSM's dynamic lifetime.
+    fn is_final(&self) -> bool;
```
mbrobbel, line 35: Could be a `const` too I think?

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+    }
+
+    fn is_final(&self) -> bool {
+        self.data.is_final()
+    }
+}
+
+impl<T> AnalyzedTransition<T> {
+    /// Returns resources held for the state ending at the next transition.
+    pub fn usages(&self) -> &[AnalyzedUsage] {
```
mbrobbel, line 103: This should be `Transition::usages` right?

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+        &self.usages
+    }
+}
+
+pub struct UsageWithSpan<'a> {
```
mbrobbel: Maybe add a doc comment to this `pub` item?

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+/// wrapping) sequence number.
+pub struct AnalyzedFsmBuilder<T> {
+    id: Uuid,
+    transitions: OrderedCollector<AnalyzedTransition<T>>,
+    duplicate_order_key: Option<(TimeUnixNanoSec, u16)>,
+}
+
+impl<T: TransitionEvent> AnalyzedFsmBuilder<T> {
+    pub fn try_new(id: Uuid) -> AnalyzerResult<Self> {
+        if id.is_nil() {
```
mbrobbel, line 142: Maybe handle this in the type instead of relying on validation checks everywhere?

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// collected.
+    pub fn try_build(self) -> AnalyzerResult<AnalyzedFsm<T>> {
+        if let Some((timestamp, sequence)) = self.duplicate_order_key {
+            return Err(AnalyzerError::Validation(format!(
+                "fsm '{}' (id={}) has multiple transitions with timestamp {timestamp} and sequence {sequence}",
+                T::NAME,
+                self.id,
+            )));
+        }
+        let transitions: SmallVec<[AnalyzedTransition<T>; 4]> =
```
mbrobbel: Maybe use a `const` for `4` with a note about the rationale for that value?

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+/// An FSM reconstructed from application-specific transition data.
+///
+/// This type can be constructed through [`AnalyzedFsmBuilder`].
+///
+/// Application-specific data remains available through [`Self::transitions`].
+pub struct AnalyzedFsm<T> {
+    id: Uuid,
+    transitions: SmallVec<[AnalyzedTransition<T>; 4]>,
```
mbrobbel: Would be used here too then.

crates/analyzer/src/fsm/native.rs
```diff
@@ -0,0 +1,480 @@
...
+}
+
+impl<T: TransitionEvent> std::fmt::Debug for AnalyzedFsm<T> {
```
mbrobbel, line 246: This can be derived?

crates/analyzer/src/resource/mod.rs
```diff
@@ -12,35 +12,18 @@ use uuid::Uuid;
...
-            contains_resource_types,
-        }
-    }
+/// Analysis-time representation of one resource usage.
+#[derive(Debug)]
+pub struct AnalyzedUsage {
+    /// Resource used by the entity.
+    pub resource_id: Uuid,
+    /// Capacity values reserved from the resource.
+    pub capacities: SmallVec<[CapacityValue; 3]>,
```
mbrobbel: Same question here.

crates/analyzer/src/resource/tree.rs
```diff
@@ -1,188 +1,243 @@
...
-        let resource_children = resources
-            .resource_group_child_resources(root_group_id)?
-            .map(|child_resource| Ok(ResourceTreeNode::Resource(child_resource)));
-        Ok(ResourceTreeNode::ResourceGroup(
-            root_group_id,
-            group_children
-                .chain(resource_children)
-                .collect::<AnalyzerResult<_>>()?,
-        ))
+        let mut resource_ids = HashSet::default();
```
mbrobbel: Nit: preallocate

crates/analyzer/src/context.rs
```diff
@@ -0,0 +1,158 @@
...
+/// Maps each analysis target to all runtime contexts contributing telemetry.
+///
+/// Multiple independent targets are supported, such as two implementations
+/// running the same benchmark workload whose results will be compared.
+#[derive(Debug, Default)]
+pub struct ContextIndex {
+    contexts_by_analysis_target: FxHashMap<Uuid, BTreeSet<ContextId>>,
+}
+
+impl ContextIndex {
```
mbrobbel, line 49: No by-id lookup method?

experimental/vibe/simulator/application/src/lib.rs
```diff
@@ -324,86 +324,59 @@ struct Plan<T>
...
-                                    operator_id: Ref::new(op.id),
-                                    instance_name: edge.weight().source.name.to_string(),
-                                },
-                            )
-                        }),
+                worker,
+            )
+            .unwrap();
+
+        for node_idx in self.dag.node_indices().collect::<Vec<_>>() {
```
mbrobbel: Do you need the `collect` here?

experimental/vibe/simulator/application/src/lib.rs
```diff
@@ -324,86 +324,59 @@ struct Plan<T>
...
+                    op.name(),
+                    Default::default(),
                 )
-            {
-                port_obs.create(id).declaration(event)
-            }
+                .unwrap();
+        }
+
+        for edge_idx in self.dag.edge_indices().collect::<Vec<_>>() {
```
mbrobbel: Same question here

mbrobbel (commented):
Mostly questions and nits

crates/analyzer/src/entity/mod.rs
johanpel: The `quent_event::Entity` is just a trait implemented by marker types used in the instrumentation API. It mostly surfaces in instrumentation but calling it "Entity" there I think is clear to instrumentation API users, since they need to use the marker type named exactly like their schema entity type wherever this bound appears.

The `quent_analyzer::Entity` trait represents an schema-agnostic interface to analysis-time entities. This is a type which has (had) visibility into all the events in the available data for one entity instance, irrespective of how it is stored.

crates/analyzer/src/entity/native.rs
johanpel: This allows us to compute these values once. This is useful because downstream analysis consumers make continuous requests against this cached analyzed data, so at this level we won't have to recompute by iterating across all events again. In this case it helps with the entity list route for example which can have a varying time window filter.

crates/analyzer/src/entity/native.rs
johanpel: Yes this is a good question. I should have removed this bound and made the event type associated.

crates/analyzer/src/entity/mod.rs
johanpel: Technically yes, but without adding more code generation, there are still a few places where this needs to be used as a trait object. It's something I might spend some time on later.

crates/analyzer/src/entity/native.rs
johanpel: I think we could add this but it would only become useful when we have some type of storage that would know this capacity. We don't have that today.

crates/analyzer/src/fsm/native.rs
johanpel: We can't since one enum type represents all transition payloads.

crates/analyzer/src/fsm/native.rs
johanpel: Same as above.

crates/analyzer/src/fsm/native.rs
johanpel: Conceptually transitions don't necessarily have usages, which depends on whether the FSM is using a resource at all. This should be separated out a bit better by an additional trait for transitions that can have usages such that codegen can generate the impls for the relevant FSM types later. This trait should live in the resource module. Since this is a bit involved I will add an issue to address this in a follow up.

johanpel: https://github.com/rapidsai/quent/issues/741

mbrobbel: Oh, I was looking at `TransitionEvent` 🤦

crates/analyzer/src/entity/mod.rs
johanpel: https://github.com/rapidsai/quent/issues/742

crates/analyzer/src/fsm/native.rs
johanpel: I like the idea but since this would have to be propagated through literally every place we use and produce entity UUIDs, I'm gonna park this here for now: https://github.com/rapidsai/quent/issues/743

crates/analyzer/src/fsm/native.rs
johanpel: Deriving this would add a Debug bound to the transition payload, which is opt-in right now through a build script

crates/analyzer/src/context.rs
johanpel: That's `contexts_of_analysis_target`

crates/analyzer/src/entity/native.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/fb60ef79b836d3bb93e55ddd2d6821304773407c

crates/analyzer/src/fsm/native.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/84915468ef0b7e5adf1ccb606392192ef788e8a1

doesn't need pub

crates/analyzer/src/fsm/mod.rs
johanpel: You can only get a state ref if there are at least two transitions so it will never panic. However there was another bug related to this fixed here: https://github.com/rapidsai/quent/pull/699/commits/961cfa8bd3acd7cf1d6649bc71b1b3668687636f

Also props to 🐰

crates/analyzer/src/fsm/native.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/010102776700f342750ded1b24de0648fffaaab3

crates/analyzer/src/fsm/native.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/010102776700f342750ded1b24de0648fffaaab3

crates/analyzer/src/resource/mod.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/010102776700f342750ded1b24de0648fffaaab3

crates/analyzer/src/resource/tree.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/58baaacbcca883f76d3706965bdc38489a4e143d

experimental/vibe/simulator/application/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/837178cee88958be7c2706fd3adf6cab3685b958

experimental/vibe/simulator/application/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/837178cee88958be7c2706fd3adf6cab3685b958

mbrobbel (approved):

experimental/vibe/simulator/application/src/lib.rs
```diff
@@ -1260,69 +1414,50 @@ impl Worker {
...
-            op_obs.create(operator.id).statistics(operator::Statistics {
-                custom_attributes: attributes.into(),
-            });
+            operator.handle.statistics(attributes.into()).unwrap();
         }
 
-        let port_obs = context.port_observer();
-        for edge in physical_plan.dag.edge_references() {
-            let source = &physical_plan.dag[edge.source()];
+        for edge_idx in physical_plan.dag.edge_indices().collect::<Vec<_>>() {
```
9prady9: Nit: iterate over `edge_indices()` directly; it owns its index range, so `collect::<Vec<_>>()` adds an unnecessary allocation.

experimental/vibe/simulator/analyzer/src/model.rs
```diff
@@ -256,257 +553,282 @@ impl SimulatorModelBuilder {
...
+                    self.engine = Some(Engine::try_from_event(event)?);
+                    Ok(())
+                }
             }
-            SimulatorEvent::Port(e) => {
-                self.query_engine
-                    .try_push(Event::new(id, timestamp, QueryEngineEvent::Port(e)))
+            SimulatorEvent::Worker(event) => {
+                let event = Event::new(id, timestamp, event);
+                if let Some(worker) = self.workers.get_mut(&id) {
```
9prady9: Nit: use `entry(id)` here, as in the Task branch, to avoid a second lookup when inserting a new worker.

crates/analyzer/src/entity/native.rs
```diff
@@ -0,0 +1,160 @@
...
+pub struct AnalyzedEntity<A: EntityEventAccumulator> {
+    id: Uuid,
+    earliest_timestamp: TimeUnixNanoSec,
+    latest_timestamp: TimeUnixNanoSec,
+    accumulator: A,
+}
+
+impl<A: EntityEventAccumulator> std::fmt::Debug for AnalyzedEntity<A> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EntityAnalysis")
```
9prady9: Nit: change `debug_struct("EntityAnalysis")` to `debug_struct("AnalyzedEntity")` so debug output matches the type name.

9prady9 (approved):
lg, great work! it is finally going in :)

Left minor nits to follow up later.

experimental/vibe/simulator/application/src/lib.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/9852c5ff4e4f7162e03d36db22b0df5692bdd91b

experimental/vibe/simulator/analyzer/src/model.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/9852c5ff4e4f7162e03d36db22b0df5692bdd91b

crates/analyzer/src/entity/native.rs
johanpel: https://github.com/rapidsai/quent/pull/699/commits/9852c5ff4e4f7162e03d36db22b0df5692bdd91b

PR #701, opened 2026-09-10
feat(ui): display worker info for stats panel operators

### LINKED ISSUE #700: feat: add worker info to operators in stats panel
### Summary

Currently you are unable to distinguish the operators in the stats panel from one another; similar to what we did for the operator drop-down in the entities table, we should add that information to the stats panel.

cmatzenbach:
# Description

Adds worker information to each operator section in `DAGNodeInfoPanel`, so that the various operators can be distinguished from one another.  Additionally, extracts worker and worker+plan label buildings functions as helpers that be can used throughout the app rather than localized in components. Also updates unit tests to test for new conditions.

## Related Issues
Closes: #700 

## Testing

1. Select an operator in DAG, check stats panel - if available, worker info should now be displayed in-line with the operator name on the trigger of the collapsible panel

## Screenshots

https://github.com/user-attachments/assets/b4092fc5-aa2a-4940-b791-0e6900d87a1c

### DIFF
--- a/ui/packages/@quent/components/src/node-info/OperatorAccordion.tsx
+++ b/ui/packages/@quent/components/src/node-info/OperatorAccordion.tsx
@@ -36,13 +36,21 @@ export const OperatorAccordion = ({
         className="group flex w-full min-w-0 cursor-pointer items-center gap-2 rounded-sm px-1.5 py-1 my-1 text-left hover:bg-muted/50"
         aria-label={`Toggle ${operator.label} details`}
       >
-        <DataText className="min-w-0 truncate text-xs font-medium" title={operator.label}>
+        <DataText className="min-w-0 flex-1 truncate text-xs font-medium" title={operator.label}>
           {operator.label}
         </DataText>
-        <DataText className="shrink-0 rounded bg-muted px-1.5 py-0.5 text-xs capitalize text-muted-foreground">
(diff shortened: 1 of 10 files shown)


ui/packages/@quent/components/src/node-info/OperatorAccordion.tsx
```diff
@@ -34,15 +34,27 @@ export const OperatorAccordion = ({
...
-        <DataText className="min-w-0 truncate text-xs font-medium" title={operator.label}>
+        <DataText className="min-w-0 flex-1 truncate text-xs font-medium" title={operator.label}>
           {operator.label}
         </DataText>
-        <DataText className="shrink-0 rounded bg-muted px-1.5 py-0.5 text-xs capitalize text-muted-foreground">
+        <DataText className="shrink-0 rounded bg-muted px-1.5 pt-0.5 pb-px text-xs leading-none capitalize text-muted-foreground">
           {operator.operationType}
         </DataText>
-        <ChevronDown className="ml-auto h-3 w-3 shrink-0 text-muted-foreground transition-transform group-data-[state=closed]:-rotate-90" />
+        {operator.workerLabel && (
```
johallar, line 49: Would be good to add a max width so these don't squish the operator name:
Screenshot 2026-09-15 at 10 11 35 AM

johallar (approved):
👍 one small UX improvement

PR #702, opened 2026-09-11
feat: Introduce engine-defined grouping for information in operator detail display

wence-:
# Description

When attaching information to operators, I want to be able to provide some structure (e.g. grouping by "batch info", "global info", etc..). Quent can't do this, because it doesn't know anything about my data.

I also want to be able to attach data to the operator ports. For example, for a hash-join, which input port was chosen as the build side (if I do this at runtime rather than plan time).

To that end, introduce two concepts.

An `InformationGroup` is a collection of "grouped" information, with a `String` header and ordered `DynamicAttributes` items that are the key-value pairs of information.

To attach information to ports, I introduce a `PortRelation` mapping a `Ref<Port>` to a `String` role.

Finally, because I might not have this PortRelation until runtime, Operators now gain an `OperatorObservation` that is a timestamped bundle of information (I think this might need a few changes).

The rest is then threading through typescript ui and codegen bindings.

I also make the storage a Vec rather than a HashMap so that the query engine can define the presentation order, rather than having a random order every time.

## Testing

I did this testing:
- pixi run cargo fmt --all
- pixi run cargo clippy --workspace --all-targets --all-features --locked -- -D warnings
- pixi run cargo test --workspace --all-features --locked --all-targets

UI changes:
- cd ui
- pnpm ci:check


wence-:
After discussion with @johanpel, I don't need the InformationGroup stuff. Though I think the PortRelations are still useful. Will rework on top of the ongoing schema refactor.

PR #711, opened 2026-09-14
chore(ui): fix eslint warnings

cmatzenbach:
# Description

- **Real fixes** (`DAGNodeInfoPanel`, `FsmCapacityChart`, `useEntityTable`, `tree-view.tsx`, `tree-table.tsx`): replaced `useEffect` + `setState` patterns that reset/sync derived UI state with render-time adjustment, removing a one-frame stale-render flicker. `tree-table.tsx` had a redundant state+effect copy of a prop that was removed entirely.
-  **Suppressed with inline comments** (`Timeline.tsx`, `ResourceTimeline.tsx`, `LongEntitiesRow.tsx`, `DeepLinkBoundary.tsx`, `CopyLinkButton.tsx`, and a handful of "latest ref" mirror hooks): legitimate, intentional patterns (retain-last-value refs, DOM reads on mount, one-time hydration flags) that the new rules can't distinguish from bugs.
-  **Disabled at the rule level** (`eslint.config.js`): `react-refresh/only-export-components` (dev-only Fast Refresh ergonomics, no runtime impact) and `react-hooks/incompatible-library` (only relevant if we adopt React Compiler, which we don't). Also cleaned up now-unused per-file `eslint-disable` comments this made obselete.             

## Testing

1. `pnpm lint` produces zero warnings/errors
2. `pnpm typecheck` passes cleanly
3. All existing unit test suites pass


ui/packages/@quent/components/src/dag/DAGNodeInfoPanel.tsx
```diff
@@ -54,22 +54,33 @@ export const DAGNodeInfoPanel = ({
...
     setIsExpanded(hasSelection);
     if (!hasSelection) {
       setActiveTab('stats');
     }
-  }, [hasSelection]);
+  }
 
-  useEffect(() => {
+  // Collapse per-operator sections as soon as the selected operators change
+  const [prevSelectedOperatorIdsKey, setPrevSelectedOperatorIdsKey] =
```
johallar, line 68: I think we should choose another pattern for fixing this lint rule, react docs don't even like their own solution ☹️ :
https://react.dev/reference/react/useState#storing-information-from-previous-renders

> This pattern can be hard to understand and is usually best avoided. However, it’s better than updating state in an effect.

I think creating an "keyed" inner component here is a better solution if we can make it work for these and other fixes for the same rule:
https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes

cmatzenbach: So I went through and audited these, and only one of the three use cases fit this suggested pattern (which I agree, is better). `FsmCapacityChart` fit this perfectly, just needed to add a `key` prop to its call site in `EntityDetailPanel`. This pattern doesn't apply to `useEntityTable`, since it's a hook and not a component (thus no subtree to remount). I took a long look at `DAGNodeInfoPanel`, but it doesn't fit this pattern - we could potentially make 2/3 triggers keyed sub-components (`isExpanded/activeTab` could live in a child keyed by `String(hasSelection)` and `closedOperatorIds` could have a child keyed by `selectedOperatorIdsKey`), but the third trigger can't be modeled as a key reset (which is the data-flow tab auto-switch - this bit: `if (isPlaying && isExpanded && showDataFlowTab) setActiveTab('data-flow')`). This isn't a "reset state when identity changes" case, it's more of  a "force a value when an external condition becomes true while playing" case. There's no prop who's identity changing represents that.

So while we could extract those two, the pattern would still exist, and we'd have to thread stuff like `statsContent`, `dataFlowContent`, tab-switch logic, etc across component boundaries via props...this seems a lot harder to read than three self-contained, commented `if (x !== prev)` blocks, while still not eliminating the pattern completely.

So that's my thoughts, I went ahead and committed the change that made sense. Lmk what you think.

johallar: Fair enough, I think it's a fair trade since i'd say new inner components would just be "structural" but neither solution is particularly nice. Makes me wonder what problem we're actually solving here and if we should be complicating our code to fix it (now and in future changes). I would almost prefer an eslint ignore if there's no perceivable problem (I think it could cause a flicker for a frame?)

johallar: Structural meaning they're not being wrapped in a component to be re-usable, but to serve the "keyed" purpose, so the interface/props becomes much less important IMO.

Also agree about `if (isPlaying && isExpanded && showDataFlowTab) setActiveTab('data-flow')`, can't be modeled this way, seems like it _should_ be implemented a side effect `useEffect`

johallar (approved):
Looks good, thanks!

Issue #485, opened 2026-07-29
feat: bulk entities endpoint


johallar:
### Summary

bulk version of the /entities endpoint. When loading many timelines at once, or toggling the entities chart on for many charts at once we'll want to be able to make one request instead of n. 

This should be structured the same as the /timeline bulk endpoint

Can sketch out a TS API if helpful 

### Request type

New feature

### Affected area

Rust collector/analyzer/server

PR #731, opened 2026-09-15
fix(ui): dim unrelated entities on operator filter

### LINKED ISSUE #715: Selecting operators should not completely hide unrelated entities in the timeline
https://github.com/user-attachments/assets/ab6e011e-a491-4487-a6c6-93bc2b608270

When selecting a DAG vertex, in the resource timeline, there is a mismatch in behavior between how the resource utilization graph representation changes and how the entities timeline changes.

For the utilization graph, unrelated parts are "grayed out", whereas entities are completely hidden. This is a bit confusing as you suddenly lose information of what stuff contributed to grayed out portions of resource utilization, rather than it being deemphasized.

I would prefer the unrelated entities are similarly grayed out instead of disappearing entirely. Here it would perhaps be nice if related concurrently active entities contributing to a resource usage are ordered to appear on the topmost swimlanes.

cmatzenbach:
# Description

Treat entities the same way we treat resources - when applying an operator filter, dim those that are unrelated to the selected operator rather than removing them completely. 

@johanpel One small backend change was needed for this - the backend used to filter entities by operator itself and only return  the matches. This was fixed by having the backend stop filtering and instead just expose each entity's `operator_id` (a value it was already computing internally but discarding), so the UI can fetch everything and dim the unrelated ones itself. 

## Related Issues

Closes: #715 

## Testing

1. Select a physical plan and expand the associated worker tree
2. Select an operator, ensure all unrelated entities are dimmed
3. Select multiple operators within that plan, ensure more entities become active
4. Select all operators, ensure all entities within the tree are active
5. Select operators in logical plan, ensure entity highlight/dimming works for those as well
6. Select active entity to open detail pane, ensure all other operators are dimmed
7. Close detail pane, ensure all entities previously highlighted by the operator selection are now highlighted

## Screenshots

https://github.com/user-attachments/assets/86f6ef65-7164-4189-a26b-85a2f00359c9


johallar:
From an API point of view I think we want to keep the operator ids filter, would this also affect the entities table filter?

The way resource timelines do this is with two bulk request entries (one for operator filter and one without). If https://github.com/rapidsai/quent/issues/485 is viable maybe we can do the same here and maintain operator ids as filters.

johallar:
OK, I misunderstood, for posterity: entities still filter by operator id, but the UI will just not be passing filters to render entity swimlanes.

ui/packages/@quent/components/src/long-entities/LongEntitiesGantt.tsx
```diff
@@ -128,7 +129,7 @@ export function LongEntitiesGantt({
 
       const hasSelection = selectedEntityId != null;
       const isSelected = hasSelection && entry.entityId === selectedEntityId;
-      const opacity = hasSelection && !isSelected ? 0.3 : 1;
+      const opacity = entry.isDimmed ? DIMMED_OPACITY : hasSelection && !isSelected ? 0.3 : 1;
```
johallar: Why not use the same DIMMED_OPACITY for the non selected items? 

```suggestion
      const opacity = (hasSelection && !isSelected) || entry.isDimmed ? DIMMED_OPACITY : 1;
```

johallar (approved):
🚀

crates/ui/src/lib.rs
```diff
@@ -243,6 +243,8 @@ pub struct FiniteStateMachine {
     pub instance_name: String,
     /// The transitions of this FSM.
     pub transitions: Vec<FsmTransition>,
+    /// The operator that caused this FSM's activity, if known.
+    pub operator_id: Option<Uuid>,
```
johanpel: This is a simple fix but it breaks some of the philosophy about core Quent being application-agnostic. It introduces an application-specific field to an application-agnostic type.

One alternative approach would be to wrap FSM in a new "QueryEngineFSM" type or something in the query engine domain model which adds this field. This is a bit more involved though.

I wonder if we could also just display the filtered FSMs on top of the unfiltered ones that are grayed out as long as we're in operator selection mode?

cmatzenbach:
Note: Holding off until the schema migration is complete. @johanpel will have another look once that is finished

johanpel:
@cmatzenbach I merged in the latest main and added: https://github.com/rapidsai/quent/pull/731/commits/062260c0c0c9c1c15c202c9967f84bd4f3aaf42c with my proposal.

Not sure if the high amount of ui lint errors are expected and will be fixed in https://github.com/rapidsai/quent/pull/711 ?

PR #732, opened 2026-09-15
feat: quent open work with local UI dev server

johallar:
# Description


johallar:
This works without this small re-org, although this allows for a .env setup. Might just close tho

johallar:
/

PR #734, opened 2026-09-15
fix: bump zod to get .validate speed boost

johallar:
# Description
new .validate function apparently "35x faster", so we'll bump and go with that where we can.

## Related Issues
None

## Testing
All deep linking functionality works as before, no functional changes


cmatzenbach (approved):
Deep links still work as expected, looks good

PR #736, opened 2026-09-16
feat(nvtx): group presentation domains by name

9prady9:
# Description

Group explicitly named, non-default NVTX domains by exact, case-sensitive name at the presentation boundary. This turns duplicate raw domains such as `CCCL` handles into one logical catalog and timeline domain without changing capture or analyzer semantics.

Raw source identity remains available through `source_domain_id` / `source_domain_ids` and continues to namespace registered strings, categories, reconstructed stacks, and lanes. The UI keeps raw-source lanes separate under the logical domain, exposes source handles in diagnostics, and reports category-name conflicts deterministically rather than silently overwriting them.

The server round-trip and generated TypeScript contract are covered alongside the Rust presentation model and frontend behavior.

## Related Issues

N/A

## Testing

- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p nvtx-analyzer -p nvtx-ui -p nvtx-server` (57 passed)
- [x] `cargo clippy -p nvtx-analyzer -p nvtx-ui -p nvtx-server --all-targets --locked -- -D warnings`
- [x] TypeScript binding generation
- [x] UI TypeScript typecheck
- [x] UI ESLint and Prettier checks for changed files
- [x] Full UI Vitest suite (71 files, 902 tests passed)
- [x] `git diff --check`
- [ ] Replay the Sirius SF50 TPC-H Q1 capture from the local Sirius worktree and verify one visible `CCCL` group contains events from raw handles `5` and `172`, with source handles retained in item diagnostics.


9prady9:
image

johanpel (approved):
The Rust changes look good to me w.r.t. the goal of the PR.

I wonder about is whether merging these domains as far as presentation in the UI is the best approach.

Suppose an application uses two orthogonal libraries that coincidentally don't namespace their domains very well and they happen to create domains with a duplicate name. The result is that in the UI the ranges are now merged into seemingly one domain while they are not logically related.

I'm not sure if there's any other metadata besides the domain id that we could visually add to provide a better distinction? If there is, we could consider continuing to display these domains separately but add some of that additional metadata instead. How does NSight Systems deal with this?

That said I'm approving this so you can merge with your best judgement about the above. I think it is more likely that name clashes occur due to the scenario you encountered in Sirius with two different versions of the same library creating domains.

9prady9:
> The Rust changes look good to me w.r.t. the goal of the PR.
> 
> I wonder about is whether merging these domains as far as presentation in the UI is the best approach.
> 
> Suppose an application uses two orthogonal libraries that coincidentally don't namespace their domains very well and they happen to create domains with a duplicate name. The result is that in the UI the ranges are now merged into seemingly one domain while they are not logically related.
> 
> I'm not sure if there's any other metadata besides the domain id that we could visually add to provide a better distinction? If there is, we could consider continuing to display these domains separately but add some of that additional metadata instead. How does NSight Systems deal with this?
> 
> That said I'm approving this so you can merge with your best judgement about the above. I think it is more likely that name clashes occur due to the scenario you encountered in Sirius with two different versions of the same library creating domains.

Yes incorrect or accidental identical name will group them in wrong swim-lane but that is all - it won't remove or corrupt any actual data.

I’m inclined to keep the grouping and validate it against the Sirius capture. Since the raw identity is retained, we could add a `split by source` view later if real captures show that accidental collisions are a practical problem.

9prady9:
>How does NSight Systems deal with this?

Forgot to clarify that, it does name based grouping in the view AFAIU

Issue #618, opened 2026-08-25
Phase 3a: Use single exporter, importer and analyzer for NVTX or future event types


9prady9:
This should alleviate the burden of changing or maintaining `quent-open` compatibility while also letting us quickly scale and extend for new event types.

## Scope

- Represent NVTX and future auxiliary event types through model/schema metadata. The exact schema entry will be designed as part of this issue.
- Route schema-described event streams through the common exporter, importer, and analyzer paths instead of maintaining dedicated NVTX implementations.
- Make the shared path available consistently to generated Rust, CXX, and Python instrumentation contexts.
- Keep source-specific attachment concerns, such as the process-global NVTX injection hook, separate from ordinary event transport and analysis.
- Replace the temporary CXX-only NVTX codegen used to unblock `sirius-db/sirius#1439` once the shared schema-driven path is available.

## Done when

- Adding a schema-described event type does not require bespoke exporter, importer, or analyzer plumbing.
- NVTX events use the shared path without regressing current capture, reconstruction, server, or UI behavior.
- Generated Rust, CXX, and Python contexts can expose the same schema-described event streams.
- `quent-open` compatibility does not require special handling for each additional event type.

9prady9:
In the initial mock, I had a separate code path for it for enable faster POC capture. We have to move to a more composable approach.

9prady9:
# Design decision for #618

Proceed with Johan's trait-based NVTX integration.

Three approaches considered:

1. Shared pipeline with a reusable NVTX component and generated event
   conversions.
2. Schema-defined reconstruction rules executed by a generic engine.
3. **Selected:** schema-expanded NVTX events, generated field-access traits, and
   reusable Rust NVTX reconstruction; analysis policies stay out of the YAML
   DSL.

Reuse #525's canonical process/thread identities and validation, including
once-only identity events and process-scoped threads. #618 owns common
schema/code generation, transport, loading/analysis integration, and viewer
compatibility. #696 owns the process-level router, context
registration/lifetimes, route ownership, and metadata replay. Multiple contexts
may represent the same OS process; no global process-entity singleton is
required.

The full implementation plan stays local. #740 remains a draft implementation PR
with a temporary placeholder; no implementation is included yet.

Issue #374, opened 2026-07-13
Phase 4: Coexist With Other Tools (Fan-out Mediator)


9prady9:
**What this delivers:** Quent can capture NVTX from a process at the same time as other tools (like Nsight Systems), without breaking them.

**In scope**
- A mediator that lets multiple NVTX consumers observe one process through a single real injection slot
- Passthrough so an external tool attached the normal way keeps producing its own trace, unmodified
- Isolation so one slow or failing consumer can't stop the others

**Out of scope**
- New capture or UI features (this sits underneath the proven slice)
- Real-GPU-workload validation (next phase)

**Deliverables**
- Fan-out mediator with per-consumer routing and failure isolation

**Done when**
- [ ] Multiple NVTX consumers observe one process simultaneously
- [ ] An external tool (e.g. Nsight Systems) attached the usual way still produces its correct trace while Quent captures the same process
- [ ] A slow or erroring consumer does not stop delivery to the others

> Note: this mediator has never been prototyped — worth a spike against real Nsight during earlier phases.

---
Part of #76 · Phase 4 of 5

PR #740, opened 2026-09-17
docs: compare shared event analysis approaches for #618

9prady9:
# Description

This documentation-only draft presents two implementation plans for #618 and asks for feedback before choosing an architecture. **No implementation is included.**

Both approaches share schema-based instrumentation, transport, storage, and the analysis lifecycle. They differ in where event reconstruction rules live:

- **Approach 1 — reusable reconstruction components:** retain NVTX pairing, nesting, and name resolution in a reusable Rust component behind a shared analysis pipeline.
- **Approach 2 — declarative reconstruction:** describe those rules in the schema and execute them using shared interval, stack, registry, and inventory primitives. A second source with matching semantics should need declarations, not a source-specific reducer.

**My current preference is approach 2**, because it extends schema-driven instrumentation into analysis. That preference is open for discussion, especially around rule-language scope, implementation cost, and preserving existing NVTX behavior. The two plans are alternatives; approach 1 is not a prerequisite for approach 2.

## Reading guide

- [Comparison, examples, and review questions](https://github.com/9prady9/quent/blob/unify_nvtx_exp_imp_analy/docs/plans/618-plan.md)
- [Approach 1: reusable reconstruction components](https://github.com/9prady9/quent/blob/unify_nvtx_exp_imp_analy/docs/plans/618-approach-1-components.md)
- [Approach 2: declarative reconstruction](https://github.com/9prady9/quent/blob/unify_nvtx_exp_imp_analy/docs/plans/618-approach-2-declarative.md)
- [Codebase research and migration dependencies](https://github.com/9prady9/quent/blob/unify_nvtx_exp_imp_analy/docs/plans/618-research.md)

Both approach documents start with a Mermaid architecture diagram, followed by concrete examples, ordered implementation tasks, compatibility constraints, risks, and acceptance criteria for the later implementation handoff.

## Feedback requested

1. Is declarative reconstruction the right scope for #618, or should the first delivery retain Rust components?
2. Is the proposed bounded rule set sufficient for NVTX without hiding reconstruction in source-specific adapters? Should it use a validated schema annotation or a first-class schema type?
3. Are the shared proposals for context streams, wire-format compatibility, source attachment, and the viewer entrypoint appropriate for existing consumers?

## Related Issues

- Relates to #618; this planning PR does not complete or close the issue.
- #638 has merged; #699 is still open at the planning snapshot. The plans target the post-#699 architecture, and implementation remains gated by the migration freeze.
- Live multi-context NVTX routing (#696) and multi-tool coexistence (#374) remain separate work.

## Testing

- `rumdl check docs/plans/618*.md` — passed with rumdl 0.1.67.
- Local Markdown link targets — checked.
- `git diff HEAD^ HEAD --check` — passed.
- Confirmed the commit contains only the four planning documents.
- Runtime tests were not run: this is documentation only. Tests described in the plans are future acceptance criteria, not claimed implementation results.

## Screenshots

Not applicable; the architecture diagrams are embedded in the two approach documents.


johanpel:
The way I would approach this is:

# YAML
- needs quent-os constraint https://github.com/rapidsai/quent/pull/525 to be able to mark entities as a process and threads
```yaml
quent: alpha
model: m

entities:
  MyProcess:
    events:
      init:
        attributes:
          process:
            type: { os: process } # marks this entity as representing a process
    nvtx: true # adds a top-level NvtxEvent type and all necessary records to the schema
    
  MyThread:
    events:
    . init:
        attributes:
          process:
            type: { os: thread } # allows correlation with NVTX events that carry OS TIDs
```

# Instrumentation:
- as soon as the handle for MyProcess is created, 
	- the NVTX callback table is populated to feed into the NVTX observer
- don't expose the NVTX observer publicly on the generated  quent context, we dont want users to touch this at all
- Potentially add a run-time flag to disable NVTX capture when constructing the Quent context

# Application-agnostic analysis:
- the analysis module for NVTX defines traits for all NVTX event data fields
- code generation for the application-specific schema's NVTX event types simply forwards all fields by implementing these traits for those event types
- the analysis module provides what it currently already provides but just bounds its implementation to these traits
- a lot of stuff i see in the "reconstruction" part of the declarative approach can all be run-time options for the analysis module, they need to be kept out of the YAML DSL

# Application-specific analysis:
- Sees all NVTX events.
- Does whatever it wants with NVTX events, but...
- Typically forwards the NVTX events to the application-agnostic analysis described above
- Does whatever it wants, but is able to correlate NVTX events by process and thread id if it wants to
- E.g. in query engine simulator, it can correlate NVTX ranges with task executor thread entities
- E.g. in the query engine UI, each thread resource can have its own little NVTX lane with sublanes for domains etc.


# Open questions

- Not entirely sure where the PID comes in but I think NVTX is process global so by constraining it to an { os : process } we can kind of provide more guardrails against misuse and potentially reject instrumentation trying to produce multiple entities marked as process within the same context or even process, but this is something to figure out later.

9prady9:
> The way I would approach this is:
> 
> # YAML
> * needs quent-os constraint [feat(os): add OS process and thread mod #525](https://github.com/rapidsai/quent/pull/525) to be able to mark entities as a process and threads
> 
> ```yaml
> quent: alpha
> model: m
> 
> entities:
>   MyProcess:
>     events:
>       init:
>         attributes:
>           process:
>             type: { os: process } # marks this entity as representing a process
>     nvtx: true # adds a top-level NvtxEvent type and all necessary records to the schema
>     
>   MyThread:
>     events:
>     . init:
>         attributes:
>           process:
>             type: { os: thread } # allows correlation with NVTX events that carry OS TIDs
> ```
> 
> # Instrumentation:
> * as soon as the handle for MyProcess is created,
>   
>   * the NVTX callback table is populated to feed into the NVTX observer
> * don't expose the NVTX observer publicly on the generated  quent context, we dont want users to touch this at all
> * Potentially add a run-time flag to disable NVTX capture when constructing the Quent context
> 
> # Application-agnostic analysis:
> * the analysis module for NVTX defines traits for all NVTX event data fields
> * code generation for the application-specific schema's NVTX event types simply forwards all fields by implementing these traits for those event types
> * the analysis module provides what it currently already provides but just bounds its implementation to these traits
> * a lot of stuff i see in the "reconstruction" part of the declarative approach can all be run-time options for the analysis module, they need to be kept out of the YAML DSL
> 
> # Application-specific analysis:
> * Sees all NVTX events.
> * Does whatever it wants with NVTX events, but...
> * Typically forwards the NVTX events to the application-agnostic analysis described above
> * Does whatever it wants, but is able to correlate NVTX events by process and thread id if it wants to
> * E.g. in query engine simulator, it can correlate NVTX ranges with task executor thread entities
> * E.g. in the query engine UI, each thread resource can have its own little NVTX lane with sublanes for domains etc.
> 


That is a reasonable middle ground, no need to over engineer the solution without any additional use cases right away.

What I meant by middle ground is we can keep reusing nvtx reconstruction while letting trait implementations provide the necessary info for reconstruction.

> # Open questions
> * Not entirely sure where the PID comes in but I think NVTX is process global so by constraining it to an { os : process } we can kind of provide more guardrails against misuse and potentially reject instrumentation trying to produce multiple entities marked as process within the same context or even process, but this is something to figure out later.

Entity instantiation doesn't mean there is a new process/thread right, it only records the process's presence in relation to this entity instance. So, when there are two Quent contexts, we should allow such recording because they are not two DB processes, they are just two contexts in the same process.

When nvtx comes into the play, I think we can use something I described here https://github.com/rapidsai/quent/issues/696 to handle multiple contexts.

johanpel:
> Entity instantiation doesn't mean there is a new process/thread right, it only records the process's presence in relation to this entity instance. So, when there are two Quent contexts, we should allow such recording because they are not two DB processes, they are just two contexts in the same process.

Yes, that is correct. That is the type of misuse I would want to implement guardrails for, so it would be impossible to create a stream of events from one process in which multiple entities exist that represent that OS process.

